PR #1, opened 2025-01-13
Improve Dockerfile to allow build with arm64 and amd64 architecture

WeinaJi:
## Context
With Apple M1 we will build with arm64 machine from time to time. The `Dockerfile` needs to be able to handle both arm64 and amd64 architectures, mainly the location of pre-installed libraries, either `/usr/lib/x86_64-linux-gnu` or `/usr/lib/aarch64-linux-gnu`. 

## Scope
Add the variable `ARCH` in `Dockerfile` and  `build_neurodamus.sh` with the value of "x86_64" or "aarch64" determined at build time.

## Testing
Local Apple M1 laptop and Linux amd64 PC. 

## Review


docker/Dockerfile
jplanasc, line 1: @WeinaJi I think there's a typo in the email ;)

docker/Dockerfile
```diff
@@ -1,5 +1,13 @@
...
+ARG TARGETARCH
+
+FROM ubuntu:22.04 AS linux_amd64
+ARG ARCH=x86_64
+
+FROM ubuntu:22.04 AS linux_arm64
+ARG ARCH=aarch64
+
+FROM linux_$TARGETARCH
+LABEL org.opencontainers.image.authors="Weina Ji <weina.ji@openblueinstitute.org>"
```
jplanasc, line 10: @WeinaJi I think there's a typo in the email ;)

WeinaJi: haha, indeed, I will correct that.

.github/workflows/simulation_test.yml
```diff
@@ -261,8 +261,8 @@ jobs:
         unset CORENEURONLIB
         pytest -s -x --forked --durations=5 --durations-min=15 tests/scientific-ngv
 
-    - name: live debug session, comment out
-      if: failure()
-      uses: mxschmitt/action-tmate@v3
-      with:
-       timeout-minutes: 60  # or any other value you need
+    # - name: live debug session, comment out
```
juanjosegarcan, line 264: The debugging session is no longer needed?

docker/Dockerfile
```diff
@@ -47,7 +57,7 @@ RUN mkdir -p $WORKDIR \
  && cmake --build rep_build --parallel \
  && cmake --install rep_build \
  && rm -rf libsonatareport rep_build
-ENV SONATAREPORT_DIR "$INSTALL_DIR"
+ENV SONATAREPORT_DIR="$INSTALL_DIR"
```
juanjosegarcan, line 60: Is there any difference with this change? Or is about style?

juanjosegarcan (approved):
I left a few comments to continue learning more about the repository

.github/workflows/simulation_test.yml
WeinaJi: We used to comment out those lines and uncomment them in a PR when we need to debug a failing session. I don't know in which PR those lines were put back. 
As the debug session is set 60 mins long, I think it is better to enable it when we want to debug. Otherwise, any failing workflow will last 60 mins longer.

docker/Dockerfile
WeinaJi: Yes, it is just about style. I saw the warnings with the `docker build` installed in my mac. The version is newer than I used in my linux desktop.

Issue #9, opened 2025-01-21
Minimal installation for unit tests


WeinaJi:
Currently we build the whole `neurodamus` tool chain including libsonata-report, libsonata, neuron and neurodamus from source. In this task I will check the minimal requirement of dependencies to run neurodamus unit tests, preferably from wheels.

evakennyobi:
fyi, I'm moving the sub-issues out of the main board, just the main issues we keep in the OBI 2025- board

mgeplf:
Nice progress, I'd say.

IMO, I think it would be great if one could run tox to launch all the tests, with the correct dependencies, and the build mod files.
So for `1)` above:
```
$ tox -eunit
```
should work once https://github.com/openbraininstitute/neurodamus/issues/6 is merged

Likewise, having the integration tests accessible with
```
$ tox -eintegration
```
And have tox install the dependencies, and build whatever is necessary would be the best.
I think even including the mod files SonataReport / ProbAMPANMDA_EMS in the repo for now could work, but the best would be to get them from a their actual repo.
If the above could just be a shell script (like the build_ndcore.sh), that can be used for the ci: https://github.com/openbraininstitute/neurodamus/blob/main/.github/workflows/simulation_test.yml and the Dockerfile would mean that there isn't redundancy and possible differences in all these places, which would also be a win, IMO.

WeinaJi:
Yes, I agree. The minimal installation could go to `build_ndcore.sh` to run with integration tests. I think we can use unit+integration tests for coverage measurement.

cattabiani:
I followed both this method and the one in [contributing](https://github.com/openbraininstitute/neurodamus/blob/main/CONTRIBUTING.md#testing) but the integration tests keep failing. 

- One thing I noticed is that neuron is never required. Should we require it at least in the tests? Maybe just for the integration tests? I would not bother to have 2 sets of requirements for unit and integration tests but I let it up to you
- it keeps crashing at `test_load_cell` (the first one) and I cannot get more log. I tried already various methods. Do you know what is the suggested way to get more log here?

WeinaJi:
The dependencies of different tests will be specified in `https://github.com/openbraininstitute/neurodamus/blob/main/tox.ini` . Currently the `[testenv:integration]` needs to be adapted to the minimal installation as I mentioned above. @juanjosegarcan may have already started. 

About the error in `test_load_cell`, can you try `pytest -s -x --forked tests/integration/test_cell.py`  ?

[...]

PR #12, opened 2025-01-22
Fix integration test from tox

juanjosegarcan:
## Context
Integration test need to be adapted to allow minimal installation form devs to run them as well as unit tests. 

## Scope
https://github.com/openbraininstitute/neurodamus/issues/9

## Review
* [ ] iIntegration tests running

### DIFF
--- a/ci/build_ndcore.sh
+++ b/ci/build_ndcore.sh
@@ -11,11 +11,11 @@ if [ -d _lib ]; then
 fi
 
 # Get the common synapses
-COMMON_DIR=_common
+COMMON_DIR=neurodamus-models/common
 if [ -d "$COMMON_DIR" ]; then
     ( cd "$COMMON_DIR" && git pull --quiet )
 else
-    git clone git@bbpgitlab.epfl.ch:hpc/sim/models/common.git $COMMON_DIR  --depth=1
+    git clone https://github.com/openbraininstitute/neurodamus-models.git neurodamus-models  --depth=1
 fi
 
 MOD_BUILD_DIR="mods.tmp"
@@ -23,12 +23,13 @@ mkdir -p $MOD_BUILD_DIR
 cp -f $CORE_DIR/mod/*.mod $MOD_BUILD_DIR
 cp -f $COMMON_DIR/mod/*.mod $MOD_BUILD_DIR
 nrnivmodl -incflags "-DDISABLE_REPORTINGLIB -DDISABLE_HDF5 -DDISABLE_MPI" $MOD_BUILD_DIR
-if [ ! -f x86_64/special ]; then
+ARCH=$(uname -m)
+if [ ! -f $ARCH/special ]; then
     echo "Error running nrnivmodl"
     exit 1
 fi
 mkdir -p _lib
-cp -f x86_64/libnrnmech* _lib/
+cp -f $ARCH/libnrnmech* _lib/
 cp -f $CORE_DIR/hoc/*.hoc _lib/
 cp -f $COMMON_DIR/hoc/*.hoc _lib/
 
--- a/ci/build_ndcore.sh
+++ b/ci/build_ndcore.sh
@@ -5,34 +5,43 @@
 set -euxo pipefail
 
 CORE_DIR="$1"
+BUILD_DIR="$2"
 
-if [ -d _lib ]; then
+LIBRARY_DIR=$BUILD_DIR/lib
+if [ $(uname) = "Darwin" ]; then EXT="dylib"; else EXT="so"; fi
+if [ -f "$LIBRARY_DIR/libnrnmech.$EXT" ]; then
+    echo "export HOC_LIBRARY_PATH=$LIBRARY_DIR" > $BUILD_DIR/.envfile
+    echo "export NRNMECH_LIB_PATH=$LIBRARY_DIR/libnrnmech.$EXT" >> $BUILD_DIR/.envfile
     exit 0
 fi
 
+
 # Get the common synapses
-COMMON_DIR=neurodamus-models/common
-if [ -d "$COMMON_DIR" ]; then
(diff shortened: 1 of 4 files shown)


mgeplf:
Nice!
LGTM, I can run `tox -eintegration` on my debian install, and it cleanly runs the tests (there are warnings about pytest usage, but they were there before, and can be fixed elsewhere.

If everyone else can check that it runs on their machines, then I think this should be merged.

cattabiani (dismissed):
I see that you used my `build_ndcore`, nice to be of service! LGTM

cattabiani:
knitpick: we may want to add `neurodamus-models` to the `.gitignore` file

mgeplf:
> knitpick: we may want to add neurodamus-models to the .gitignore file

Alternatively, the installation of such files could happen within the tox environment (ie: `{toxworkdir}`), so they don't pollute the namespace at all.

WeinaJi:
It is also working in my local apply M1. 
I see there is `libnrnmech.dylib` created instead of `libnrnmech.so`. However, it is still working without the change of `NRNMECH_LIB_PATH` in `tox.ini`. Tox is doing something fancy. Anyone knows why ? 
BTW, I think  we can also remove this line as there is no such folder.  
```
PYTHONPATH = {toxinidir}/core/python
```

mgeplf:
> However, it is still working without the change of NRNMECH_LIB_PATH in tox.ini. Tox is doing something fancy. Anyone knows why ?

IIRC, doesn't NEURON include $ARCH/libnrnmech.so to it's search path automatically? I don't think it's tox that's doing anything - it doesn't know about how NEURON works.

WeinaJi:
> > However, it is still working without the change of NRNMECH_LIB_PATH in tox.ini. Tox is doing something fancy. Anyone knows why ?
> 
> IIRC, doesn't NEURON include $ARCH/libnrnmech.so to it's search path automatically? I don't think it's tox that's doing anything - it doesn't know about how NEURON works.

I thought that was only when we launched with `$ARCH/special`. But now it seems like that. Then `NRNMECH_LIB_PATH` is not needed in `tox.ini`. 
@juanjosegarcan , can you try to remove that ?

mgeplf:
if we move the mech building to be not in the root directory, then we'll probably need that environment variable set.
Not building the mechs in the repository root directory would be good, I think.

WeinaJi:
> if we move the mech building to be not in the root directory, then we'll probably need that environment variable set. Not building the mechs in the repository root directory would be good, I think.

Indeed. It is safer to set `NRNMECH_LIB_PATH` explicit, then we need to handle `.so` and `.dylib` extensions.
 @juanjosegarcan , sorry for the change again, it seems that we need `NRNMECH_LIB_PATH` back.

WeinaJi:
A little context, neurodamus required  the mechanisms `SonataReport` and `ProbAMPANMDA_EMS`, the check is done [in this function](https://github.com/BlueBrain/neurodamus/blob/e5fef8d5fb2047e4de76f21351da7b1c048de94d/neurodamus/core/_neurodamus.py#L82). The logic is that either the `mech` should be already loaded in NEURON or compiled in the library `NRNMECH_LIB_PATH`.
It seems that NEURON has included $ARCH/libnrnmech.so(dylib) to its path by default as @mgeplf  mentioned above, so in such case the check in `NRNMECH_LIB_PATH` is skipped. 

Simply we can check the default loaded mechanism in Neuron with python
```
>>> from neuron import h
>>> hasattr(h, "SonataReport")
True
```
When the built `libnrnmech.so(dylib)` is in another location such as `tmp/$ARCH/libnrnmech.so(dylib)`, Neuron won't load it by default. In this case, we need to check `NRNMECH_LIB_PATH`.

WeinaJi:
FYI, The default library loading in NEURON is done [in this function](https://github.com/neuronsimulator/nrn/blob/90900fa17f75849c665f6b1e45194538dec4fb22/share/lib/python/neuron/__init__.py#L300). The actual library loaded by default is `$ARCH/.libs/libnrnmech.so` that also exists for `arm64`. 
@juanjosegarcan  I think we need to keep `{toxinidir}/_lib/libnrnmech.so` in `tox.ini`. And in the shell script, you can copy `$ARCH/.libs/libnrnmech.so` to `_lib`. Then it is good for both arm64 and x86_64.

juanjosegarcan:
@WeinaJi @mgeplf Please review. I added extension check for the library and moved the build_folder to {toxworkdir}

ci/build_ndcore.sh
```diff
@@ -6,32 +6,40 @@ set -euxo pipefail
 
 CORE_DIR="$1"
 
-if [ -d _lib ]; then
+LIBRARY_DIR=$BUILD_MODS_DIR/lib
```
mgeplf, line 9: instead of requiring `BUILD_MODS_DIR` in the environment, the script can be made more flexible by passing it in like `CORE_DIR`

Also, I find the naming a bit confusing, since there is a `BUILD_MODS_DIR`, and `MOD_BUILD_DIR`

juanjosegarcan: I added your suggestions!

WeinaJi: Maybe `MOD_DIR` instead of `MOD_BUILD_DIR` ? Just one nitpick.

juanjosegarcan: @WeinaJi done!

WeinaJi (approved):
The integration tests look good to me. Thanks for the implementation.

@mgeplf , do you know if we can push the coverage data for several workflows of the same build to coveralls? It would be nice to have the coverage report for integration test separately from the unit test, and ideally another report of the combined coverage. We can implement it in another PR.

mgeplf:
I will add coverage separately, I looked into coveralls with different labels/facets, but it doesn't seem to support our use case.  I think I will work around it by using coveralls and codecov to do the separation.

PR #15, opened 2025-01-24
Format

mgeplf:
Format `neurodamus` so that it's uniform.

Once this is merged, a second commit will happen that will include add an `.git-blame-ignore-revs`
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,3 +77,13 @@ source = [
 [tool.coverage.report]
 show_missing = true
 precision = 0
+
+[tool.ruff]
+line-length = 100
+target-version = "py39"
+
+[tool.ruff.lint.isort]
+known-local-folder = [
+    "neurodamus",
+]
+combine-as-imports = true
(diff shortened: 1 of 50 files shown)


mgeplf:
FYI: the code changes shouldn't require any review; they are the result of running `tox -eformat`

pyproject.toml
```diff
@@ -77,3 +88,13 @@ source = [
 [tool.coverage.report]
 show_missing = true
 precision = 0
+
+[tool.ruff]
+line-length = 100
+target-version = "py39"
```
WeinaJi, line 94: I guess that's the same as `requires-python=">=3.9"` in line 23. Maybe no need to add again ?

WeinaJi (commented):
Just one nitpick. Otherwise all look good. Thanks!

pyproject.toml
mgeplf: ah, cool, I didn't know ruff inherited it from pyproject:
`If you're already using a pyproject.toml file, we recommend project.requires-python instead, as it's based on Python packaging standards, and will be respected by other tools. For example, Ruff treats the following as identical to target-version = "py38":`

Thanks!

WeinaJi (approved):

Issue #6, opened 2025-01-20
split view into category of tests


mgeplf:
split view into category of tests
versus getting summary at end
tool that shows break down (coveralls, codecov)

mgeplf:
I broke out the unit tests in: https://github.com/openbraininstitute/neurodamus/pull/7

Once using coveralls.io is approved, then I will wire that up.

For now, one should be able to run: `tox -eunit`
And get the current coverage for the unit tests:

```
---------- coverage: platform linux, python 3.9.21-final-0 -----------
Name                                                                                                                        Stmts   Miss  Cover   Missing
---------------------------------------------------------------------------------------------------------------------------------------------------------
/home/runner/work/neurodamus/neurodamus/.tox/unit/lib/python3.9/site-packages/neurodamus/__init__.py                           10      2    80%   15-16
/home/runner/work/neurodamus/neurodamus/.tox/unit/lib/python3.9/site-packages/neurodamus/cell_distributor.py                  603    459    24%   47-52, 59, 66, 70, 78-82, 88, 91, 94, 121-139, 158, 161, 166, 169, 172-186, 190-193, 198-221, 225-237, 240-253, 256-266, 275-281, 285-302, 313-354, 357-359, 364-387, 397, 400-405, 408-411, 414, 417, 420, 425-436, 439-443, 447-450, 454, 464-466, 469, 472, 478-482, 485, 488-494, 497-498, 506-511, 522-526, 529-530, 542-545, 550-561, 570-586, 615-621, 627-633, 638-639, 649-679, 688-715, 723-739, 745-755, 766-771, 775-806, 813-818, 822-826, 835-836, 843-845, 851-876, 882-898, 904-909, 913, 919-920, 925-926, 930-931, 935, 942-962
/home/runner/work/neurodamus/neurodamus/.tox/unit/lib/python3.9/site-packages/neurodamus/commands.py                          113    113     0%   4-230
[...]

PR #17, opened 2025-01-28
Add baseline import coverage, turn on `integration` test reporting

mgeplf:


mgeplf:
This solves #6 
It's not as smooth as I would like, but the differently labeled tests get comments on Github:
image

And if one clicks around in coveralls, once can see an overview with the different labels here:
https://coveralls.io/builds/71949441

.github/workflows/unit-test.yml
```diff
@@ -7,7 +7,7 @@ on:
         - master
 
 jobs:
-  build:
+  baseline-coverage:
```
WeinaJi, line 10: And maybe a better name as this job does more than `baseline-coverage`?

.github/workflows/unit-test.yml
WeinaJi, line 1: And shall we also rename `unit-test.yml` to something like `coverage.yaml`?

WeinaJi (commented):
The coverage report looks great! Thanks!
Just a few nitpicks about naming.

tox.ini
```diff
@@ -9,18 +9,25 @@ deps =
...
+[testenv:baseline]
+deps =
+    {[testenv]deps}
+    pytest-cov
+commands =
+    pytest --cov=neurodamus --cov-report=term-missing tests/test_import.py
+
 [testenv:integration]
 deps =
     {[testenv:unit]deps}
```
juanjosegarcan, line 27: I put this deps starting from unit ones, but maybe to be coherent we should set all deps explicitly (except for general deps)?

juanjosegarcan (approved):

.github/workflows/unit-test.yml
mgeplf: done

.github/workflows/unit-test.yml
mgeplf: done

tox.ini
mgeplf: that would probably make sense, although I'm not sure how it interacts with the other testing (ie: `bbp-model`, for instance), and I didn't want to disturb things.  I guess it could be done in another PR.

juanjosegarcan (approved):

PR #18, opened 2025-01-28
Remove compat types

WeinaJi:
## Context
`Compat` types were designed to behavior similar to Hoc objects. It was necessary when neurodamus needed to parse the legacy config file (BlueConfig) as well as the SONATA config file. Now with the only SONATA case, it should be possible to replace them with python types.

## Scope
Remove `compat.py` which contains
- [ ] Compat.Vector
- [ ] Compat.List
   - [ ]  required by `ConnectionUtils.hoc`
- [x] Compat.Map
   - [ ] required by `RNGSetting.hoc`
- [ ] function `hoc_vector`

## Review

### DIFF
--- a/docs/api/subpackages/utils.rst
+++ b/docs/api/subpackages/utils.rst
@@ -15,7 +15,6 @@ neurodamus.utils.compat
 
    .. autosummary::
       List
-      Map
       Vector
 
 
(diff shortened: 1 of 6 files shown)


cattabiani:
are you sure you do not want help for this? It looks like a difficult task

WeinaJi:
> are you sure you do not want help for this? It looks like a difficult task

hmm... I am still studying the case whether it is a feasible task. 
ATM, I already see that `Compat.List` and `Compat.Map` are required by several hoc fuctions. If we can rewrite those hoc functions to python, the compat types can be removed. But not easy to rewrite. Not mentioning `Compat.Vector`, which is mostly likely used by hoc functions.

PR #20, opened 2025-01-30
Improvements on the Dockerfile

WeinaJi:
## Scope
1. Use the bash command `uname -m` to detect the build architecture when needed instead of the Global ARG variable. 
2. Copy those common bluebrain hoc and mod files from neurodamus-models to $NEURODAMUS_MODS_DIR which are required for instantiating neurodamus.
3. Add a note in README about the architecture
 
## Testing
Locally tested with apply M1 as well as Ubuntu linux.

## Review


docker/Dockerfile
```diff
@@ -94,11 +84,21 @@ ENV NEURODAMUS_MODS_DIR="$WORKDIR/neurodamus/neurodamus/data/mod"
 ENV PATH="$INSTALL_DIR/bin:$USR_VENV/bin:$PATH"
 ENV PYTHONPATH="$INSTALL_DIR/lib/python:$PYTHONPATH"
 
-# Copy and edit module building script and test build
+# Copy common bluebrain hoc and mod files from neurodamus-models, required for instantiating neurodamus
+RUN git clone https://github.com/openbraininstitute/neurodamus-models.git --recursive --depth 1 \
```
juanjosegarcan, line 88: Since we are using only 4 files from the repo, would it be better to just download those files? (eg: curl -O https://github.com/openbraininstitute/neurodamus-models/blob/main/common/mod/ProbAMPANMDA_EMS.mod )

juanjosegarcan (approved):

docker/Dockerfile
WeinaJi: Done

juanjosegarcan (approved):
For me just to be rebased and merged!

PR #22, opened 2025-01-30
Move to conventional source & build sphinx configuration

mgeplf:
* use sphinx provided Makefile; call it via tox
* this conventional layout (`docs/source`, `docs/build`) means one knows what can be safely removed.
* the Makefile can also take care of cleanup
* finally, this doesn't require PYTHONPATH hacks in tox


docs/changes.rst
WeinaJi, line 1: Where is this file moved to? I prefer to keep it somewhere.

cattabiani:
@mgeplf may you be less succinct in the description? It would help for people reviewing the PR. For example, I remember you mentioned in a meeting that you want to move everything in source to clearly define what ... can be thrown away and what is source code? That would help understand why this changes are necessary

docs/changes.rst
WeinaJi: ah, I see the changelog is still shown. That's good! Please ignore the question.

mgeplf:
> @mgeplf may you be less succinct in the description?

I added some more, hope that helps.

requirements_docs.txt
```diff
@@ -1,3 +0,0 @@
-sphinx>=2.0.0
-sphinx-bluebrain-theme
-sphinx-autorun
```
cattabiani, line 3: this is more a curiosity question. I see that in the CI https://neurodamus--22.org.readthedocs.build/en/22/

the theme is still bbp. However, we removed it from here. From the repo I cannot find anything about `bluebrain-theme`. Where does it pick it from?

cattabiani (approved):
apart from my question (non-blocking) the rest is LGTM. I am not a sphinx expert but the changes look reasonable

WeinaJi (approved):

requirements_docs.txt
mgeplf: IMO, all the dependencies should be managed from one place when it's clear they can be (as opposed to requirements.txt) so I put them there:

https://github.com/openbraininstitute/neurodamus/pull/22/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R28

cattabiani: strange that I could not find that line. Maybe I confused repos or branches as I was jugling with 3 branches in 2 repos. My bad. Thank you for the clarification

juanjosegarcan (approved):

PR #26, opened 2025-02-03
remove Map.hoc, compat.Map, and compat.PyMap

### LINKED ISSUE #24: Replace Compat.Map with Python dict
As I already tried in https://github.com/openbraininstitute/neurodamus/pull/18, `Compat.Map` is used by https://github.com/openbraininstitute/neurodamus/blob/8f1722ee325bcab32f061fba1f1bece99e575a44/neurodamus/data/hoc/RNGSettings.hoc#L35 .
To replace `Compat.Map`, I think we need to rewrite `RNGSettings.hoc` to a python class. The essential part for setting the random seed in Neuron is 
```
        // random123 can set the global index now
        rng = new Random()
        rng.Random123_globalindex( globalSeed )
```

### LINKED ISSUE #27: Simplify `connection._modoverride` to be a string instead of a dict objection
As @cattabiani  has found out, in the case of modoverride, we are passing the whole connection config object to the helper class
https://github.com/openbraininstitute/neurodamus/blob/55e83348f5a686df84097eb338a7d704fd5b9ae0/neurodamus/connection.py#L534
However, the only modoverride helper `GluSynapseHelper.hoc` reads it as object (`parseConnection = $o8`) but does not uses it. 
https://github.com/openbraininstitute/neurodamus-models/blob/main/common/hoc/GluSynapseHelper.hoc#L45

We can simplify it by changing `connection._modoverride` to a string. It will require to remove the `parseConnection` from https://github.com/openbraininstitute/neurodamus-models/blob/main/common/hoc/GluSynapseHelper.hoc. 
@jamesgking , do you remember why there is an used `parseConnection` in  `GluSynapseHelper.hoc` ? Maybe it was used before but not needed any more?

cattabiani:
## Context

Since the removal of BlueConfig these objects became obsolete because the sonata reader returns proper python objects. Thus, we can just use python objects without passing by a hoc.

A neurodamus-models tag has been created in the new obi repo. We can move forward.

Fix #24 
Parital Fix #27

## Scope

- remove Map.hoc
- remove compat.Map
- remove compat.PyMap
- fix `RNGSettings.hoc`: make `proc interpret()` to accept a python dict as `runInfo`.

## Testing

Atm I hope that the unit and integration tests catch any problems that may arise from this. Let me know if you think I need to add more tests (adding a constructive suggestion of what test I should add exactly and where)

## Review


cattabiani:
I feel this PR will always fail until:
- I test https://github.com/openbraininstitute/neurodamus-models/pull/2 (catch 22 situation)
- I merge it
- I create a new tag for the neurodamus-models
- I point to that tag series (have #28 solved)

neurodamus/connection.py
```diff
@@ -527,11 +529,11 @@ def _create_synapse(self, cell, params_obj, x, syn_id, base_seed):
         """
         is_inh = params_obj["synType"] < 100
         if self._mod_override is not None:
-            mod_override = self._mod_override.get("ModOverride").s
+            mod_override = self._mod_override
```
WeinaJi, line 532: It seems that the var `mod_override` is not needed, we can pass directly `self._mod_override` to build `override_helper` ?

neurodamus/connection.py
```diff
@@ -265,8 +266,9 @@ def add_synapse_configuration(self, configuration):
         if configuration is not None:
             self._configurations.append(configuration)
 
-    def override_mod(self, mod_override):
-        assert mod_override.exists("ModOverride"), "ModOverride requires hoc config obj"
+    def set_mod_override(self, mod_override):
+        """Set a valid mod override"""
+        assert mod_override is not None, "ModOverride cannot be None"
```
WeinaJi, line 271: For readability, I would prefer to write a setter and getter with `property` decorator. 
And do you think it possible to have some unit tests for the `ConnectionBase` class? You can add a test for the getter and setter of modoverride in this PR.

neurodamus/connection.py
cattabiani: ah yes, that is a relic of debugging

neurodamus/connection.py
cattabiani: how do you write a setter with the property decorator? @property functions cannot have inputs except self and setters have always at least 1 input apart from self...

cattabiani: ok, after discussion I got it. @property for the getter and @mod_override.setter for the setter. Done

neurodamus/utils/multimap.py
```diff
@@ -5,10 +5,13 @@
 
 import numpy as np
 
-import collections.abc
+try:
+    import collections.abc as collections_abc  # Py >= 3.3
```
juanjosegarcan, line 9: In the commit just merged to master (https://github.com/openbraininstitute/neurodamus/commit/f8e710a3a0daf8827c4816ec554655587e4790eb) we are removing all the support for python 2. If is clear we are not supporting it any more we can just change it here too.

cattabiani: 3.2 is > 2 and < 3.3. Do we also remove support for everything below 3.3 but above 2?

juanjosegarcan: May it has sense to not support it. CI is running starting from 3.9 version. What do you think? @mgeplf @WeinaJi @jamesgking

mgeplf: anything below 3.9 can be removed, since we're setting that as the min python version;
https://github.com/openbraininstitute/neurodamus/blob/main/pyproject.toml#L23

Everything prior to 3.9 is already end of life, so things like numpy updates don't exist for it

cattabiani: ah I missed that line. Thx!

juanjosegarcan (approved):

WeinaJi (approved):

Issue #23, opened 2025-01-30
Case study: replace Compat.Vector with python data type


WeinaJi:
Compat types (https://github.com/openbraininstitute/neurodamus/blob/main/neurodamus/utils/compat.py) were designed to behavior similar to Hoc objects. It was necessary at the time when neurodamus needed to parse the legacy config file (BlueConfig parsed by a hoc parser) as well as the SONATA config file. Now with the only SONATA case, is it possible to replace them with python types?

When I tried to replace `Compat.Map` in https://github.com/openbraininstitute/neurodamus/pull/18, I notice that it is the input of the hoc function `inteprete()` in `RNGSetting.hoc` which requires a hoc [Map](https://github.com/openbraininstitute/neurodamus/blob/main/neurodamus/data/hoc/Map.hoc).  And `Compat.List` is required by `ConnectionUtils.hoc`. 

This task is to study whether `Compat.Vector` can be simply replaced by a Python type, array, list or numpy array. Or is it used in a hoc function?

PR #33, opened 2025-02-05
remove `compat.list`

cattabiani:
## Context

The compat functions were used mostly for when we had to use `BlueConfig`. Now it is no more. This is to investigate if it is possible to simplify the code. 

Fix: #23

Since I need it, I am also fixing the following:

#39

## Scope

Remove mentions of the compat objects and replace them with either python objects (leveraging the neuron python API) or hoc functions. 

## Testing

Check if the unit and all the various integration tests run.

## Review


cattabiani:
this pr became so outdated that it is better to restart

Issue #11, opened 2025-01-22
Create a SONATA version of the ring test


mgeplf:
* There should be a script that creates the node/edge files (see something like https://github.com/BlueBrain/libsonata/blob/master/tests/data/generate.py or  https://github.com/BlueBrain/sonata-extension/tree/master/sonata-generator/sonata_generator, but cut and paste the code into the neurodamus repo, so it doesn't get lost)
* the 'length' of the ring could be a parameter (ie: 2 for 2 nodes, 100 for 100)
* different population configurations could also be tested; but these should create different files, so that the tests are decoupled, and don't have to be rewritten if someone augments a particular generator

WeinaJi:
The NEURON ringtest uses the model template [B_BallStick](https://github.com/neuronsimulator/ringtest/blob/master/cell.hoc) which doesn't require extra mechanisms or any mophology. The number of dendrites and segments can be customized in this model template. An example of creating a ring of `B_BallStick` cells in NEURON is [ring.py](https://github.com/neuronsimulator/ringtest/blob/master/ring.py).
If we create a SONATA circuit of two rings of that cell, currently what I am thinking of is:
- Node populations:
1. RingA: a ring with N cells 
2. RingB: a ring with M cells
- Edges populations:
1. RingA_RingA_chemical
2. RingB_RingB_chemical
3. RingA_RingB_chemical
4. RingA_RingB_electrical (for gap junction tests)

Then what values are we going to set for the mandatory fields in each node or edge file?  The `model_template` could be something like `B_BallStick` w.r.t to the cell model.  How about the others?

mgeplf:
@WeinaJi, that makes sense to me.
In addition to that, I think we could have a separate version that includes the morphology file that has a (similar? getting exact) geometry to what is in the ring test.

I would also probably start with a "simpler" one, where there is just a single node and edge population, but the same connectivity of what you describe.
That way, both could be run, and it would help narrow things down: for instance if the single population version works, it's the "gold standard", and if the other one doesn't, then it's likely that it's population handling that is the culprit.

> Then what values are we going to set for the mandatory fields in each node or edge file? The model_template could be something like B_BallStick w.r.t to the cell model. How about the others?

Not sure. I believe you had something in https://github.com/openbraininstitute/neurodamus/commit/352af41bf7d385ba5fa7db235e99def625733675 but I'm not sure if that could be used in this case.

WeinaJi:
[...]

PR #49, opened 2025-02-07
Add a SONATA circuit for the ringtest

WeinaJi:
## Context
As per discussion in https://github.com/openbraininstitute/neurodamus/issues/11, this PR adds the ringtest circuit which creates 2 rings of simple `BallStick` cells.
```
Nodes:
- RingA: 3 cells, 2 sections/cell, 2 segments/section
- RingB: 2 cells, 2 sections/cell, 2 segments/section
Edges:
- RingA: 0 -> 1 -> 2 ->0
- RingB: 0 -> 1 -> 0
- RingA: 0 -> RingB: 0
```
## Scope

1. In the test folder: adds a SONATA version of the ring test circuit including 2 nodes files + 3 edges files, the relevant SONATA config files, reference cell dump output json files.
2. In neurodamus/utils, a new function `dump_cellstate` to dump cell/section/segments/netcons/synapse attributes from the NEURON context

## Testing
1. `test/integration/test_ringcells.py` can be run with `tox -e integration` 
     `test_dump_RingB_2cells`: dump 1 population with 2 cells
     `test_dump_RingA_RingB`: dump 2 populations 

## Review


tox.ini
```diff
@@ -36,7 +36,7 @@ allowlist_externals =
     sh
 commands =
     {toxinidir}/ci/build_ndcore.sh {toxinidir}/neurodamus/data {toxworkdir}
-    sh -c ". {toxworkdir}/.envfile && pytest -x --cov=neurodamus --cov-report=term-missing tests/integration"
+    sh -c ". {toxworkdir}/.envfile && pytest -x --cov=neurodamus --cov-report=term-missing --forked tests/integration"
```
WeinaJi, line 39: I had to run the tests in this folder in separate forked processes, otherwise failure in CI. Instead of setting`pytestmark = pytest.mark.forked` in each test file,  I would prefer to set it centrally during launching.

jamesgking (dismissed):

jamesgking (approved):

tests/simulations/ringtest/local_edges_A.h5
mgeplf, line 1: could we include the code that generated these `.h5` files (same for the nodes*.h5); that will make it easier in the future to make more tests that have different configurations?

WeinaJi: I use the `sonata-generator` built from https://github.com/openbraininstitute/sonata-extension/tree/master/sonata-generator, with customized `neuronal_generators.py`. 
Can I push to a branch in `sonata-extension` and reference it in a README doc in the circuit folder ?

mgeplf: Interesting, I will have to think about that.
I don't think it should go into `sonata-extension`, but since it depends on that, I don't think we should copy in the whole source tree.
Keeping the test data generation self-contained is useful, imo.

I will think about it.

PR #57, opened 2025-02-07
Local tests

mgeplf:


cattabiani:
is this still relevant @mgeplf ?

mgeplf:
Yes, I need to finish it, but it will reduce the differences in our testing.

Issue #47, opened 2025-02-07
Fix Linting Umbrella task


mgeplf:
If no files are excluded from the `neurodamus/` directory, currently `ruff check` reports 1387 errors, even with the exceptions in:
https://github.com/openbraininstitute/neurodamus/blob/f8e710a3a0daf8827c4816ec554655587e4790eb/pyproject.toml#L99

This umbrella ticket is to try and tackle this.
These are the files that need to be fixed:

- [ ] neurodamus/cell_distributor.py
- [ ] neurodamus/commands.py
- [ ] neurodamus/connection_manager.py
- [ ] neurodamus/connection.py
- [ ] neurodamus/core/cell.py
- [ ] neurodamus/core/configuration.py
- [ ] neurodamus/core/coreneuron_configuration.py
- [ ] neurodamus/core/_engine.py
- [ ] neurodamus/core/__init__.py
- [ ] neurodamus/core/mechanisms.py
- [ ] neurodamus/core/_mpi.py
- [ ] neurodamus/core/_neurodamus.py
- [ ] neurodamus/core/_neuron.py
- [ ] neurodamus/core/nodeset.py
- [ ] neurodamus/core/random.py
- [ ] neurodamus/core/_shmutils.py
- [ ] neurodamus/core/stimuli.py
- [ ] neurodamus/core/synapses.py
- [ ] neurodamus/core/_utils.py
- [ ] neurodamus/data/init.py
- [ ] neurodamus/gap_junction.py
- [ ] neurodamus/gj_user_corrections.py
- [ ] neurodamus/hocify.py
- [ ] neurodamus/__init__.py
- [ ] neurodamus/io/cell_readers.py
- [ ] neurodamus/io/__init__.py
- [ ] neurodamus/io/sonata_config.py
- [ ] neurodamus/io/synapse_reader.py
- [ ] neurodamus/lfp_manager.py
- [ ] neurodamus/metype.py
- [ ] neurodamus/modification_manager.py
- [ ] neurodamus/morphio_wrapper.py
- [ ] neurodamus/neuromodulation_manager.py
- [ ] neurodamus/ngv.py
- [ ] neurodamus/node.py
- [ ] neurodamus/replay.py
- [ ] neurodamus/report.py
- [ ] neurodamus/stimulus_manager.py
- [ ] neurodamus/target_manager.py
- [ ] neurodamus/utils/cli.py
- [ ] neurodamus/utils/compat.py
[...]

PR #63, opened 2025-02-10
Transpose lint file except to rule except

mgeplf:
* switch from file lint exceptions to rule lint exceptions (as agreed in meeting Feb 10/2025)
* Clean up some of the lints:
  remove exc_info=True, ignore one that is ok
  clean up some imports
  "RET503", # implicit-return
  "G003",   # logging-string-concat
  "RUF052", # used-dummy-variable


juanjosegarcan (approved):

mgeplf:

Issue #60, opened 2025-02-10
Add testing for stimuli from sonata extensions


jamesgking:
https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#inputs

the documentation is built from here: https://github.com/openbraininstitute/sonata-extension

This ticket can be for subset:

 “linear”, “relative_linear”, “pulse”, “sinusoidal”,  “subthreshold”, “hyperpolarizing”,  “seclamp”

cattabiani:
It is not clear to me what we want in this issue. Do we want to expand the unit tests (and reorganize)? Do we want to expand integration tests? A concrete example: 
- how should I serve the input that I need to test? via sonata config file? Can I mock that so that I test only the input class as in the unit tests we have? 
- how should I verify that the input is working? can I follow what the current unit/integration tests are doing or should I run a simulation and check the voltage/current traces?

@jamesgking

jamesgking:
For this, I was thinking that we add tests such that the unit test coverage expands (you had started this I recall)

And then we can add an integration tests using the ring network @WeinaJi started.  Happy to discuss in detail

Issue #62, opened 2025-02-10
Add testing for rng stimuli from sonata extensions


jamesgking:
https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#inputs

This ticket can be for subset:

“noise”, “shot_noise”, “relative_shot_noise”, “absolute_shot_noise”, “ornstein_uhlenbeck”, “relative_ornstein_uhlenbeck”

cattabiani:
solved

PR #69, opened 2025-02-13
test stimuli

cattabiani:
## Context
We need to expand our testing campaign.

Fix: #60 #62 #77 #78

## Scope
Stimuli is a well defined chunk of code that can be easily tested extensively. Atm we have 2 test files that tests only parts of the stimuli object, one in unit and one in integration tests. They should be collected in 1 file that fully tests everything in the stimuli file. 

It should be done as an integration test because we need the full h.Vector. The mocked version of that used in the unit tests is not enough for the alrady existing integration tests. The h.Vector is fundamental for the stimuli object because of the `play` capabilities. 

#### Tests
- [x] `SignalSource` methods
  - [x] remove `while ev[-1] < duration:` as we always overshoot
  - [x] emit error for shot pulse with `tau_D == tau_R` since we do not know what should the function be. Better than a divide by 0 error.
- [x] `CurrentSource` methods
  - [x] ~~remove `stim_vec_mode` since it is always set to `True`. Let me know if `_Clamp` is used explicitly somewhere~~ EDIT: used by `Constant`
- [x] `ConductanceSource`
  - [x] remove `stim_vec_mode` as it is a copy-paste from `CurrentSource`. However, since `Constant` is missing here, it is always True
  - [x] fix behavior around 0 for `ConductanceSource`

- [x] after discussion, add assert to be sure that we do not have regressions

## Testing
This is TESTING!!! (it should sound as: this is SPARTAAA!)

## Blockers:
- [x] Decide what to do with: #77 

## Review


tests/unit/test_stimuli.py
```diff
@@ -1,55 +1,499 @@
...
-    def test_delay_ramp(self):
-        # When a delay is specified (in ctor or factory) base_amp is set on t=0 too
-        sig = CurrentSource.ramp(1, 2, 2, base_amp=-1, delay=10)
-        assert list(sig.time_vec) == [0, 10, 10, 12, 12]
-        assert list(sig.stim_vec) == [-1, -1, 1, 2, -1]
-
-    def test_train(self):
+        expected = np.array([-self.base_delay, 0, 0, 10, 10]) + self.base_delay
+        assert np.allclose(list(self.stim.time_vec), expected)
+        assert np.allclose(list(self.stim.stim_vec), [2, base_amp, 1.2, 1.2, base_amp])
```
juanjosegarcan, line 82: Could we use self.base_amp instead of the literal 2?

tests/unit/test_stimuli.py
```diff
@@ -1,55 +1,499 @@
...
+                    320,
+                    320,
+                    350,
+                ]
+            )
+            + self.base_delay
+        )
+        assert np.allclose(self.stim.time_vec, expected)
+        assert np.allclose(
+            self.stim.stim_vec, [self.base_amp] + [2, 1.2, 1.2, 2] * 4 + [self.base_amp]
```
juanjosegarcan, line 170: Literal 2 to self.base_amp?

cattabiani: done

tests/unit/test_stimuli.py
cattabiani: done

juanjosegarcan (dismissed):

neurodamus/core/stimuli.py
```diff
@@ -201,22 +217,17 @@ def add_shot_noise(self, tau_D, tau_R, rate, amp_mean, amp_var, duration, dt=0.2
...
+        ev = Nd.h.Vector()
         ev.integral(iei, 1).mul(1000)  # generate events in ms
         # add events if last event falls short of duration
-        while ev[-1] < duration:
-            iei_new = Neuron.h.Vector(100)  # generate 100 new inter-event intervals
-            iei_new.setrand(rng)  # here rng is still negexp
-            ev_new = Neuron.h.Vector()
-            ev_new.integral(iei_new, 1).mul(1000).add(ev[-1])  # generate new shifted events in ms
-            ev.append(ev_new)  # append new events
+
```
WeinaJi, line 226: Is the last event no more added? The comment is still there.

neurodamus/core/stimuli.py
```diff
@@ -465,37 +487,32 @@ def __init__(
...
-                # replace self.stim_vec with inverted and clamped signal
-                # rs is in MOhm, so conductance is in uS (micro Siemens)
-                self.stim_vec = Neuron.h.Vector(
-                    [1 / x if x > 1e-9 and x < 1e9 else 1e9 for x in self.stim_vec]
-                )
-                self.stim_vec.play(self.clamp._ref_rs, self.time_vec, 1)
-            else:
-                for param, val in clamp_params.items():
-                    setattr(self.clamp, param, val)
+            assert time_vec is not None
```
WeinaJi, line 503: Is it better to keep the `if-else` condition for the future implementation ?

neurodamus/core/stimuli.py
cattabiani: ah thank you. that is a typo

neurodamus/core/stimuli.py
cattabiani: atm there is no way to reach that point (except ad hoc examples where we use _* methods) as the only function that instantiate these objects is `attach_to`. I would things as lean as possible. We can easily re-add if the feature is needed

juanjosegarcan (dismissed):

WeinaJi (approved):
LGTM. Thanks!

PR #73, opened 2025-02-14
allow running of specified unit test using tox: ex `tox -eunit -- tests/unit/test_ringcells.py 

mgeplf:


mgeplf:
> This is very useful feature. Is it mentioned in our internal dev documentation?

Not at the moment, I'm not sure exactly where it would go that it would be noticed.
Perhaps we can link to the tox user guide - it mentions it, and is generally a short and good "intro" type document: https://tox.wiki/en/latest/user_guide.html

PR #76, opened 2025-02-18
Set the working directory of every test pytest "tmp_path" and copy the simulation config file to it.

WeinaJi:
## Review


tests/conftest.py
```diff
@@ -32,3 +34,35 @@ def sonata_config():
...
+    node_sets_file = sim_config_data.get("node_sets_file")
+    if node_sets_file and not os.path.isabs(node_sets_file):
+        sim_config_data["node_sets_file"] = str(src_dir / node_sets_file)
+    for input in sim_config_data.get("inputs", {}).values():
+        spike_file = input.get("spike_file", "")
+        if spike_file and not os.path.isabs(spike_file):
+            input["spike_file"] = str(src_dir / input["spike_file"])
+    if extra_config:
+        sim_config_data.update(extra_config)
+    with open(str(tmp_path / config_file), "w") as dst_f:
```
mgeplf, line 66: `open` accepts `Path` instances, and you can also do `(tmp_path / config_file).open(....)` if you want; I personally prefer the former.

tests/conftest.py
```diff
@@ -32,3 +34,35 @@ def sonata_config():
...
+
+@pytest.fixture
+def create_tmp_simulation_file(request, tmp_path):
+    """ copy simulation config file to tmp_path
+    """
+    params = request.param
+    src_dir = Path(params.get("src_dir"))
+    extra_config = params.get("extra_config")
+    config_file = Path(params.get("simconfig_file"))
+    with open(str(src_dir / config_file)) as src_f:
```
mgeplf, line 52: I wonder if it's worthwhile to have a `load_json` helper, since this pattern gets used a lot in the codebase

tests/integration-e2e/test_coreneuron_directmode.py
```diff
@@ -1,17 +1,27 @@
...
+
+
+@pytest.mark.parametrize("create_tmp_simulation_file", [
+    {
+        "src_dir": str(SIM_DIR / "usecase3"),
+        "simconfig_file": "simulation_sonata_coreneuron.json"
+    }
+], indirect=True)
+def test_coreneuron_no_write_model(create_tmp_simulation_file):
     from libsonata import SpikeReader, ElementReportReader
```
mgeplf, line 18: is it required that these are loaded late? Are we breaking a circular dependency or something?

tests/integration-e2e/test_coreneuron_directmode.py
```diff
@@ -1,17 +1,27 @@
...
     from neurodamus import Neurodamus
     from neurodamus.core.configuration import SimConfig
-    nd = Neurodamus(
-        str(USECASE3 / "simulation_sonata_coreneuron.json"),
-        keep_build=True,
-        coreneuron_direct_mode=True
-    )
+
+    tmp_file = create_tmp_simulation_file
```
mgeplf, line 22: nit: I think it's cleaner if this rename doesn't happen (same below)

tests/integration-e2e/test_coreneuron_directmode.py
```diff
@@ -24,7 +34,9 @@ def test_coreneuron_no_write_model(USECASE3):
                                                                 4.2, 5.5, 7., 7.4, 8.6]))
     npt.assert_allclose(spike_dict["node_ids"][:10], np.array([0, 1, 2, 0, 1, 2, 0, 0, 1, 2]))
 
-    soma_reader = ElementReportReader(SimConfig.reports.get("soma_report").get('FileName'))
+    soma_file = SimConfig.reports.get("soma_report").get("FileName")
```
mgeplf, line 37: Assuming `SimConfig.reports` is a dict, since we don't want the `None` return from `.get`, does it make sense to use `SimConfig.reports["soma_report"]["FileName"]`?

tests/integration-e2e/test_dry_run_workflow.py
```diff
@@ -1,16 +1,27 @@
...
 import pytest
 import os
+import tempfile
 from pathlib import Path
 from unittest.mock import patch
 from neurodamus.utils.memory import DryRunStats
 from neurodamus.core.configuration import GlobalConfig, LogLevel
 
 SIM_DIR = Path(__file__).parent.parent.absolute() / "simulations"
+TMP_FOLDER = tempfile.mkdtemp()
```
mgeplf, line 12: when more tests are added to this file, does this mean that they will use the same test folder? I think that might be confusing.

tests/integration-e2e/test_loadbalance.py
```diff
@@ -191,7 +160,34 @@ def _read_complexity_file(base_dir, pattern, cx_pattern):
             print(f"File not found: {file_path}")
 
 
-def test_loadbal_integration():
+@pytest.mark.parametrize("create_tmp_simulation_file", [
+    {
+        "src_dir": str(SIM_DIR / "usecase3"),
+        "simconfig_file": "simulation_sonata.json",
+        "extra_config": {
+            "network": os.path.join(str(SIM_DIR / "usecase3"), "circuit_config_virtualpop.json"),
```
mgeplf, line 168: maybe:
`str(SIM_DIR / "usecase3" / "circuit_config_virtualpop.json")`?

mgeplf (commented):
I think this is great progress, thanks!
Just some small points I noticed; I'll let others review, but it LGTM.

cattabiani (commented):
Very nice! I have some suggestions

tests/integration-e2e/test_coreneuron_directmode.py
```diff
@@ -24,7 +34,9 @@ def test_coreneuron_no_write_model(USECASE3):
...
+    soma_path = os.path.join(SimConfig.output_root, os.path.basename(soma_file))
```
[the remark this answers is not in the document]
mgeplf: I agree, I think there should also be at ticket for that.

I think the problem is that some places only `str` paths can be used when interacting w/ NEURON (and maybe other libraries).

Maybe the way forward is to use `Path` for all new path handling code?

tests/integration-e2e/test_dry_run_workflow.py
mgeplf: Each invocation of pytest uses its own path for `tmppath` / `tmpdir` fixtures: https://docs.pytest.org/en/stable/how-to/tmp_path.html#temporary-directory-location-and-retention which can be controlled with `--basetemp`.

In this file, however, since it doesn't use pytest's `tmpdir`, it will pick a different location, afaik pytest doesn't override this, so it's not controllable.

I now remember that @WeinaJi mentioned that this had to do w/ how the tests are structured, and inter-depencencies.  I think it would be worthwhile to restructure the tests so it's not the case - but maybe that needs to be a in a different PR.

tests/integration-e2e/test_dry_run_workflow.py
```diff
@@ -1,16 +1,27 @@
...
+
+
+@pytest.fixture(autouse=True)
+def change_test_dir(monkeypatch, autouse=True):
```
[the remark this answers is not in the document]
WeinaJi: Yes, it is overriding here for this test module. As I explained above.

tests/integration-e2e/test_dry_run_workflow.py
```diff
@@ -21,9 +32,16 @@ def neurodamus_instance(request: pytest.FixtureRequest, USECASE3: Path):
     path_to_config = params.get('path_to_config', USECASE3)
     lb_mode = params.get('lb_mode', "")
 
+    with open(str(path_to_config / config_file)) as src_f:
+        sim_config_data = json.load(src_f)
+    sim_config_data["network"] = str(path_to_config / "circuit_config.json")
+    sim_config_data["node_sets_file"] = str(path_to_config / "nodesets.json")
+    with open(str(tmp_path / config_file), "w") as dst_f:
```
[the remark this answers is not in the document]
WeinaJi: Done

tests/integration-e2e/test_coreneuron_directmode.py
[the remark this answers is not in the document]
mgeplf: AFAIK, libsonata will accept `Path`:
```
import libsonata
p =Path('circuit_config.json')
p.exists() # -> returns true
libsonata.CircuitConfig.from_file(p)
```
If there are paths that must be passed as a string, they need to be fixed; please open an issue in those cases.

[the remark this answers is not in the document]
WeinaJi: Ah, I didn't know that. Then I will switch to Path as much as I can.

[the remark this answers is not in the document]
mgeplf: >  Then I will switch to Path as much as I can.

If we can use `Path` for new code, I think that's great - again, let me know if there are issues w/ libsonata.
For old code, I think there needs to be a ticket, and probably more coverage, before we can try and switch everything to `Path` - right now we're in the dark if a Path is used but a str is needed.

tests/conftest.py
```diff
@@ -32,3 +34,35 @@ def sonata_config():
...
+        sim_config_data["node_sets_file"] = str(src_dir / node_sets_file)
+    for input in sim_config_data.get("inputs", {}).values():
+        spike_file = input.get("spike_file", "")
+        if spike_file and not os.path.isabs(spike_file):
+            input["spike_file"] = str(src_dir / input["spike_file"])
+    if extra_config:
+        sim_config_data.update(extra_config)
+    with open(str(tmp_path / config_file), "w") as dst_f:
+        json.dump(sim_config_data, dst_f, indent=2)
+    return str(tmp_path / config_file)
```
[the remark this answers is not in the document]
cattabiani: `create_tmp_config_file`? My point is `simulation` =/= `config`

tests/integration-e2e/test_coreneuron_directmode.py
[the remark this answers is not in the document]
cattabiani: Ok, we go for path. Anybody can close this if it is as soon as it is handled.

cattabiani (approved):
LGTM if the discussions are addressed :D

juanjosegarcan (dismissed):

tests/conftest.py
WeinaJi: I am still thinking about it. A helper that replaces these lines ? Is it really needed? 
```
    with open(str(src_dir / config_file)) as src_f:
        sim_config_data = json.load(src_f)
```

tests/conftest.py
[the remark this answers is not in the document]
WeinaJi: ok, I've changed to `create_tmp_simulation_config_file`

tests/integration-e2e/test_dry_run_workflow.py
WeinaJi: I improved the comment, and will summarize this PR change in one note.

cattabiani (approved):
LGTM!

I just did a full test. The only folders that remain are: `.tox` and `.coverage`. They both can remain IMHO

tests/unit/test_replay_manager.py
```diff
@@ -23,7 +23,7 @@ def replay_sim_config_file(ringtest_baseconfig, input_spikes_file):
         "node_set": "RingB",
     }
 
-    with NamedTemporaryFile("w", suffix='.json', delete=False) as config_file:
+    with NamedTemporaryFile("w", suffix='.json', dir=tmp_path, delete=False) as config_file:
         json.dump(ringtest_baseconfig, config_file)
     yield config_file
     os.unlink(config_file.name)
```
mgeplf, line 29: what happens if there is an exception thrown? should the `os.unlink` be in a `finally`?

tests/conftest.py
mgeplf: No worries, for now, I'll do it in another PR.

mgeplf (approved):

tests/unit/test_replay_manager.py
mgeplf: (this is non-blocking of merge, btw)

WeinaJi: In case of exception thrown during `json.dump`, it may be useful to look at the broken `config_file` for debugging. Right?
It is coded like that in many places. We could discuss and refactor in another PR.

PR #79, opened 2025-02-18
Brute force ringtest data creation

mgeplf:
This is for the two population, version: 
* RingA with 3 cells, RingB with 2
* RingA has 0 -> 1 -> 2 -> 0 ring
* RingB has 0 -> 1 -> 0 ring
* There is one connection from Ring0:0 -> Ring1:0


tests/simulations/ringtest/reference/cellstate_1.json
```diff
@@ -253,15 +253,15 @@
         "A_GABAB": 0.0,
         "A_GABAB_step": 0.9718328750329811,
         "B_GABAA": 0.0,
-        "B_GABAA_step": 0.9415031533855924,
+        "B_GABAA_step": 0.9928825924309503,
```
mgeplf, line 256: this appears to be computed

tests/simulations/ringtest/create_data.py
```diff
@@ -0,0 +1,152 @@
+import h5py
+import libsonata
+import numpy as np
+
+
+def make_nodes():
+    name = "RingA"
+    count = 3
+    with h5py.File("nodes_A.h5", "w") as h5:
+        h5.create_dataset(f"/nodes/{name}/0/model_template", data=["hoc:B_BallStick"] * count)
```
mgeplf, line 10: these are the minimum set of columns that allows the tests to pass

tests/simulations/ringtest/create_data.py
```diff
@@ -0,0 +1,152 @@
...
+        h5.create_dataset(f"/nodes/{name}/0/model_type", data=["biophysical"] * count)
+        h5.create_dataset(f"/nodes/{name}/node_type_id", data=[-1] * count)
+
+        # neurodamus/io/cell_readers.py:140: SonataError
+        h5.create_dataset(f"/nodes/{name}/0/mtype", data=["MIKE"] * count)
+
+        # neurodamus/io/cell_readers.py:162: SonataError
+        h5.create_dataset(f"/nodes/{name}/0/x", data=[0.0] * count, dtype=np.float32)
```
mgeplf, line 18: This looks like it could be a source of optimization; I don't think x/y/z positions are needed other than in the LFP case? So they may be read unnecessarily?

tests/simulations/ringtest/create_data.py
```diff
@@ -0,0 +1,158 @@
...
+# dependencies = ['h5py', 'libsonata', 'numpy']
+# ///
+# the above allows one to run `uv run create_data.py` without a virtualenv
+
+import h5py
+import libsonata
+import numpy as np
+
+
+def make_nodes():
```
WeinaJi, line 12: Are there the same data for NodeA and NodeB ? If so, can we use a common function?

tests/simulations/ringtest/create_data.py
```diff
@@ -0,0 +1,158 @@
...
+    name = "RingA"
+    count = 3
+    with h5py.File("nodes_A.h5", "w") as h5:
+        h5.create_dataset(f"/nodes/{name}/0/model_template", data=["hoc:B_BallStick"] * count)
+        h5.create_dataset(f"/nodes/{name}/0/model_type", data=["biophysical"] * count)
+        h5.create_dataset(f"/nodes/{name}/node_type_id", data=[-1] * count)
```
WeinaJi, line 18: Neurodamus doesn't read these 2 attributes. It might be fine to remove them.

tests/simulations/ringtest/create_data.py
mgeplf: for now, it seems they are, but I'd prefer if they weren't completely alike, so that we know what is contributing to what.
At some point I would like to make helper functions to create all the pieces in a more declarative way, but that will come later, once we have a couple examples of data being created.

tests/simulations/ringtest/create_data.py
mgeplf: `node_type_id` is used by libsonata to determine the size of the population, so we can't remove it.
interesting that `model_type` isn't used; thanks

mgeplf:
One thing that puzzles me, @WeinaJi / @jamesgking is that I haven't provided any `dynamics_params`, but the tests are passing.  Any ideas?

tests/simulations/ringtest/create_data.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # neurodamus/io/cell_readers.py:140: SonataError
+        h5.create_dataset(f"/nodes/{name}/0/mtype", data=["MIKE"] * count)
```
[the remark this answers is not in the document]
mgeplf: Is resource estimation always performed?
If not, it might be worthwhile to only do these loads when they are needed, instead of every time?

WeinaJi:
> One thing that puzzles me, @WeinaJi / @jamesgking is that I haven't provided any `dynamics_params`, but the tests are passing. Any ideas?

Neurodamus reads "threshold_current", "holding_current" when they appear in `node_pop.dynamics_attribute_names`. And `input_resistance` when there is the input pattern of `relative_shot_noise` or `relative_ornstein_uhlenbeck`. I think they are just skipped in the test case.

mgeplf:
> Neurodamus reads "threshold_current", "holding_current" when they appear in node_pop.dynamics_attribute_names. 

Ah, good to know.

For some reason I thought they were required parameters, so I was expecting the loading to fail if they were missing.
I see that a `None` is used in `neurodamus/io/cell_readers.py` if they are missing, what does this do to the configuration of the model?

tests/simulations/ringtest/create_data.py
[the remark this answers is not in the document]
WeinaJi: The resource estimation is by request. But we may need `mtype` in this ring circuit to test the dry run. 

As for the general simulation workflow, I would need @jamesgking here. We have the reading of `mtype` since the legacy time (mvd3). What is that used for? Can we drop it?

tests/simulations/ringtest/create_data.py
mgeplf: I don't see calls to `local_to_global_coord_mapping`, other than in the tests, maybe I am missing something.

WeinaJi (dismissed):
LGTM. I think we can merge this PR (after rebasing). For the attributes `mtype` and locations, we can discuss later whether we need load them when refactoring the code.

tests/simulations/ringtest/create_data.py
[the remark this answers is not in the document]
mgeplf: added a ticket here: https://github.com/openbraininstitute/neurodamus/issues/85

tests/simulations/ringtest/create_data.py
mgeplf: add a ticket here: https://github.com/openbraininstitute/neurodamus/issues/86

cattabiani:
to me it is fine but I would prefer weina to decide since she just wrote the old many.h5 files version

PR #81, opened 2025-02-19
Quell some lints

mgeplf:
* "PT018",  # pytest-composite-assertion
* "C405",   # unnecessary-literal-set
* "ISC003", # explicit-string-concatenation
* "C401",   # unnecessary-generator-set
* "FA100",  # future-rewritable-type-annotation
* "A002",   # builtin-argument-shadowing
* "C416", # unnecessary-comprehension


neurodamus/neuromodulation_manager.py
```diff
@@ -88,8 +88,7 @@ def finalize(
                 if GlobalConfig.debug_conn == [self.tgid]:
                     log_all(
                         logging.DEBUG,
-                        "Neuromodulatory event on tgid: %d, "
-                        + "weights: [%f, %f, %f], nc_type: %d",
+                        ("Neuromodulatory event on tgid: %d, weights: [%f, %f, %f], nc_type: %d"),
```
juanjosegarcan, line 91: Is the parenthesis really needed?

mgeplf: yeah, otherwise the string concat doesn't happen across multiple lines

cattabiani (dismissed):
these are not the formatting I am finding hard to automatically fix but I will not certainly complain for less checks!

pyproject.toml
```diff
@@ -186,9 +183,6 @@ ignore = [
     "S324",   # hashlib-insecure-hash-function
     "S403",   # suspicious-pickle-import
     "S404",   # suspicious-subprocess-import
-    "A002",   # builtin-argument-shadowing
-    "C416",   # unnecessary-comprehension
-    "PT018",  # pytest-composite-assertion
```
WeinaJi, line 191: Are we going to lint also the test scripts?

WeinaJi (dismissed):
LGTM. Thanks. I have just a question, but not blocking.

juanjosegarcan (dismissed):

pyproject.toml
mgeplf: > Are we going to lint also the test scripts?

I'm agnostic to that; historically, I'm less worried about tests being as "lint free" as the other code, but ultimately I think we need to make that call as a group.

neurodamus/neuromodulation_manager.py
mgeplf: Ahem, :facepalm: yes, you're right. I'll fix it.

mgeplf: thanks.

juanjosegarcan (approved):

mgeplf:

mgeplf:

PR #88, opened 2025-02-21
Quell more lints; turn on linting for more ignored ones

mgeplf:
* "PLW1508",# invalid-envvar-default
* "RUF015", # unnecessary-iterable-allocation-for-first-element
* "S108",   # Probable insecure usage of temporary file or directory: "/dev/shm"
* "PLC1901",# compare-to-empty-string
* "PYI024", # collections-named-tuple
* "SIM103", # needless-bool
* "SIM105", # suppressible-exception
* "SIM401", # if-else-block-instead-of-dict-get
* "SIM223", # expr-and-false
* "SIM113", # enumerate-for-loop
* "FURB110",# if-exp-instead-of-or-operator
* "PLW1641",# eq-without-hash
* "PLW3301",# nested-min-max
* "SIM102", # collapsible-if
* "B007",   # unused-loop-control-variable

Allow `fixme` for now


neurodamus/node.py
```diff
@@ -1306,7 +1314,7 @@ def _sim_init_neuron(self):
             Nd.t = -1e9
             prev_dt = Nd.dt
             Nd.dt = fwd_skip * 0.1
-            for flushIndex in range(10):
+            for _flushIndex in range(10):
```
mgeplf, line 1317: this looks very weird to me; we set time to be negative? and then we advance it?

neurodamus/core/_engine.py
```diff
@@ -42,7 +42,7 @@ def find_plugins(cls):
         if plugin_module:
             importlib.import_module(plugin_module)
         # Auto import correctly named modules
-        for finder, name, ispkg in pkgutil.iter_modules():
+        for _finder, name, _ispkg in pkgutil.iter_modules():
```
cattabiani, line 45:
```suggestion
        for _, name, _ in pkgutil.iter_modules():
```

standard way to disregard elements of a tuple

neurodamus/core/nodeset.py
```diff
@@ -221,7 +221,7 @@ def add_gids(self, gids, gid_info=None):
         """Add raw gids, recomputing gid offsets as needed"""
         self._gidvec.extend(gids)
         if len(gids) > 0:
-            self._max_gid = max(self.max_gid, max(gids))
+            self._max_gid = max(self.max_gid, np.max(gids))
```
cattabiani, line 224: is this for efficiency?

neurodamus/ngv.py
```diff
@@ -537,7 +537,7 @@ def _connect_endfeet(self, astro_id):
             astrocyte.create_endfeet(parent_section_ids.size)
 
             # Iterate through endfeet: insert mechanisms, set values and connect to parent section
-            for sec, parent_section_id, l, d, p in zip(
+            for sec, parent_section_id, l, d, _p in zip(
```
cattabiani, line 540:
```suggestion
            for sec, parent_section_id, l, d, _ in zip(
```

neurodamus/node.py
```diff
@@ -440,13 +440,13 @@ def compute_load_balance(self):
         circuit = self._base_circuit
         for name, circuit in self._extra_circuits.items():
             if circuit.get("PopulationType") != "virtual":
+                logging.info("Activating experimental LB for Sonata circuit '%s'", name)
```
cattabiani, line 443:
```suggestion
                logging.info(f"Activating experimental LB for Sonata circuit {name}")
```

I find this easier to read. No problem in any case, this is just personal preference

neurodamus/node.py
cattabiani: I think it is on purpose. Objectives:
- reach steady-state giving some initial time to the simulation to settle
- have the time stamps aligned: we want to skip the initial settling period and start counting time (for the recordings) from our chosen 0 (where steady state is reached)

Solution:
- the simulation starts from a negative time step and proceeds until 0. Recording should be skipped?

I let @WeinaJi correct me if I recall incorrectly.

neurodamus/utils/progressbar.py
```diff
@@ -247,22 +247,22 @@ def _set_progress(self, val):
...
+    for _j in ProgressBar.iter(l1):
         # Do something with j
         time.sleep(0.2)
 
     # Instances can also consume, be reused, and work with sub-selection
     # It is the user responsability to initialize with the right number of iterations
     bar = ProgressBar(15)
     l2 = range(100, 200, 10)
-    for x in bar(l1):
+    for _x in bar(l1):
```
cattabiani, line 258:
```suggestion
    for _ in bar(l1):
```

neurodamus/utils/progressbar.py
```diff
@@ -247,22 +247,22 @@ def _set_progress(self, val):
...
         time.sleep(0.2)
     l2 = list(bar(l2, 5))
     assert l2 == [100, 110, 120, 130, 140]
 
     # Alternativelly, if unknown, use False (creates a spinner)
     bar = ProgressBar(False)
-    for x in bar(l1):
+    for _x in bar(l1):
```
cattabiani, line 265:
```suggestion
    for _ in bar(l1):
```

neurodamus/utils/progressbar.py
```diff
@@ -247,22 +247,22 @@ def _set_progress(self, val):
...
         time.sleep(0.2)
-    for y in bar(range(60)):
+    for _y in bar(range(60)):
```
cattabiani, line 267:
```suggestion
    for _ in bar(range(60)):
```

cattabiani (commented):
a few suggestions and improvements

neurodamus/core/_engine.py
mgeplf: this way it's clear what they are, in case someone wants to use them

neurodamus/core/nodeset.py
mgeplf: either that, or I have to use disable the lint; using the ringtest, I put a break on this, and it was of type np.array, so I think it's safe.  The `max(max.gids, *gids)` alternative seems less safe, and less performant.

neurodamus/node.py
mgeplf: that would case a lint; logging, for efficiency reasons, suggests you pass each of the params; that way the string isn't interpolated unless the logging level is actually using it.

neurodamus/utils/progressbar.py
mgeplf: done

neurodamus/utils/progressbar.py
mgeplf: done

neurodamus/node.py
mgeplf: Ok, I just found it strange; sounds like something that would be helpful to have a comment.
Also, are there particular meanings to -1e9, for instance, how can one tell if the settling will finish in that time?

juanjosegarcan (approved):

neurodamus/node.py
mgeplf: This will be fixed if/when `fwd_skip` is removed

mgeplf:

PR #92, opened 2025-02-25
Add unit test simulation scenario: connection between ringtest populations

### LINKED ISSUE #54: Test simulation scenario: connection between cells, fire AP from pre cell and trigger response in post cell
Add a test where two cells are connected, the presynaptic cell fires (e.g. using IClamp) and a synaptic response is triggered on the post synaptic cell

juanjosegarcan:
## Context
Add a unit test to check the connection between cells in a simulation scenario 

Fixes #54 
## Scope
Based on the ring test simulation circuit 

* [x] Inject a stimulus in presynaptic cells (ringA population) making them fire and check the voltage difference in postsynaptic cells (ringA population)

## Testing
Added unit test test_current_injection.py

## Review


tests/unit/test_current_injection.py
```diff
@@ -0,0 +1,48 @@
...
+
+@pytest.mark.parametrize("create_tmp_simulation_config_file", [
+    {
+        "simconfig_fixture": "ringtest_baseconfig",
+        "extra_config": {
+            "inputs": {
+                "Stimulus": {
+                    "module": "pulse",
+                    "input_type": "current_clamp",
+                    "delay": 1,
```
cattabiani, line 13: I would avoid to use 1s as they are often default values. In addition if something multiplies we cannot distinguish if it multiplies or it replaces:

`a = b` is the same as `a = 1 * b`

cattabiani (dismissed):
1 comment but it is not sooo important to block you. Work on it if you think my suggestion is legit

WeinaJi (approved):
LGTM. If @jamesgking agrees on the new `conductance` value (100), I think we can merge it.

Issue #108, opened 2025-02-26
Fix all documentation lints


mgeplf:
We currently are ignoring:
```
    "D100",   # undocumented-public-module
    "D101",   # undocumented-public-class
    "D102",   # undocumented-public-method
    "D103",   # undocumented-public-function
    "D106",   # undocumented-public-nested-class
    "D107",   # undocumented-public-init
    "D417",   # undocumented-param
```
However, having documentation of public methods would be extremely helpful to those working on neurodamus.

PR #109, opened 2025-02-26
Quell more lints

mgeplf:
Fixed the following, so we don't need to ignore them:
* "PLC2801",# unnecessary-dunder-call
* "FURB103",# write-whole-file
* "FURB122",# for-loop-writes
* "B006", # mutable-argument-default
* "E741",   # ambiguous-variable-name
* "D104",   # undocumented-public-package
* "D212",   # multi-line-summary-first-line
* "PLC0205",# single-string-slots
* "FURB101",# read-whole-file
* "B019", # cached-instance-method
* "A001", # builtin-variable-shadowing
* "PERF102",# incorrect-dict-iterator
* "PLE1141",# dict-iter-missing-items
* "ARG005", # unused-lambda-argument
* "ARG003", # unused-class-method-argument
* "TRY401", # verbose-log-message

Moved ignoring the docs one to a different ticket (https://github.com/openbraininstitute/neurodamus/issues/108) to make this more manageable.


mgeplf:

neurodamus/node.py
```diff
@@ -11,6 +11,7 @@
 from collections import defaultdict
 from contextlib import contextmanager
 from os import path as ospath
```
juanjosegarcan, line 13: Can we not use only Pathlib?

juanjosegarcan (dismissed):

neurodamus/node.py
mgeplf: we could, but I'd prefer to do that all in one go and when we have better tests.
I'm mainly trying to fix lints so that we can turn them on in these PRs.

juanjosegarcan (approved):
LGTM!

PR #112, opened 2025-02-27
Cleanup connection manager

mgeplf:
remove only_gids=None from connect_all since it is never used

* in addition, `connect_all` suggests that it is not filtering


neurodamus/connection_manager.py
```diff
@@ -605,7 +603,7 @@ def connect_group(
             return
 
         for sgid, tgid, syns_params, extra_params, offset in self._iterate_conn_params(
-            src_target, dst_target, mod_override=mod_override
+            src_target, dst_target, show_progress=None, mod_override=mod_override
```
juanjosegarcan, line 606: Can show_progress be removed from the call? by default is None, right?

juanjosegarcan (approved):

neurodamus/connection_manager.py
mgeplf: Yeah, but I like to be explicit and not use the default when I can, it makes refactoring easier, which I suspect this will need.  I haven't quite followed why `None` is used rather than `False`, but I haven't looked too carefully.

PR #116, opened 2025-02-28
Quell more lints

mgeplf:
Cleaned up these lints, so they can be turned on:

* "UP031",  # printf-string-formatting
* "S403",   # suspicious-pickle-import / "S301",   # suspicious-pickle-usage
* "PIE810", # multiple-starts-ends-with 
* "T203",   # p-print 
* "EM103",  # dot-format-in-exception 
* "A005",   # stdlib-module-shadowing 
* "RET503", # implicit-return 

* allow RET504: unnecessary-assign


neurodamus/commands.py
```diff
@@ -71,7 +71,7 @@ def neurodamus(args=None):
 
     if not os.path.isfile(config_file):
         logging.error("Config file not found: %s", config_file)
-        return 1
+        return None
```
juanjosegarcan, line 74: Why here specifically is changed to None but the rest keeps returning 1 when error

mgeplf: that's a mistake, should be `1`; thanks.

Issue #98, opened 2025-02-26
Test for reporting feature


jamesgking:
Create test for reporting membrane variables

In the past we had ASCIIRecord mod file for writing single compartment variables from a cell to a text file
Something similar could be recreated to avoid needing the advanced hdf5 and MPI requirements for sonata reports

WeinaJi:
Hi @jamesgking, 
About the ASCIIRecord mod file, is it https://github.com/BlueBrain/SimulationRunner/blob/master/lib/mod/ASCIIrecord.mod ?

WeinaJi:
After testing with `ASCIIRecord.mod`, I see how the report is written in that way. 
I think it is possible to record the variables in python using hoc vector.record
```
            var_ref = getattr(section(x), "_ref_" + variable_name)
            voltage_vec = Nd.Vector()
            voltage_vec.record(var_ref, rep_dt)
```
, and then dump the `voltage_vec` in a text file.

WeinaJi:
An example of the soma and compartment voltage reports
[compartment.txt](https://github.com/user-attachments/files/19051996/compartment.txt)
[soma.txt](https://github.com/user-attachments/files/19051995/soma.txt)

WeinaJi:
#119  is ready for review.

PR #119, opened 2025-03-03
Unit tests for reporting

WeinaJi:
## Context
For unit tests we don't build libsonata-report, therefore we can't create SONATA reports. Alternatively, we can dump the reporting variables in ASCII format, as per dission in #98.
This PR adds unit tests to check reporting functionalities, mainly `Node.enable_reports()` and `report.py`. For compartment reports, a test dumps the reporting values to ASCII report files and check the values.

## Scope
- Remove the unused legacy parameter `ReportingBufferSize` (https://github.com/openbraininstitute/neurodamus/issues/146) 
- In `Node.enable_reports()`, remove the checks for sonata configuration parameters which should have been checked by libsonata parser.
- In `tests/utils.py`, new functions for recording compartment reports, and write/read the ascii report files. 

## Testing
New `tests/unit/test_ringcells_reports.py` for testing the variables reports for NEURON and CoreNEURON. It also takes over some tests from `tests/integration-e2e/test_reports.py`. 
However, the test for creating synapse report stays in `tests/integration-e2e/test_reports.py` because we can't create it with the current ring circuit (lack of synapses per cell).

## Review


tests/unit/test_ringcells_reports.py
```diff
@@ -0,0 +1,135 @@
...
+        "simconfig_fixture": "ringtest_baseconfig",
+        "extra_config": {
+            "target_simulator": "NEURON",
+            "reports": {
+                "soma_v": {
+                    "type": "compartment",
+                    "cells": "Mosaic",
+                    "variable_name": "v",
+                    "sections": "soma",
+                    "dt": 10,
```
mgeplf, line 75: any particular reason for `dt = 10`?

tests/unit/test_ringcells_reports.py
```diff
@@ -0,0 +1,135 @@
...
+                    "unit": "nA",
+                    "dt": 10,
+                    "start_time": 0.0,
+                    "end_time": 40.0
+                }
+            }
+        }
+    }
+], indirect=True)
+def test_neuorn_report(create_tmp_simulation_config_file):
```
mgeplf, line 101: `test_neuron_report`?

tests/unit/test_ringcells_reports.py
```diff
@@ -0,0 +1,135 @@
...
+    from neurodamus import Neurodamus
+    from neurodamus.core.configuration import SimConfig
+    from neurodamus.core import NeurodamusCore as Nd
+
+    n = Neurodamus(create_tmp_simulation_config_file)
+    assert len(n.reports) == 3
+
+    # For unit tests, we don't build libsonatareport to create the standard sonata reports,
+    # instead we use custom functions to record and write report vectors in ASCII format,
+    # but currently only for comparment reports
```
mgeplf, line 111: maybe we should call this test `test_compartment_report`, and constrain it to only that, because of this limitation?

ie: only do the soma and compartment_report?

mgeplf (commented):
looks great to me, just couple small things

juanjosegarcan (dismissed):
I have no additional comments on the implementation aside from Mike's feedback. However, I thought the issue aimed to include the ASCII reports in neurodamus itself, not just at the test level. Could this be considered for a future update?

tests/unit/test_ringcells_reports.py
WeinaJi: No particular reason, I just want the report data to be shorter. Initially I planned to check the voltage data but gave up in the end as there is no spike.

tests/unit/test_ringcells_reports.py
WeinaJi: ok, I will rename it.

WeinaJi:
> I have no additional comments on the implementation aside from Mike's feedback. However, I thought the issue aimed to include the ASCII reports in neurodamus itself, not just at the test level. Could this be considered for a future update?

As Mike has pointed out, the ASCII report is not going to be official.

tests/unit/test_ringcells_reports.py
cattabiani: you can inject a current to force spikes

WeinaJi: Do you have an example of the stimulus?

cattabiani: there are many in `tests/unit/test_current_injection.py`. I think the simplest one may be:

```
@pytest.mark.parametrize("create_tmp_simulation_config_file", [
    {
        "simconfig_fixture": "ringtest_baseconfig",
        "extra_config": {
            "inputs": {
                "Stimulus": {
                    "module": "linear",
                    "input_type": "current_clamp",
                    "delay": 5,
                    "duration": 1,
                    "node_set": "RingA",
                    "represents_physical_electrode": True,
                    "amp_start": 10
                }
            }
        }
    },
], indirect=True)
```

this is just 1 impulse at 10nA for 1ms after 5ms delay I think

cattabiani: I feel like I am missinterpreting your question, let me know if you wanted something else

WeinaJi: yes, that's what I want. I will try to add it. Thanks!

WeinaJi: I've added the same stimulus as `test_current_injection` and check the soma voltage peak with the same expected result. Can you review again ?

tests/unit/test_ringcells_reports.py
WeinaJi: I've renamed it to be `test_neuorn_compartment_report`.

tests/simulations/ringtest/reference/cellstate_1.json
```diff
@@ -1,7 +1,12 @@
 {
   "B_BallStick": {
     "gid": 1.0,
-    "nSecAll": 3,
+    "nSecAll": 3.0,
```
cattabiani, line 4: do we need nSecAll if it is just the sum of the others? I am wondering because if this number may be extrapolated from the others maybe we should remove it completely. Otherwise we could have wierd bugs down the line

tests/simulations/ringtest/reference/cellstate_1001.json
```diff
@@ -1,7 +1,12 @@
 {
   "B_BallStick": {
     "gid": 1001.0,
-    "nSecAll": 3,
+    "nSecAll": 3.0,
```
cattabiani, line 4: same as before

tests/unit/test_ringcells_reports.py
```diff
@@ -0,0 +1,196 @@
+from pathlib import Path
```
cattabiani, line 1: I was thinking that maybe it would be useful to have a small intro that explains what these tests are about

tests/unit/test_ringcells_reports.py
```diff
@@ -0,0 +1,196 @@
...
+                        "start_time": 0.0,
+                        "end_time": 40.0,
+                    },
+                },
+            },
+        }
+    ],
+    indirect=True,
+)
+def test_neuorn_compartment_report(create_tmp_simulation_config_file):
```
cattabiani, line 124:
```suggestion
def test_neuron_compartment_report(create_tmp_simulation_config_file):
```

tests/unit/test_ringcells_reports.py
```diff
@@ -0,0 +1,196 @@
...
+        rep_type = rep_conf["Type"]
+        if rep_type == "compartment":
+            ascii_recorders[rep_name] = record_compartment_report(rep_conf, n._target_manager)
+    Nd.finitialize()  # reinit for the recordings to be registered
+    n.run()
+
+    # Write ASCII reports
+    for rep_name, (recorder, tvec) in ascii_recorders.items():
+        ascii_report = Path(n._run_conf["OutputRoot"]) / (rep_name + ".txt")
+        write_ascii_report(ascii_report, recorder, tvec)
```
cattabiani, line 147: can we devise also a test where there are no compartments? So we check the edge case.

Also, do you think it would be useful to have this test also run with 2 cores so that the problems that arise when 2 players try to write the same stuff are covered? Following this also what happens if there are 2 cores, one has all the neurons and they try to write the ascii reports?

tests/utils.py
```diff
@@ -258,3 +260,67 @@ def check_signal_peaks(x, ref_peaks_pos, threshold=1):
     """
     peaks_pos = find_peaks(x, prominence=threshold)[0]
     np.testing.assert_equal(peaks_pos, ref_peaks_pos)
+
+
+def record_compartment_report(rep_conf: dict, target_manager: TargetManager):
```
cattabiani, line 265: do you want to keep this only in the tests or do we want to offer this capability also to users? It may be useful. In that case we should stuff this in neurodamus itself no?

tests/simulations/ringtest/reference/cellstate_1.json
WeinaJi: It is the sum of the others, and it is needed for reporting https://github.com/openbraininstitute/neurodamus/blob/87afb1ada216bedd1d6ce48393aa6dae5b9a27d3/neurodamus/target_manager.py#L546
The value is computed inside emodel hoc template, e.g. `B_BallStick.hoc`

tests/unit/test_ringcells_reports.py
WeinaJi: I've added the doctrings for the error handling tests.

tests/unit/test_ringcells_reports.py
WeinaJi: Good spot! Thanks!

tests/unit/test_ringcells_reports.py
WeinaJi: >can we devise also a test where there are no compartments? So we check the edge case.

What is that? A cell with 0 comparment? Soma is also compartment. 

> Also, do you think it would be useful to have this test also run with 2 cores so that the problems that arise when 2 players try to write the same stuff are covered? Following this also what happens if there are 2 cores, one has all the neurons and they try to write the ascii reports?

The ASCII reporter here is not designed for multi-processing, as we don't need that for the official neurodamus run. `libsonata-report` does, but we don't build it for the unit tests.

tests/utils.py
WeinaJi: It is only for testing. Users should only produce SONATA report.

tests/unit/test_ringcells_reports.py
cattabiani: ok, in this case the tests are not needed

tests/simulations/ringtest/reference/cellstate_1.json
cattabiani: can't we calculate it from the other sections in the .hoc file?

WeinaJi: `nSecAll` is provided in all the BBP hoc templates, why can't use it?

cattabiani: I was not aware of that. If it is customary, ok

tests/unit/test_ringcells_reports.py
cattabiani: I meant a docstring at the beginning of the file just to give a rundown of what we want to achieve with this file

cattabiani (dismissed):
one comment but it is non-blocking

neurodamus/core/configuration.py
```diff
@@ -949,13 +949,7 @@ def _check_restore(config: _SimConfig, run_conf):
...
 
 @SimConfig.validator
-def _coreneuron_params(config: _SimConfig, run_conf):
-    buffer_size = run_conf.get("ReportingBufferSize")
-    if buffer_size is not None:
-        assert buffer_size > 0
-        log_verbose("ReportingBufferSize = %g", buffer_size)
-        config.corenrn_buff_size = int(buffer_size)
-
+def _coreneuron_params(config: _SimConfig, _run_conf):
```
cattabiani, line 952: why this change? It is not mentioned in the description or the issue

tests/unit/test_ringcells_reports.py
cattabiani: so, no incipit to explain what we want to do with these tests?

cattabiani (approved):
just 1 question, not blocking

neurodamus/core/configuration.py
WeinaJi: Done, I have clarified in the PR description.

tests/unit/test_ringcells_reports.py
WeinaJi: I thought the filename is self explanatory, plus each test function has its docstrings.

PR #121, opened 2025-03-03
remove unused ConnectionManagerBase::reenable* functions

mgeplf:


cattabiani:
I feel I dunno enough (I do not know anything. it is the first time I see these snippets) and I cannot comment if we need still need this or not. @WeinaJi ?

cattabiani:
(also, a small description or pointing to an issue would be nice)

juanjosegarcan:
LGTM from code pov but as Catta I can not say much more.

cattabiani:
@mgeplf 

Example of questions that may be explained
- what was it use for?
- why is it no more relevant?
- why there were no tests about this?

mgeplf:
I can't answer any of those; as far as I can tell, it was unused since it's never referenced by anything else, which is also why it's not relevant and probably why there were no tests for it.

WeinaJi:
> I can't answer any of those; as far as I can tell, it was unused since it's never referenced by anything else, which is also why it's not relevant and probably why there were no tests for it.

I guess those functions were used by some scientists at the BBP time. There are also functions like `disable_group`, `disable` which seems not tested. 
Is there an issue to track this change? We could discuss further in the ticket.

mgeplf:
> Is there an issue to track this change? We could discuss further in the ticket.

There isn't an issue, I guess I could make one; is there an advantage to that compared to having the discussion here?

cattabiani:
because this discussion will be lost a little no?

mgeplf:
Not sure how it would get lost, but I have made a ticket: https://github.com/openbraininstitute/neurodamus/issues/225

PR #124, opened 2025-03-04
`spont_minis` simple test

### LINKED ISSUE #114: `spont_minis` to be tested
#107 left this parameter untested because this requires an `ad hoc` test

After reviewing this I think this issue is more complex than expected. `spont_minis` is stochastic and should produce events picked from a poisson process. It can be split in 2:

- check that the synapse is triggered at intervals that roughly follow the suggested frequency. Easy to do with 1 run
- chck that the stochastic nature of the release actually follows a poisson process. Still easy to do but it requires multiple runs with different seeds to feed a statistical relevant number of traces in a KS test for example. AFAIK the capabilities to do multiple runs for a single test is not there. I could create a workaround with python subprocess, and an exchange of data with files but it looks overkill and a waste of resources. I will continue on this if required

### LINKED ISSUE #123: `init` vs `sim_init`
`init` calls `sim_init`. It is not clear what should I use when. Is one of them for internal use? May we merge them in one that takes, for example, a parameter to discriminate between the 2 functionalities and explain why and when someone may want to call one or the other? Like this it is very confusing.

cattabiani:
## Context

`spont_minis` uses a stochastic process to generate events. Synapses, if ready, release generating a voltage change. 

## Scope
Proper testing of this requires a stochastic test campaign that is out of the scope of this PR. 

Here I just want to see that I see a voltage change in `RingB` if there is a `spont_minis` attached. Without the `spont_minis` no peaks are recorded. Thus, checking the precise position of the peaks for this `spont_minis` should be sufficient as reproducing this randomly or from other sources is very unlikely.

I am also testing that a pont minis is attached to the synapse and that in the weight it really serves its type as spont_minis

- [x] remove accidental debug code that remained around

## Fix

#114 only partially,

## Additional bug fixes

#123 while working on this we noticed that `sim_init` can be called twice (it has protection) but `init` does not. I added the same protection to `init` too. 

## Review

### DIFF
--- a/tests/unit/test_connection_overrides.py
+++ b/tests/unit/test_connection_overrides.py
@@ -194,42 +194,40 @@ def test_synapse_modoverride(create_tmp_simulation_config_file):
     },
 ], indirect=True)
 def test_spont_minis_simple(create_tmp_simulation_config_file):
-    """Test that spont_mini fires with roughly """
+    """Test spont_minis standard behavior.
+
+    This test verifies that a spont_minis is correctly added to
+    the simulation, preserving essential properties from existing
+    netcons while enforcing expected modifications.
+
+    The test is divided in 2 parts:
+    - check that the spont_minis is added, neuron sees it
+    and the other netcons are unaffected
+    - run a simulation and see spikes at roughly the correct
+    firing rate
+    """
     from neurodamus import Neurodamus
     from neurodamus.core import NeurodamusCore as Ndc
-    from neurodamus.connection import NetConType, SpontMinis
-
-    src_pop, src_gid, tgt_pop, tgt_gid = "RingA", 1, "RingB", 1001
+    from neurodamus.connection import NetConType
 
     nd = Neurodamus(create_tmp_simulation_config_file)
-    edges_a_b = nd.circuits.get_edge_manager(src_pop, tgt_pop)
-    # nd.connections not netcon
-    # this should retrieve still only 1 connection. The
-    # spont_minis attaches an orphan netcon directly to the
-    # synapse. Here we check that the old netcon is there
-    # and the spont_minis is placed
-    connections = list(edges_a_b.get_connections(tgt_gid))
-    assert len(connections) == 1
-    conn = next(iter(connections))
-    # old netcon is still up and active right?
-    assert conn.sgid == src_gid
-    assert conn.tgid == tgt_gid
-    # check that there is also a spont_minis attached
-    # to this synapse
-    assert conn.minis_spont_rate == 200
-    assert isinstance(conn._spont_minis, SpontMinis)
-    assert len(list(conn._spont_minis.netcons)) == 1
-    nc = next(iter(conn._spont_minis.netcons))
-    # check that the netcon associated to the spont_minis
-    # (because spont_minis create ad-hoc netcons when
-    # isntantiated) is aware that it is a spont_minis
-    # and serves that in the weight array
-    assert nc.weight[4] == int(NetConType.NC_SPONTMINI)
+    # get all the netcons targetting 1001 from neuron directly
+    cell = nd._pc.gid2cell(1001)
+    nclist = Ndc.cvode.netconlist("", cell, "")
+    assert len(nclist) == 3
+    # assert that the old netcons are still there
+    assert nclist[0].srcgid() == 1
+    assert nclist[1].srcgid() == 1002
+    # test the spont_minis netcon
+    assert nclist[2].srcgid() == -1
+    assert nclist[2].weight[4] == int(NetConType.NC_SPONTMINI)
+    # weight is set as the weight of the original netcon
+    assert nclist[2].weight[0] == nclist[0].weight[0]
+    # delay is always set to 0.1. Check connection.SpontMinis.create_on
+    assert nclist[2].delay == pytest.approx(0.1)
 
-    # check that the voltage trace has spikes. Since there
-    # are no other stimuli, they are produced by the spont_minis
     voltage_trace = Ndc.Vector()
-    cell_ringB = nd.circuits.get_node_manager(tgt_pop).get_cell(tgt_gid)
+    cell_ringB = nd.circuits.get_node_manager("RingB").get_cell(1001)
     voltage_trace.record(cell_ringB._cellref.soma[0](0.5)._ref_v)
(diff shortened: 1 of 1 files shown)


tests/utils.py
```diff
@@ -7,6 +7,38 @@
 from neurodamus.core.configuration import SimConfig
 
 
+def find_peaks(trace, threshold=-0.05):
```
juanjosegarcan, line 10: we need to decide between this and the ones using scipy in https://github.com/openbraininstitute/neurodamus/pull/127 . If we go for this one I can wait for this merge and adapt my code. 
For the rest LGTM

juanjosegarcan (approved):

tests/utils.py
juanjosegarcan: I would also prefer move from find_peaks to check_peaks and do the assert inside the function.

cattabiani: ah yes, I can wait for your pr to merge and fix mine. I think this one should go second. For me, using the scipy function is better

cattabiani: for the specific point about find_* vs check_* I think I prefer to leave all the checks in 1 function if they are simple enough to write as the function is more general and can be used in more cases but it is just a knitpick

juanjosegarcan: ok, so I'll wait til this is merged and then adapt my PR

cattabiani: what? no, I meant that I wait until yours is merged. You have the new find_peaks

juanjosegarcan: but I'm using check_peaks not find peaks

juanjosegarcan: done

tests/unit/test_connection_overrides.py
```diff
@@ -174,3 +174,41 @@ def test_synapse_modoverride(create_tmp_simulation_config_file):
...
+                    "source": "RingA",
+                    "target": "RingB",
+                    "spont_minis": 200,
+                    "synapse_configure": "%s.verboseLevel = 1"
+                }
+            ]
+        }
+    },
+], indirect=True)
+def test_spont_minis_simple(create_tmp_simulation_config_file):
```
WeinaJi, line 197: Can we check  the netcon object created by `self._spont_minis.create_on()`?  That should be stored in the `Nd.cvode.netconlist`.

cattabiani: the netcon is always there. How do you check if there is a spont minis attached or not?

PS: yes, I know how to retrieve a netcon as I worked on `connection_overrides`

WeinaJi: Maybe something like the test for Replay netcons ? 
https://github.com/openbraininstitute/neurodamus/blob/ae27d40ac931e49cc122353f08d2111a7bb9c772/tests/unit/test_replay_manager.py#L175

cattabiani: it does not work. netcon.weight[4] is 0. Even if clearly the `spont_minis` is attached. How do I know that it works?

because I see peaks with the spont minis and no peaks without

cattabiani: also weight[4] is not documented. I think for the replay in particular it needs to pass something to the netcon and it puts the type there but I do not think it is a common feature in neurodamus

cattabiani: btw, the weight vector is: `[100.0, 0.0, 0.0, 0.0, 0.0]`

neurodamus/connection.py
```diff
@@ -490,6 +490,8 @@ def _init_artificial_stims(self, cell, replay_mode=ReplayMode.AS_REQUIRED):
             self._spont_minis = None
 
         # Delayed vecs: release if not used, sort if over 1 value
+        if self._delay_vec is None and self._delayweight_vec is None:
```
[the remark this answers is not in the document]
cattabiani: done

tests/unit/test_connection_overrides.py
cattabiani: done!

tests/unit/test_connection_overrides.py
```diff
@@ -174,3 +174,76 @@ def test_synapse_modoverride(create_tmp_simulation_config_file):
         for nc_id, nc in enumerate(nclist):
             utils.check_netcon(src_gid, nc_id, nc, edges, selection, **kwargs)
             utils.check_synapse(nc.syn(), edges, selection, **kwargs)
+
+
+def inspect(v):
```
juanjosegarcan, line 179: This function is not used.

cattabiani: ah thanks! yes, it was for debugging

cattabiani: done

juanjosegarcan (approved):

tests/unit/test_connection_overrides.py
```diff
@@ -174,3 +174,65 @@ def test_synapse_modoverride(create_tmp_simulation_config_file):
...
+    conn = next(iter(connections))
+    # old netcon is still up and active right?
+    assert conn.sgid == src_gid
+    assert conn.tgid == tgt_gid
+    # check that there is also a spont_minis attached
+    # to this synapse
+    assert conn.minis_spont_rate == 200
+    assert isinstance(conn._spont_minis, SpontMinis)
+    assert len(list(conn._spont_minis.netcons)) == 1
+    nc = next(iter(conn._spont_minis.netcons))
```
WeinaJi, line 222: How about checking netcon in the neuron cvode context? In that way, we are sure the netcons will be sent to neuron.

cattabiani: the spont mini creates a new netcon that does not have a source. How would you get the netcon from cvode context? I can try, maybe I did not find the correct neuron context yet

WeinaJi: yes, in this case `netcon.srcgid()` shall be negative, @juanjosegarcan  found it out for ReplayStim netcons. We can do similar as 
https://github.com/openbraininstitute/neurodamus/blob/ae27d40ac931e49cc122353f08d2111a7bb9c772/tests/unit/test_replay_manager.py#L175
check `netcon.srcgid()` is negative and `netcon.weight[4] == int(NetConType.NC_SPONTMINI)`, as well as the delay and `netcon.weight[0]`

PR #127, opened 2025-03-04
Adds test for synapse replay with virtual population

### LINKED ISSUE #96: Test using projections
Users create projections which instantiate synapses on neurons while not having a presynaptic cell on any rank

These can be activated by the replay stimulus feature

juanjosegarcan:
## Context
Add a unit test using projections

Fixes #96 
## Scope
Based on the ring test simulation circuit test a replay synapse input changing RingA population type to virtual, testing that way the use of projections

## Testing
Extended and improved test_replay_manager 

## Review

### DIFF
--- a/tests/unit/test_replay_manager.py
+++ b/tests/unit/test_replay_manager.py
@@ -1,52 +1,120 @@
 import pytest
+import numpy as np
 import numpy.testing as npt
-import os
+
+
+from .conftest import RINGTEST_DIR
+
+INPUT_SPIKES_FILE = str(RINGTEST_DIR / "input_spikes.h5")
 
 
 @pytest.fixture
-def input_spikes_file(ringtest_dir):
-    return str(ringtest_dir / "input_spikes.h5")
-
-
-@pytest.fixture
-def replay_sim_config_file(ringtest_baseconfig, input_spikes_file, tmp_path):
+def ringtest_virtual_pop_config():
     from tempfile import NamedTemporaryFile
+    import os
     import json
 
-    ringtest_baseconfig["inputs"] = {}
-    ringtest_baseconfig["inputs"]["spikeReplay"] = {
-        "module": "synapse_replay",
-        "input_type": "spikes",
-        "spike_file": input_spikes_file,
-        "delay": 0,
-        "duration": 50,
-        "node_set": "RingB",
+    circuit_config_data = {
+        "version": 2,
+        "networks": {
+            "nodes": [
+                {
+                    "nodes_file": str(RINGTEST_DIR / "nodes_A.h5"),
+                    "populations": {
+                        "RingA": {
+                            "type": "virtual",
+                        }
+                    }
+                },
+                {
+                    "nodes_file": str(RINGTEST_DIR / "nodes_B.h5"),
+                    "populations": {
+                        "RingB": {
+                            "type": "biophysical",
+                            "morphologies_dir": "swc",
+                            "biophysical_neuron_models_dir": str(RINGTEST_DIR / "hoc"),
+                        }
+                    }
+                }
+            ],
+            "edges": [
+                {
+                    "edges_file": str(RINGTEST_DIR / "local_edges_B.h5"),
+                    "populations": {
+                        "RingB__RingB__chemical": {
+                            "type": "chemical"
+                        }
+                    }
+                },
+                {
+                    "edges_file": str(RINGTEST_DIR / "edges_AB.h5"),
+                    "populations": {
+                        "RingA__RingB__chemical": {
+                            "type": "chemical"
+                        }
+                    }
+                }
+            ]
+        }
     }
 
-    with NamedTemporaryFile("w", suffix='.json', dir=tmp_path, delete=False) as config_file:
-        json.dump(ringtest_baseconfig, config_file)
-    yield config_file
+    with NamedTemporaryFile("w", suffix='.json', delete=False) as config_file:
+        json.dump(circuit_config_data, config_file)
+
+    yield dict(
+        network=config_file.name,
+        node_sets_file=str(RINGTEST_DIR / "nodesets.json"),
+        target_simulator="NEURON",
+        run={
+            "random_seed": 1122,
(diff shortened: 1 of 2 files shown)


tests/unit/test_replay_manager.py
```diff
@@ -1,94 +1,261 @@
...
+    nd.run()
+
+    v_increase_rate = np.diff(voltage_vec, 2)
+    window_sum = np.convolve(v_increase_rate, [1, 2, 4, 2, 1], 'valid')
+    strong_reduction_pos = np.nonzero(window_sum < -0.5)[0]
+    non_consecutives_pos = strong_reduction_pos[np.insert(
+        np.diff(strong_reduction_pos) > 1, 0, True)]
+    expected_positions = np.array([32, 75])
+
+    npt.assert_equal(non_consecutives_pos, expected_positions)
```
cattabiani, line 198: what about putting this in `tests/utils.py`? It is used in various tests. I also have a PR about it if you want to wait

cattabiani, line 198: in addition: why didn't you opt for detecting the peaks in the trace directly using something like `scipy.signal.find_peaks`? Why the second derivative? Why the [1, 2, 4, 2, 1] filtering kernel? 

these are all curiosities that came to me later, after your previous PR

tests/unit/test_replay_manager.py
```diff
@@ -1,94 +1,261 @@
...
```
cattabiani, line 261: the same as before: this may go in a function

cattabiani (commented):
very nice! a few suggestions

tests/unit/test_replay_manager.py
cattabiani: I also see [here](https://nrn.readthedocs.io/en/8.2.6/tutorials/ball-and-stick-4.html) that neuron has already a spike detector. Maybe we can use that

cattabiani: apparently it works only if you have netcons attached to the cell and it is a rudimentary threshold crossing detector. Maybe it is better to have this function instead.

cattabiani: another small thing: the filtering kernel does not have an integral == 1. In fact, this one is equal to 10. This changes the threshold detection (-0.5 I think). We should have a kernel that has an integral equal to 1 so we can swap them without the need to change the thresholds

juanjosegarcan:
@cattabiani I added a check_signal_peaks function based on scipy.signal.find_peaks to tests utils following your suggestions. Check it please and let me know what you think. May we should use it all around the tests?

tests/unit/test_replay_manager.py
```diff
@@ -1,94 +1,249 @@
...
+
+    # get voltage variations in cell 1001
+    cell = nd.circuits.get_node_manager("RingB").get_cell(1001)
+    voltage_vec = Nd.Vector()
+    voltage_vec.record(cell._cellref.soma[0](0.5)._ref_v)
+
+    Nd.finitialize()
+    nd.run()
+
+    expected_positions = np.array([39, 81])
```
cattabiani, line 248: are you sure you need to convert it to a `np.array` explicitly? I think it works with a simple list too

tests/unit/test_test_utils.py
```diff
@@ -114,3 +115,9 @@ def test_check_synapses(create_tmp_simulation_config_file):
     # wrong type
     with pytest.raises(TypeError):
         utils.check_synapses([0], edges, selection)
+
+
+def test_check_signal_peaks():
```
cattabiani, line 120: I am not sure that this function works even if the signal rises or descends. May you test also that case? An inclined trace with some peaks

tests/utils.py
```diff
@@ -186,3 +187,26 @@ def check_synapse(syn, edges, selection, **kwargs):
...
+        taken as reference.
+        threshold: peak detection threshold with values
+        between 0 and 1, measured over the normalized
+        signal.
+
+    Raises:
+        AssertionError: If any of the reference peak
+        positions doesn't match with the obtained peaks
+    """
+    x_norm = (x - x.min()) / (x.max() - x.min())
```
cattabiani, line 209: using prominence is usually good but I am concerned about this norm. May you try the test where you combine a spiking signal with a steady rising one? 

y = spike_signal + a * t

where a is a fixed coefficient. It may happen if we are measuring accumulations of ions or even voltage or current

cattabiani (commented):
I am concerned about an edge case. I would like a test for that case too

tests/unit/test_replay_manager.py
cattabiani: even in that case you can handle the conversion in the function itself no?

cattabiani: this is still missing no?

tests/utils.py
```diff
@@ -237,3 +238,25 @@ def check_synapse(syn, edges, selection, **kwargs):
...
+        respect of the surrounding baseline of the signal
+
+    Raises:
+        AssertionError: If any of the reference peak
+        positions doesn't match with the obtained peaks
+    """
+    x = np.array(x)
+    peaks_pos = find_peaks(x, prominence=threshold)[0]
+
+    np.testing.assert_equal(peaks_pos, ref_peaks_pos)
```
cattabiani, line 262:
```suggestion
    np.testing.assert_equal(peaks_pos, np.array(ref_peaks_pos))
```

cattabiani (commented):
I think 1 change and we are done

tests/utils.py
juanjosegarcan: that's not really needed right?

cattabiani: I did not see the x = np.array(x) before

juanjosegarcan: I removed too

tests/unit/test_test_utils.py
cattabiani: done

cattabiani (approved):
LGTM!

PR #128, opened 2025-03-04
Quell lints

mgeplf:
Clean up
* "SIM108", # if-else-block-instead-of-if-exp
* "SIM115", # open-file-with-context-handler
* "E731",   # lambda-assignment
* "N804",   # invalid-first-argument-name-for-class-method

So they can be enabled

### DIFF
--- a/docs/source/development.rst
+++ b/docs/source/development.rst
@@ -84,7 +84,7 @@ Let's now try to run some tests.
 That looks good. Notice a couple of things:
 
 If you're a developer, you probably want to start with only one MPI rank. Beware of the size of the simulations, though. You probably want to solve a problem with a minimum reproducer having a couple of cells only.
-**"Special available. Replacing binary."** informs us that it detected a "special" binary and is going to use it (instead of attempting to load libnrnmech.so). If you need to disable that, set the env variable: export neurodamus_special=1
+**"Special available. Replacing binary."** informs us that it detected a "special" binary and is going to use it (instead of attempting to load libnrnmech.so). If you need to disable that, set the env variable: export NEURODAMUS_SPECIAL=1
 
 .. warning::
     NOTE: Unfortunately, for practical reasons, loading a model like neurodamus-neocortex will add neurodamus-py from the system to PYTHONPATH. While this isn't fully investigated yet, for the moment, we need to override that entry by pushing our own neurodamus-py to the top of PYTHONPATH.
--- a/neurodamus/commands.py
+++ b/neurodamus/commands.py
@@ -176,7 +176,8 @@ def show_exception_abort(err_msg, exc_info):
         f.write(str(MPI.rank) + "\n")
 
     with open(err_file) as f:
-        line0 = open(err_file).readline().strip()
+        line0 = f.readline().strip()
+
     if str(MPI.rank) == line0:
         logging.critical(err_msg, exc_info=exc_info)
 
--- a/neurodamus/commands.py
+++ b/neurodamus/commands.py
@@ -74,7 +74,7 @@ def neurodamus(args=None):
         return 1
 
     # Shall replace process with special? Don't if is special or already replaced
-    if not sys.argv[0].endswith("special") and not os.environ.get("neurodamus_special"):
+    if not sys.argv[0].endswith("special") and not os.environ.get("NEURODAMUS_SPECIAL"):
         _attempt_launch_special(config_file)
 
     # Warning control before starting the process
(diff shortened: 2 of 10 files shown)


docs/source/development.rst
```diff
@@ -84,7 +84,7 @@ Let's now try to run some tests.
 That looks good. Notice a couple of things:
 
 If you're a developer, you probably want to start with only one MPI rank. Beware of the size of the simulations, though. You probably want to solve a problem with a minimum reproducer having a couple of cells only.
-**"Special available. Replacing binary."** informs us that it detected a "special" binary and is going to use it (instead of attempting to load libnrnmech.so). If you need to disable that, set the env variable: export neurodamus_special=1
+**"Special available. Replacing binary."** informs us that it detected a "special" binary and is going to use it (instead of attempting to load libnrnmech.so). If you need to disable that, set the env variable: export NEURODAMUS_SPECIAL=1
```
cattabiani, line 87: was this needed? It may cause regressions

neurodamus/gap_junction.py
```diff
@@ -85,10 +85,11 @@ def _compute_gj_offsets(self, gj_dir):
...
         gj_sum = 0
 
-        for line in open(gjfname):
-            gj_offsets.append(gj_sum)  # fist gid has no offset. the final total is not used
-            _, offset = map(int, line.strip().split())
-            gj_sum += 2 * offset
+        with open(gjfname) as fd:
+            for line in fd:
+                gj_offsets.append(gj_sum)  # fist gid has no offset. the final total is not used
+                _, offset = map(int, line.strip().split())
```
cattabiani, line 91: didn't you want to opt for `_<name>`?

neurodamus/node.py
```diff
@@ -258,7 +258,7 @@ def read_population_offsets(cls, read_virtual_pop=False):
         return pop_offsets, alias_pop
 
     @classmethod
-    def _pop_offset_file(self, create=False):
+    def _pop_offset_file(cls, create=False):
```
cattabiani, line 261: knitpick: since you changed `self` to `cls` without adding other changes in the class I suspect that this classmethod may be a staticmethod instead

neurodamus/report.py
```diff
@@ -158,12 +158,12 @@ def add_synapse_report(self, cell_obj, point, vgid, pop_name="default", pop_offs
...
-            for synapse in synapse_list:
-                try:
+            try:
+                for synapse in synapse_list:
                     var_ref = getattr(synapse, "_ref_" + variable)
                     self.report.AddVar(var_ref, synapse.synapseID, gid, pop_name)
-                except AttributeError:
-                    raise AttributeError(f"Variable '{variable}' not found at '{synapse.hname()}'.")
+            except AttributeError:
+                raise AttributeError(f"Variable '{variable}' not found at '{synapse.hname()}'.")
```
cattabiani, line 166: why did you invert these 2? Since it raises an exception I think the final effect is the same so I suspect it is just for performance but I prefer to ask

cattabiani (commented):
I have a couple of questions

docs/source/development.rst
mgeplf: I had the same question; since it seems to be a `development` thing, I hope it's safe; I agree with the lint that env vars should be capitalized, which is why I'd like to apply it, but if you think it's safer not to, I'm ok reverting that.

neurodamus/gap_junction.py
mgeplf: yeah, but I don't know enough about this code, I'm just fixing the dangling file pointer

neurodamus/report.py
mgeplf: yeah, it's a perf thing that linting noticed, which is pretty cool, imo.

neurodamus/gap_junction.py
cattabiani: ah c'mon. I think it is safe if you put a new unused variable there with a `_` affix

cattabiani: just this and you get an approval :D

mgeplf: the original code was `_, offset = map(int, line.strip().split())`, so it wasn't new.
I don't know what's supposed to be in `gjinfo.txt` is the problem - it's not documented anywhere that I know of

neurodamus/commands.py
```diff
@@ -113,7 +113,7 @@ def hocify(args=None):
         --output-dir=<PATH>     Output directory for hoc files.
     """
     options = docopt_sanitize(docopt(hocify.__doc__, args))
-    morph_path = abspath(options.pop("MorphologyPath"))
+    morph_path = os.path.abspath(options.pop("MorphologyPath"))
```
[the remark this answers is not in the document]
cattabiani: good catch! +1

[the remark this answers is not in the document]
mgeplf: yeah, probably; I just don't know if `hocify` can handle no string paths, which is why I kept it the same as before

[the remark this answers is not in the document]
mgeplf: (*none string paths)

mgeplf:
replaced w/ https://github.com/openbraininstitute/neurodamus/pull/245

PR #133, opened 2025-03-06
Cleanup layers of indirection in sonata_config.py::SonataConfig

mgeplf:
For historical reasons, the SONATA config was being loaded and translated heavily.

The in this PR include:
* remove the `def __getattr__(self, item)`: this made things difficult to find and debug.
  for instance, if one references `self.network`; this was doing a "translation", and was being returned from
  one of the `_entries`, `_sections`, or, after some manipulation, a `_translation`
* reduce redundant copies of the same information, less likely to modify one, and then another being used:
    ie: removed the fields, since they weren't necessary, and could be accessed from the others:
        _circuit_networks, _config_dir, _config_json, _entries, _resolved_manifest, _sections,
* removed custom implementation of functionality that exists in libsonata; this fixes defaults that were being incorrectly handled (see #217)
* fixed tests that were relying on `v_init` being -65, even though the default should be -80. in SONATA: seehttps://github.com/openbraininstitute/neurodamus/issues/217


WeinaJi:
I see the  `execute_neuron_configures` has been removed by #207. 
@mgeplf , do you still need this PR for other changes? If not, we can close it.

mgeplf:
@WeinaJi > do you still need this PR for other changes? If not, we can close it.

yes, this includes a lot of cleanup of the sonata -> config translation layer that isn't used, and makes debugging what is going on quite complicated.

neurodamus/io/sonata_config.py
```diff
@@ -219,17 +158,19 @@ def _extract_circuits_info(self) -> dict:
...
+        if not simulation_nodeset_name:
             logging.warning("Simulating all populations from all node files...")
-        network = self._circuit_networks
+
+        network = json.loads(self._circuit_conf.expanded_json)["networks"]
 
         def make_circuit(nodes_file, node_pop_name, population_info):
             if not os.path.isabs(nodes_file):
-                nodes_file = os.path.join(os.path.dirname(self.network), nodes_file)
+                nodes_file = os.path.join(os.path.dirname(self._sim_conf.network), nodes_file)
```
WeinaJi, line 169: We could take this change to change to `pathlib.Path`.

neurodamus/io/sonata_config.py
```diff
@@ -274,7 +215,9 @@ def make_circuit(nodes_file, node_pop_name, population_info):
                     ):
                         edges_file = edge_config["edges_file"]
                         if not os.path.isabs(edges_file):
-                            edges_file = os.path.join(os.path.dirname(self.network), edges_file)
+                            edges_file = os.path.join(
+                                os.path.dirname(self._sim_conf.network), edges_file
+                            )
```
WeinaJi, line 220: `pathlib.Path` as well.

neurodamus/io/sonata_config.py
```diff
@@ -18,93 +18,34 @@ class ConnectionTypes(str, Enum):
...
-                section_value, self._resolved_manifest
-            )
+        with open(config_path) as fd:
+            if inputs := json.load(fd).get("inputs", None):
+                self._stable_inputs_order = tuple(inputs.keys())
+            else:
+                self._stable_inputs_order = ()
 
-        self._circuit_conf = libsonata.CircuitConfig.from_file(self.network)
-        self._circuit_networks = json.loads(self._circuit_conf.expanded_json)["networks"]
```
WeinaJi, line 61: I see we load this network in `_extract_circuits_info` and `parsedProjections`. Perhaps we can keep this class attributes.

neurodamus/io/sonata_config.py
```diff
@@ -315,7 +261,7 @@ def parsedProjections(self):
 
                 edges_file = edge_config["edges_file"]
                 if not os.path.isabs(edges_file):
-                    edges_file = os.path.join(os.path.dirname(self.network), edges_file)
+                    edges_file = os.path.join(os.path.dirname(self._sim_conf.network), edges_file)
```
WeinaJi, line 264: same for `pathlib.Path`

neurodamus/io/sonata_config.py
```diff
@@ -380,17 +323,19 @@ def parsedStimuli(self):
...
+
         return stimuli
 
     @property
     def parsedInjects(self):
         injects = {}
         # the order of stimulus injection could lead to minor difference on the results
         # so better to preserve it as in the config file
-        for name in self._sections["inputs"]:
+        for name in self._stable_inputs_order:
```
WeinaJi, line 334: Is `self._stable_inputs_order` going to be replace by the future `libsonata.inputs -> list` ? If so, please leave a TODO comment here.

tests/simulations/ringtest/create_data.py
```diff
@@ -74,6 +74,7 @@ def make_ringtest_nodes():
         "y": it.count(1),
         "z": it.count(2),
         "morphology": "cell_small",
+        "synapse_class": ['EXC', 'EXC', 'EXC']
```
WeinaJi, line 77: Is `synapse_class` required for the ringtest ?

neurodamus/io/sonata_config.py
mgeplf: I actually have a separate PR that fixes a bunch of this; libsonata returns absolute paths, so it's not necessary.

neurodamus/io/sonata_config.py
mgeplf: as above

neurodamus/io/sonata_config.py
mgeplf: I considered that, but I think that reparsing is better compared to keeping redundant information on the class, when it's cheap to parse it.  That way there is less chance we get into a mismatch of what is being read, like we had in the past.

neurodamus/io/sonata_config.py
mgeplf: as above

tests/simulations/ringtest/create_data.py
mgeplf: good point, I think not, as I'm not replacing the h5 files; I wonder where that came from.
I will remove it.

neurodamus/io/sonata_config.py
WeinaJi: I thought we would use pathlib.Path to replace `os.path` everywhere, non? We need to agree on the convention. If you will replace them in another PR, that's also fine.

neurodamus/io/sonata_config.py
WeinaJi: >there is less chance we get into a mismatch of what is being read, like we had in the past.

Did that happen in the past ?

neurodamus/io/sonata_config.py
mgeplf: yeah, we should agree and do it in a different PR, IMO.

neurodamus/io/sonata_config.py
mgeplf: That's one of the things I'm fixing in this commit; "v_init" coming from different places.

neurodamus/io/sonata_config.py
WeinaJi: ok, great!

neurodamus/io/sonata_config.py
WeinaJi: That was due to the `if conditions:` condition IMO. But it seems fine if you keep reparsing.  I will approve this PR.

WeinaJi (approved):
LGTM.

neurodamus/io/sonata_config.py
mgeplf: Right, but the only reason one could do `if conditions` was that there were two copies of the config: the json one and the lisonata one.  IMO, had there been only the libsonata one, it wouldn't have happened.

PR #144, opened 2025-03-11
synapse bluecellulab checks

### LINKED ISSUE #142: Add tests for synapse properties covered by bluecellulab
bluecellulab checks for additional synapse fields

nrrp https://github.com/openbraininstitute/BlueCelluLab/blob/main/tests/test_circuit/test_validate.py

also, there are checks for specific synapse fields

NMDA_ratio, tau_r_NMDA, tau_d_NMDA

https://github.com/openbraininstitute/BlueCelluLab/blob/main/tests/test_circuit/test_circuit_access.py

checks regarding placement of synapses

https://github.com/openbraininstitute/BlueCelluLab/blob/main/tests/test_synapse/test_synapse_factory.py

cattabiani:
## Context

After talking with bluecellulab we decided to be inspired by some of their tests (copy). This PR adds some checks to the synapses and the ways you can override values with the `simulation_config.json` that I was not aware of.

## Scope

- [x] `nrrp` was already checked. Minor improvements: it should be a positive integer
- [x] `"%s.NMDA_ratio = 1.22 %s.tau_r_NMDA = 3.9 %s.tau_d_NMDA = 148.5"`.  
    - [x] `NMDA_ratio`
    - [x] `tau_r_NMDA` and `tau_d_NMDA` are global variables. Added test to override these
- [x] add test to verify synapse location
- [x] test in particular modoverride for `GluSynapse`
- [x] move/merge `test_connection.py` into unit tests
- [x] check synapse override from [conditions](https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#parameters-required-for-modifications) and how it interacts with the one from `connection_overrides.synapse_configure`. Spoiler: ``connection_overrides.synapse_configure` wins

## Additional improvements, not stemming from bluecellulab

- [x] `np.isclose(10001.1, 10001.2) == True`. We should use smaller values. default rtol is 1e-5 
- [x] modoverride test was not really overriding (it was already ampa). Switch to gaba in the test
- [x] fix pytest assert reporting if `tests.utils` is imported in `conftest.py`
- [x] create issue for inconsistend behavior for synapse global variable override #147 
- [x] fix [docs](https://github.com/openbraininstitute/sonata-extension/pull/2)


tests/unit/test_connection.py
WeinaJi, line 1: Could you try to merge `tests/integration-e2e/test_connection.py` into this file? I see that that file doesn't test very advanced features, could use the ringtest circuit as well, and combine with yours.

tests/unit/test_connection_overrides.py
```diff
@@ -140,7 +157,7 @@ def test_synapse_without_weight(create_tmp_simulation_config_file):
                     "name": "A2B",
                     "source": "RingA",
                     "target": "RingB",
-                    "modoverride": "AMPANMDA"
+                    "modoverride": "GABAAB",
```
WeinaJi, line 160: Does the test also pass with `"modoverride": "GluSynapse"` ? 
That's so far the only case that scientists use for 'modoverride', maybe good to test here.

WeinaJi (commented):
>np.isclose(10001.1, 10001.2) == True. We should use smaller values. default rtol is 1e-5

What shall we do? Reduce the `rtol` when comparing big numbers?

cattabiani:
>What shall we do? Reduce the rtol when comparing big numbers?

either that or reduce the numbers

tests/unit/test_connection.py
cattabiani: I was not aware of that test. I will try

tests/unit/test_connection_overrides.py
cattabiani: let me try

cattabiani: not so simple. I get this error:
```
attribute = 'Use_d_TM', optional = False

    def _read(attribute, optional=False):
        if attribute in self._population.attribute_names:
            return self._population.get_attribute(attribute, needed_edge_ids)
        if optional:
            log_verbose("Defaulting to -1.0 for attribute %s", attribute)
            return -1
>       raise AttributeError(f"Missing attribute {attribute} in the SONATA edge file")
E       AttributeError: Missing attribute Use_d_TM in the SONATA edge file

/Users/katta/OBI/neurodamus/neurodamus/io/synapse_reader.py:325: AttributeError
```

If you want I can try to add that attribute to the edge file. What do you think?

WeinaJi: Ah, I see. Because additional fields are requested from the `GluSynapseHelper` https://github.com/openbraininstitute/neurodamus-models/blob/main/common/hoc/GluSynapseHelper.hoc#L12
And they are missing in the ringtest edges file.
Shall we just check an exception is raised in this case with `pytest.raises` ? Then we also cover the functionality of reading additional attributes from hoc file (although it fails).

tests/unit/test_connection.py
cattabiani: done

tests/unit/test_connection_overrides.py
cattabiani: ah this is a good idea, yes

tests/unit/test_connection.py
```diff
@@ -0,0 +1,87 @@
...
+                    "target": "nodesPopB",
+                    "synapse_configure": "%s.dummy=1"
+                }
+            ]
+        }
+    }
+], indirect=True)
+def test_config_error(create_tmp_simulation_config_file):
+
+    with pytest.raises(ConfigurationError):
```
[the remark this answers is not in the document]
cattabiani: added a match with "nodesPopB can't be loaded". Do you think that may work?

tests/utils.py
```diff
@@ -233,13 +233,19 @@ def check_synapse(syn, edges, selection, **kwargs):
...
+    assert np.isclose(syn.Fac,  _get_attr("facilitation_time", kwargs, edges, selection, syn_id))
 
     if _get_attr("n_rrp_vesicles", kwargs, edges, selection, syn_id) >= 0:
-        assert syn.Nrrp == _get_attr("n_rrp_vesicles", kwargs, edges, selection, syn_id)
+        assert np.isclose(syn.Nrrp ,  _get_attr("n_rrp_vesicles", kwargs, edges, selection, syn_id))
+    assert syn.Nrrp > 0
+    assert int(syn.Nrrp) == syn.Nrrp
+
+    # check NMDA ratio if existing
+    if "NMDA_ratio" in kwargs and hasattr(syn, "NMDA_ratio"):
```
WeinaJi, line 247: One last comment: can this exception msg more specific ? Is it 
```Missing attribute Use_d_TM in the SONATA edge file```
raised in the actual run ?

File: tests/unit/test_connection_overrides.py [hunk 4]
```diff
@@ -176,6 +193,33 @@ def test_synapse_modoverride(create_tmp_simulation_config_file):
...
+        },
+    },
+], indirect=True)
+def test_gluSynapse_modoverride(create_tmp_simulation_config_file):
+    """
+    Test modoverride with gluSynapse. It raises an error because of
+    missing data in the edge file
+    """
+    from neurodamus import Neurodamus
+    with pytest.raises(AttributeError, match="Use_d_TM"):
```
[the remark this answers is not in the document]
cattabiani: I did not want to annoy the developer the day they want to change the sentence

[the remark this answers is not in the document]
cattabiani: done

WeinaJi (approved):

PR #156, opened 2025-03-17
Adds test to extracellular calcium conditions.

### LINKED ISSUE #141: Add tests for simulation config settings similar to bluecellulab
https://github.com/openbraininstitute/BlueCelluLab/blob/main/tests/test_circuit/test_simulation_config.py

double check that we have tests checking for simulation config fields

celsius
v_init
extracellular_calcium

juanjosegarcan:
## Context
Add unit test checking extracellular_calcium sim configuration

Fixes #141  

## Scope
* [x] Changes test_sonata_config to use ringtest_circuit instead of usecase3 and reorganize existing tests
* [x] Adds test for extracellular_calcium conditions configuration

## Testing
Improved and extended unit tets test_sonata_config

## Review

### DIFF
--- a/neurodamus/utils/pyutils.py
+++ b/neurodamus/utils/pyutils.py
@@ -28,6 +28,11 @@ def dict_filter_map(dic, mapp):
     return {mapp[key]: val for key, val in dic.items() if key in mapp}
 
 
+def dict_is_subset(dic, subset):
+    """Checks if subset is a subset of the original dict"""
+    return all(not (key not in dic or dic[key] != value) for key, value in subset.items())
+
+
 def docopt_sanitize(docopt_opts):
     """Sanitizes docopt parsed key names"""
     opts = {}
--- a/neurodamus/utils/pyutils.py
+++ b/neurodamus/utils/pyutils.py
@@ -28,11 +28,6 @@ def dict_filter_map(dic, mapp):
     return {mapp[key]: val for key, val in dic.items() if key in mapp}
 
 
-def dict_is_subset(dic, subset):
-    """Checks if subset is a subset of the original dict"""
-    return all(not (key not in dic or dic[key] != value) for key, value in subset.items())
-
-
 def docopt_sanitize(docopt_opts):
     """Sanitizes docopt parsed key names"""
     opts = {}
--- a/tests/unit/test_sonata_config.py
+++ b/tests/unit/test_sonata_config.py
@@ -1,7 +1,7 @@
 import pytest
 from pathlib import Path
 import logging
-from neurodamus.utils.pyutils import dict_is_subset
+from tests import utils
 
 from .conftest import RINGTEST_DIR
 
@@ -41,7 +41,7 @@ def test_parse_run(create_tmp_simulation_config_file):
         'ExtracellularCalcium': 1.2
     }
 
-    assert dict_is_subset(SimConfig.run_conf, expected_conf)
+    utils.check_is_subset(SimConfig.run_conf, expected_conf)
 
 
 @pytest.mark.parametrize("create_tmp_simulation_config_file", [
@@ -96,7 +96,7 @@ def test_parse_conditions(create_tmp_simulation_config_file):
         'randomize_Gaba_risetime': 'False'
     }
 
-    assert dict_is_subset(list(SimConfig._simulation_config.Conditions.values())[0],
+    utils.check_is_subset(list(SimConfig._simulation_config.Conditions.values())[0],
                           expected_conditions)
     assert SimConfig.run_conf["SpikeLocation"] == "AIS"
 
@@ -193,7 +193,7 @@ def test_parse_connections(create_tmp_simulation_config_file):
         "NeuromodStrength": 0.75
     }
 
-    assert dict_is_subset(conn, expected_conn)
+    utils.check_is_subset(conn, expected_conn)
     assert conn.get("SpontMins") is None
     assert conn.get("Modoverride") is None
 
@@ -267,7 +267,7 @@ def test_parse_inputs(create_tmp_simulation_config_file):
         "Delay": 0.,
         "Duration": 10000.0
     }
-    assert dict_is_subset(SimConfig.stimuli["hypamp_mosaic"], expected_input_hp)
+    utils.check_is_subset(SimConfig.stimuli["hypamp_mosaic"], expected_input_hp)
 
     input_RSN = SimConfig.stimuli["RelativeShotNoise_L5E_inject"]
     expected_input_RSN = {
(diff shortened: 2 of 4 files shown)


tests/unit/test_sonata_config.py
```diff
@@ -1,146 +1,106 @@
...
+    {
+        "simconfig_fixture": "ringtest_baseconfig",
+        "extra_config": {
+            "conditions": {
+                "extracellular_calcium": 1.2,
+                "mechanisms": {
+                    "GluSynapse": {
+                        "cao_CR": 2.0,
+                    }
+                }
```
cattabiani, line 53: if the default value of cao_CR is 2.0 you do not need this. If it is not, you should still check that the value is 2.0 and is there. An old bug I solved was that every entry in extra_config was replacing, not merging with the default config.

tests/unit/test_sonata_config.py
```diff
@@ -152,13 +112,28 @@ def test_parse_seeds(create_tmp_simulation_config_file):
     assert SimConfig.rng_info.getIonChannelSeed() == 0
     assert SimConfig.rng_info.getMinisSeed() == 0
     assert SimConfig.rng_info.getSynapseSeed() == 0
-    assert SimConfig.run_conf["SpikeLocation"] == "AIS"
```
cattabiani, line 155: why this? and why a spike location was called "AIS"?

tests/unit/test_sonata_config.py
```diff
@@ -228,3 +267,63 @@ def test_parse_inputs(create_tmp_simulation_config_file):
...
+                    "dt": 0.1,
+                    "start_time": 0.0,
+                    "end_time": 10.0,
+                    "file_name": "my_compartment_report"
+                }
+            }
+        }
+    }
+], indirect=True)
+def test_parse_reports(create_tmp_simulation_config_file):
```
cattabiani, line 300: I feel there must be a better way to compare dicts than hardcoding the checks. If not, we can write a simple function to reuse no?

cattabiani (commented):
some questions

tests/unit/test_sonata_config.py
WeinaJi: BTW, Neurodamus doesn't check the default value of the global variable in the mechansim rather than applying the "new" value directly. It is up to the user to decide which variable and its value to include in the "conditions/mechanisms".

tests/unit/test_sonata_config.py
```diff
@@ -1,146 +1,106 @@
...
+def test_extracellular_calcium_warning(create_tmp_simulation_config_file, caplog):
+    from neurodamus.core.configuration import SimConfig
+    from neurodamus.utils.logging import setup_logging
 
-    # conditions section
-    conditions = list(SimConfig._simulation_config.Conditions.values())[0]
-    assert conditions['init_depleted_ProbAMPANMDA_EMS'] is False
-    assert conditions['minis_single_vesicle_ProbAMPANMDA_EMS'] is True
-    assert conditions['randomize_Gaba_risetime'] == 'False'
+    setup_logging.logging_initted = True
```
WeinaJi, line 62: I am curious, what would happen without this setting ?

tests/unit/test_sonata_config.py
cattabiani: we have already a function to merge dicts. We can devise a function that recursively compares dicts (if it does not exist already) and put it in the utils. If you need a few values you check them directly. If you want dicts to be equal you just run the function. [DeepDiff](https://pypi.org/project/deepdiff/) is an idea for example

cattabiani (commented):
just 1 more comment

tests/unit/test_sonata_config.py
WeinaJi: I guess `capsys` gets the warning msg anyway. In some tests we check that instead. 
But I am fine with this`caplog`. I will close this conversation.

tests/unit/test_sonata_config.py
WeinaJi: Why not `==` ?  That's being used in `compare_json_files`. 
BTW, I am fine with checking the key/value separately in this test. It is clear that what are checked and the expected values.

cattabiani: ok, my bad. I thought the == worked in a different way. Nevermind. I prefer the == among dicts because:

- it is more scalable
- easier to read
- easier to maintain. If we add entries and we compare full dicts this test will fail. If we compare only some selected values we may forget to check them here explicitly

Anyway, I may be convinced to prefer the explicit asserts if you present some compelling reasoning on why it is better

juanjosegarcan: I am fine with this subset approach

neurodamus/utils/pyutils.py
```diff
@@ -28,6 +28,11 @@ def dict_filter_map(dic, mapp):
     return {mapp[key]: val for key, val in dic.items() if key in mapp}
 
 
+def dict_is_subset(dic, subset):
```
cattabiani, line 31: I think you need to add tests for this. For example, this returns False even if I think this is a subset:

```

def dict_is_subset(dic, subset):
    """Checks if subset is a subset of the original dict"""
    return all(not (key not in dic or dic[key] != value) for key, value in subset.items())


d1 = {"a": 1, "b": {"c": {"d": 2, "e": 3}}}
subset = {"a": 1, "b": {"c": {"d": 2}}}

print(dict_is_subset(d1, subset))
```

I suggest to use another approach to reduce duplication of code and make it safer. You may get `dict_is_subset` from `merge_dicts`: 

```
return merge_dicts(dic, subset) == dic
```

juanjosegarcan: I changed it. Fixed merge_dict to not consider numerical types difference as TypeErrors. Also added check_is_subset tests

tests/unit/test_test_utils.py
```diff
@@ -161,11 +162,34 @@ def test_merge_dicts_with_parent_and_child_empty():
...
 
+def test_check_is_subset():
+    dic = {'A': 1, 'B': {'C': 2., 'D': {'E': 3.}}, 'F': 4}
+    subset = {'A': 1}
+    utils.check_is_subset(dic, subset)
+    subset = {'A': 1., 'B': {'C': 2}}
+    utils.check_is_subset(dic, subset)
+    subset = {'A': 1, 'B': {'D': {'E': 3}}}
+    utils.check_is_subset(dic, subset)
+
```
cattabiani, line 178:
```suggestion
    subset = {'A': 1, 'B': {'D': {'E': 3.0}}}
    utils.check_is_subset(dic, subset)

```

tests/unit/test_test_utils.py
```diff
@@ -161,11 +162,34 @@ def test_merge_dicts_with_parent_and_child_empty():
...
+    subset = {'A': 1, 'C': 2}
+    with pytest.raises(AssertionError):
+        utils.check_is_subset(dic, subset)
+    subset = {'A': 1, 'B': {'C': 2, 'D': 3}}
+    with pytest.raises(AssertionError):
+        utils.check_is_subset(dic, subset)
+    subset = {'A': 1, 'B': {'C': 3}}
+    with pytest.raises(AssertionError):
+        utils.check_is_subset(dic, subset)
+
```
cattabiani, line 191:
```suggestion
    subset = {'A': 1, 'B': {'C': 2.1}}
    with pytest.raises(AssertionError):
        utils.check_is_subset(dic, subset)

```

cattabiani (approved):
LGTM! I have 2 additional tests. They pass out of the box but I would prefer to add them to verify the behavior with floats vs ints

cattabiani (approved):

Issue #103, opened 2025-02-26
Refactor `Node.__init__` having an "internal" API


mgeplf:
There is this warning (https://github.com/openbraininstitute/neurodamus/blob/769f000140dd0d26964b8fd19b1b40959b94b6a5/neurodamus/node.py#L353):
```
        # The Recipe being None is allowed internally for e.g. setting up multi-cycle runs
        # It shall not be used as Public API
        if config_file is not None:
```

I think this behavior is used here:
https://github.com/openbraininstitute/neurodamus/blob/769f000140dd0d26964b8fd19b1b40959b94b6a5/neurodamus/node.py#L1620

Instead of doing an in place `Node.__init__` to clear the nodes, perhaps a helper method that does that instead can be used.

PR #158, opened 2025-03-18
`Node.__init__`: split it into `__init__` and `_reset`

cattabiani:
## Context

`__init__` has an internal api where, if the config_file is None it just resets the internal variables. It is done mainly for multicycle runs. This goes against the SRP. In addition, users may still use this function in this way. This may create confusion.

## Scope

`__init__` has been split in `__init__` and `_reset`. `__init__` calls reset under the hood. The main advantages are that:

- `_reset` is clearly marked for internal use only
- `__init__` simply fails if the `config_file` is `None`
- SRP

Added also a nice docstrings for good measure

Fix: #103 

## Testing

Since this is refactoring and `__init__` is already tested this should not alter our tests and no additional tests should be required.

## Review

### DIFF
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -339,7 +339,8 @@ class Node:
             config_file: A Sonata config file
             options: A dictionary of run options typically coming from cmd line
         """
-        assert config_file, "bool(config_file) cannot be False"
+        assert isinstance(config_file, str), "`config_file` should be a string"
+        assert config_file, "`config_file` cannot be empty"
 
         if config_file.endswith("BlueConfig"):
             raise ConfigurationError(
@@ -404,7 +405,6 @@ class Node:
         self._report_list = None
         self._stim_manager = None
         self._sim_ready = False
-        self._jumpstarters = []
         self._cell_state_dump_t = None
         self._bbss = Nd.BBSaveState()
 
@@ -1605,7 +1605,7 @@ class Node:
 
         # Reset vars
         Nd.init()
-        Node._reset(self)
+        self._reset()
 
         # Clear BBSaveState
         self._bbss.ignore()
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -339,58 +339,64 @@ class Node:
             config_file: A Sonata config file
             options: A dictionary of run options typically coming from cmd line
         """
-        if config_file:
-            if config_file.endswith("BlueConfig"):
-                raise ConfigurationError(
-                    "Legacy format BlueConfig is not supported, please migrate to SONATA config"
-                )
-            import libsonata
+        assert config_file, "bool(config_file) cannot be False"
 
-            conf = libsonata.SimulationConfig.from_file(config_file)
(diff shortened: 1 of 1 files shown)


neurodamus/node.py
```diff
@@ -339,58 +339,64 @@ def __init__(self, config_file, options=None):
             config_file: A Sonata config file
             options: A dictionary of run options typically coming from cmd line
         """
-        if config_file:
-            if config_file.endswith("BlueConfig"):
-                raise ConfigurationError(
-                    "Legacy format BlueConfig is not supported, please migrate to SONATA config"
-                )
-            import libsonata
+        assert config_file, "bool(config_file) cannot be False"
```
cattabiani, line 342: should this be a string?

neurodamus/node.py
```diff
@@ -1598,7 +1603,9 @@ def clear_model(self, avoid_creating_objs=False, avoid_clearing_queues=True):
         if not avoid_creating_objs and SimConfig.use_neuron and self._sonatareport_helper:
             self._sonatareport_helper.clear()
 
-        Node.__init__(self, None, None)  # Reset vars  # noqa: PLC2801
+        # Reset vars
+        Nd.init()
+        Node._reset(self)
```
WeinaJi, line 1608: Can we call `self._reset()` ?

neurodamus/node.py
cattabiani: done

neurodamus/node.py
cattabiani: done

neurodamus/node.py
```diff
@@ -1598,7 +1603,9 @@ def clear_model(self, avoid_creating_objs=False, avoid_clearing_queues=True):
         if not avoid_creating_objs and SimConfig.use_neuron and self._sonatareport_helper:
             self._sonatareport_helper.clear()
 
-        Node.__init__(self, None, None)  # Reset vars  # noqa: PLC2801
+        # Reset vars
+        Nd.init()
```
WeinaJi, line 1607: I feel that we don't need `Nd.init()` here. It should be already set in `Node.__init__`. Can you try ?

cattabiani: this moves away from the previous implementation. yes, I can try

juanjosegarcan (dismissed):

neurodamus/node.py
cattabiani: it looks like it works. I will merge if CI passes

WeinaJi (approved):

neurodamus/node.py
```diff
@@ -1598,7 +1603,8 @@ def clear_model(self, avoid_creating_objs=False, avoid_clearing_queues=True):
         if not avoid_creating_objs and SimConfig.use_neuron and self._sonatareport_helper:
             self._sonatareport_helper.clear()
 
-        Node.__init__(self, None, None)  # Reset vars  # noqa: PLC2801
```
mgeplf, line 1601: thanks for fixing this, this line was bugging me so much :)

PR #165, opened 2025-03-20
progressBar tests

cattabiani:
## Context

ProgressBar was untested. Let's move some tests to the unit tests folder. 

This stems from a very evasive bug that I encountered while checking other stuff. I cannot find a way to reproduce the problem in a unit test. I leave the solution but I do not know how to test it. 

I need to close stdout but all my efforts are in vain

## Scope

Progressbar unit tests

## Review

### DIFF
--- a/neurodamus/utils/progressbar.py
+++ b/neurodamus/utils/progressbar.py
@@ -177,6 +177,16 @@ class ProgressBar(Progress):
         self._name = name
         super().__init__(end, start)
 
+    def _write(self, s):
+        """Safe write"""
+        if not self._stream.closed:
+            self._stream.write(s)
+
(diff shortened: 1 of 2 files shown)


juanjosegarcan (approved):

mgeplf:
> This stems from a very evasive bug that I encountered while checking other stuff.

What was the bug? I wonder if using a either tqdm or the one built into click (https://click.palletsprojects.com/en/stable/utils/#showing-progress-bars) would save us from having to maintain this.

cattabiani:
sincerely, I forgot. It was tied to save/restore ... maybe. As I said, I could not reproduce it in a standard unit test

Issue #153, opened 2025-03-17
coreneuron - multistage mode instantiation tests


jamesgking:
CoreNeuron can instantiate the model over multiple cycles to reduce total memory needed at once

Create some tests demonstrate and monitor for any regressions

WeinaJi:
FYI: https://github.com/BlueBrain/neurodamus/wiki/Using-CoreNeuron-in-Practice#multi-step-coreneuron-execution

cattabiani:
I see that `tests/integration-e2e/test_multicycle_runs.py` has already `v5_sonata` and `usecase3`. If you what do you want to test more thorroughly may you specify what does need to be tested? Should I take a file and check its coverage?

cattabiani:
do we want the tests in e2e to be moved to unit?

WeinaJi:
Yes, it would be better.

WeinaJi:
Currently the CoreNEURON multicycle run can not work with CoreNEURON direct mode. Neurodamus switches to file mode automatically with throwing a warning.  

https://github.com/openbraininstitute/neurodamus/blob/2942e785a280b0329f48dec2cceb5e86bcb4cc8a/neurodamus/core/configuration.py#L1120-L1125

In this case, it runs with file mode. Users should get their results without any failure. CoreNEURON direct mode is more an improvement on the I/O for the file system. Mostly, users may not notice any difference.

Question: @jamesgking , is it better to raise an exception or a warning for this case? As we don't have any real usage of the direct mode, I don't know what users would prefer.

PR #166, opened 2025-03-21
multicycle coreneuron tests (WIP) #153

cattabiani:
## Context

TODO

## Scope

TODO

## Testing

TODO

## Review

### DIFF
--- a/tests/unit/test_coreneuron_multicycle.py
+++ b/tests/unit/test_coreneuron_multicycle.py
@@ -1,0 +1,55 @@
+import numpy as np
+import pytest
+from pathlib import Path
+
+
+from ..conftest import SIM_DIR
+
+from pathlib import Path
+
+import numpy as np
+import numpy.testing as npt
+import pytest
+
+from tests.utils import check_directory
+
+from neurodamus import Neurodamus
+from neurodamus.core.configuration import SimConfig
+
+ref_gids = np.array([1, 2, 3, 1, 2, 3, 1, 2, 3])  # 1-based
+ref_timestamps = np.array([5.1, 5.1, 5.1, 25.1, 25.1, 25.1, 45.1, 45.1, 45.1])
+
+
+@pytest.mark.parametrize(
+    "create_tmp_simulation_config_file",
+    [
+        {
+            "simconfig_fixture": "ringtest_baseconfig",
+            "extra_config": {
+                "target_simulator": "CORENEURON",
+                "inputs": {
+                    "pulse": {
+                        "module": "pulse",
+                        "input_type": "current_clamp",
+                        "delay": 5,
+                        "duration": 50,
+                        "node_set": "RingA",
+                        "represents_physical_electrode": True,
+                        "amp_start": 10,
+                        "width": 1,
+                        "frequency": 50,
+                    }
+                }
+            }
+        }
+    ],
+    indirect=True,
+)
+def test_multicycle(create_tmp_simulation_config_file):
+    pass
+    # """TODO
+    # """
+    nd = Neurodamus(create_tmp_simulation_config_file)
+    nd.run()
+
+    # AttributeError: dlsym(0x3065e5280, malloc_trim): symbol not found
--- a/tox.ini
+++ b/tox.ini
@@ -33,6 +33,23 @@ commands =
     {toxinidir}/ci/build_ndcore.sh {toxinidir}/neurodamus/data {toxworkdir}
     sh -c ". {toxworkdir}/.envfile && pytest -x --cov=neurodamus --cov-report=term-missing --forked {posargs:tests/unit}"
 
+[testenv:removeme]
+deps =
+    {[testenv]deps}
+    pytest-cov
+    neuron
+    morphio
+setenv =
+    NEURON_INIT_MPI=0
+    PYTHONPATH={toxinidir}
+allowlist_externals =
+    {toxinidir}/ci/build_ndcore.sh
+    sh
+commands =
+    {toxinidir}/ci/build_ndcore.sh {toxinidir}/neurodamus/data {toxworkdir}
+    sh -c ". {toxworkdir}/.envfile && pytest -x --forked {posargs:tests/unit/test_coreneuron_multicycle.py}"
+
+
 [testenv:integration-e2e]
 # Please you need to spack load neurodamus-models model=neocortex +coreneuron or use an equivalent spack virtual env
 passenv = *
@@ -40,6 +57,7 @@ setenv =
     PYTHONPATH={toxinidir}:{env:PYTHONPATH}
     HOC_LIBRARY_PATH={toxinidir}/neurodamus/data/hoc:{env:HOC_LIBRARY_PATH}
     NEURON_INIT_MPI=1
+
 commands =
     python -c "import os; print(os.environ.get('HOC_LIBRARY_PATH', ''))"
     pytest -s -x --forked --durations=5 --durations-min=15 {posargs:tests/integration-e2e}


tests/unit/test_coreneuron.py
```diff
@@ -194,3 +193,46 @@ def test_coreneuron_keep_build(create_tmp_simulation_config_file):
...
+                        "width": 1,
+                        "frequency": 50,
+                    }
+                }
+            }
+        }
+    ],
+    indirect=True,
+)
+def test_coreneuron_multicycle(create_tmp_simulation_config_file):
```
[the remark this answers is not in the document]
cattabiani: done

neurodamus/node.py
```diff
@@ -1342,9 +1342,10 @@ def _coreneuron_ensure_all_ranks_have_gids(self, corenrn_data):
             return
 
         # create a fake node with a fake population "zzz" to get an unused gid.
-        # It only works with a single cycles without enough cells, but in
-        #  _instantiate_simulation we made sure we limit the number of cycles
+        # coreeuron fails if this edge case is reached multiple times as we
+        # try to add twice the same gid. pop "zzz" is forbidden
```
WeinaJi, line 1346: typo: coreeuron -> coreneuron

neurodamus/node.py
```diff
@@ -1865,50 +1866,54 @@ def _coreneuron_restore(self):
...
-        # Keep the initial RSS for the SHM file transfer calculations
-        self._initial_rss = SHMUtil.get_node_rss()
-        print_mem_usage()
-
-        self.load_targets()
-
-        # Check connection block configuration and raise warnings for overriding
-        # parameters
-        SimConfig.check_connections_configure(self._target_manager)
+    @property
```
WeinaJi, line 1869: Why use @property? Here it does not simply get the value of an attribute, and n_cycles is not an attribute of the class either.

You could call `n_cycles()` later on.

neurodamus/node.py
```diff
@@ -1342,9 +1342,10 @@ def _coreneuron_ensure_all_ranks_have_gids(self, corenrn_data):
             return
 
         # create a fake node with a fake population "zzz" to get an unused gid.
-        # It only works with a single cycles without enough cells, but in
-        #  _instantiate_simulation we made sure we limit the number of cycles
+        # coreeuron fails if this edge case is reached multiple times as we
+        # try to add twice the same gid. pop "zzz" is forbidden
         log_verbose("Creating fake gid for CoreNeuron")
+        assert not PopulationNodes.get("zzz")
```
WeinaJi, line 1348: With the fix of `max_cell_count`, I think we will add once the fake population `zzz`. Is this `assert` to protect in case a real population `zzz` exists ?

neurodamus/node.py
```diff
@@ -1949,6 +1954,26 @@ def _instantiate_simulation(self):
...
+        self._initial_rss = SHMUtil.get_node_rss()
+        print_mem_usage()
+
+        self.load_targets()
+
+        # Check connection block configuration and raise warnings for overriding
+        # parameters
+        SimConfig.check_connections_configure(self._target_manager)
+
+        self._multicycle_build_model()
```
WeinaJi, line 1975: Is there a better name than `multicycle_build_model` ? Because the normal flow, the single cycle is also covered in this process. It sounds like neurodamus always does multicycle modeling.

tests/unit/test_coreneuron_directmode.py
```diff
@@ -194,3 +193,46 @@ def test_coreneuron_keep_build(create_tmp_simulation_config_file):
...
+                        "width": 1,
+                        "frequency": 50,
+                    }
+                }
+            }
+        }
+    ],
+    indirect=True,
+)
+def test_coreneuron_multicycle(create_tmp_simulation_config_file):
```
WeinaJi, line 223: Is this test already in test_coreneuron_multicycle.py ?

tests/unit/test_coreneuron_multicycle.py
```diff
@@ -0,0 +1,87 @@
+"""Test coreneuron simultion solver and direct mode
+In file emode, spikes dat file is created by CoreNEURON, check the file content
+In direct mode, spikes vectors are returned to NEURON, no dat file created, check the spikes vector
```
WeinaJi, line 3: Docstrings not valid for this test.

neurodamus/node.py
cattabiani: done

neurodamus/node.py
cattabiani: yes. Do you think it is too much? The additional cost is negligible since we do this only at the beginning

WeinaJi: The assert is fine to be here. Maybe in the docstrings, `pop "zzz" is forbidden` => `pop "zzz" is forbidden if already existing`  ?

neurodamus/node.py
cattabiani: what about:
- `multicycle_build_model` -> `build_model`
- `build_model` -> `build_single_model`

tests/unit/test_coreneuron_directmode.py
cattabiani: obsolete code, removed. Thx!

neurodamus/node.py
cattabiani: Tristan always wanted properties for functions that return a value and have no inputs. I have no strong opinion

cattabiani: done

tests/unit/test_coreneuron_multicycle.py
cattabiani: typo. Changed

neurodamus/node.py
WeinaJi: Yes, it is ok to be brief. 
But what does the single sentence`pop "zzz" is forbidden` mean here?

neurodamus/node.py
WeinaJi: That sounds better. Please also remember to adjust the relevant docstrings.

neurodamus/node.py
```diff
@@ -1865,50 +1866,54 @@ def _coreneuron_restore(self):
...
                 n_cycles = max_num_cycles
+        return n_cycles
+
+    def _multicycle_build_model(self):
+        """Build the model iteratively over multiple cycles if required
 
+        Note: only relevant for coreNeuron
+        """
+        
+        n_cycles = self.n_cycles()
```
WeinaJi, line 1907: I notice the attribute `self._n_cycles` which was forgotten in the pervious developments (don't know when it started).  It should be `self._n_cycles = self.n_cycles()`. 
The original commit is here
https://github.com/openbraininstitute/neurodamus/commit/6745c25ace8edf63045991d1d09e61e7fb12688d#diff-b3e8ca5ec7029b6ed9b0261d7de7443c95ac4e0a0e2e879ece996c35084798e7

cattabiani: done

neurodamus/node.py
cattabiani: changed

neurodamus/node.py
```diff
@@ -1865,50 +1866,54 @@ def _coreneuron_restore(self):
...
-        if SimConfig.use_coreneuron and cell_count / n_cycles < MPI.size and cell_count > 0:
+        if SimConfig.use_coreneuron and max_cell_count / n_cycles < MPI.size and max_cell_count > 0:
             # coreneuron with no. ranks >> no. cells
-            # need to assign fake gids to artificial cells in empty threads during module building
-            # fake gids start from max_gid + 1
+            # need to assign fake gids to artificial cells in empty threads
+            # during module building fake gids start from max_gid + 1
             # currently not support engine plugin where target is loaded later
-            max_num_cycles = int(cell_count / MPI.size) or 1
+            max_num_cycles = int(max_cell_count / MPI.size) or 1
```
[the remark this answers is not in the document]
cattabiani: coreneuron fails if there are no cells in a cycle in a rank. We prevent the fail once but if this happens more than once we fail. Thus, the danger is to go up. 1 cycle is always ok, more can be problematic. Thus, we want to round the number of cycles down, not up

tox.ini
```diff
@@ -40,6 +40,7 @@ setenv =
     PYTHONPATH={toxinidir}:{env:PYTHONPATH}
     HOC_LIBRARY_PATH={toxinidir}/neurodamus/data/hoc:{env:HOC_LIBRARY_PATH}
     NEURON_INIT_MPI=1
+
```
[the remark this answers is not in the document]
cattabiani: typo

neurodamus/node.py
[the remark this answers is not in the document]
juanjosegarcan: Thanks for the clarification

neurodamus/node.py
cattabiani: done

juanjosegarcan (dismissed):

neurodamus/node.py
```diff
@@ -1851,61 +1864,73 @@ def _coreneuron_restore(self):
...
             if n_cycles > max_num_cycles:
                 logging.warning(
                     "Your simulation is using multi-cycle without enough cells.\n"
                     "  => Number of cycles has been automatically set to the max: %d",
                     max_num_cycles,
                 )
                 n_cycles = max_num_cycles
+        self._n_cycles = n_cycles
+        return None
```
WeinaJi, line 1903: It might be better to still `return n_cycles` here, so that it is consistent with the Line 1857.  We can set `self._n_cycles = self.compute_n_cycles()` in `_build_model`.  What do you think ?

cattabiani: ah yes, I did not notice line 1857

WeinaJi (approved):
LGTM

Issue #170, opened 2025-03-24
dry run tests


jamesgking:
Look at code path when using dry run more and decide tests

PR #184, opened 2025-03-28
Expands unit test dry run

juanjosegarcan:
## Context
Expand the coverage of  dry_run by unit tests 

Fixes https://github.com/openbraininstitute/neurodamus/issues/170

## Scope
Check the memory usage stats produced by dry_run based on Ringtest circuit and test cell distribution for the same circuit

## Testing
Extend unit tets test_dry_run

## Review


tests/integration-e2e/test_dry_run_workflow.py
```diff
@@ -77,6 +69,7 @@ def test_dry_run_workflow(neurodamus_instance):
     isMacOS = platform.system() == "Darwin"
     from neurodamus.utils.memory import export_allocation_stats
     from neurodamus.utils.memory import export_metype_memory_usage
+    from tests.utils import defaultdict_to_standard_types
```
cattabiani, line 72: you can import it only once at the top of the page no? It is no neurodamus import

cattabiani: (I will skip this message for all the other imports)

tests/unit/test_dry_run.py
```diff
@@ -3,6 +3,98 @@
 import numpy.testing as npt
 import unittest.mock
 
+from .conftest import RINGTEST_DIR
+
+
```
cattabiani, line 8: I think dry_run should fail/raise an error when we try to use it with an ngv test. May you add this test?

cattabiani, line 8: we should also test with multiple ranks. If you do not know how to do this already may you open at least a new issue to expand on this in a later stage? I have the same problem and I have a lead on where to check (HBP_STEPS) to find out how to do this with pytest. I just did not have time yet

cattabiani (commented):
nice PR! a few suggestions are coming

tests/unit/test_dry_run.py
juanjosegarcan: I run a test with dry_run option with the ngv circuit ("tests/simulation/ngv") and it worked. I don't know what was the issue with ngv! Can you or @WeinaJi give me any hint?

juanjosegarcan: I added a new issue to address it in the future

tests/utils.py
```diff
@@ -62,6 +62,14 @@ def merge_vals(k, parent: dict, child: dict):
     return {k: merge_vals(k, parent, child) for k in set(parent) | set(child)}
 
 
+def defaultdict_to_standard_types(obj):
```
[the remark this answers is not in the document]
juanjosegarcan: done

tests/integration-e2e/test_dry_run_workflow.py
juanjosegarcan: done

tests/unit/test_dry_run.py
cattabiani: Oh, I just remember that it was not working with ngv at all. From what I recall from the discussions with Antonio the problem was that the astrocytes were not accounted for. Our ngv test is not testing dry run in any way. Even if we were testing that the circuit is ... broken so I am not sure of the results. To be on the safe side I would:
- throw an error to prevent dry_run with ngv completely
- add an issue to re-enable it once we have time

cattabiani: ... or did you try the ngv circuit and check if the results are indeed correct?

cattabiani: because I suspect dry-run (maybe) works but it returns bad results

tests/unit/test_test_utils.py
```diff
@@ -167,6 +168,19 @@ def test_merge_dicts_deeply_nested():
...
+    dic = defaultdict(lambda: defaultdict(list))
+    dic["A"]["B"].append(1)
+    dic["A"]["C"] = [2, 3]
+    dic["D"] = [4]
+    dic["E"]["F"] = 5
+    dic["E"]["G"] = "6"
+
+    expected = {"A": {"B": [1], "C": [2, 3]}, "D": [4], "E": {"F": 5, "G": "6"}}
+
+    assert utils.defaultdict_to_standard_types(dic) == expected
```
cattabiani, line 181: this is not really testing all the capabilities of the function because the input dict is already composed of standard types. I would create a custom object for example, usable inside a dict, iterable (with an underlying list) but not a list and check if the result is composed by standard types.

I dunno, I do not see the point of this test. If you think it makes sense may you add a docstrings that explains what features of the function are really tested?

cattabiani (commented):
I still have 2 questions that I think need your attention

tests/unit/test_dry_run.py
juanjosegarcan: I added a ticket to address it once we have a unit test ngv circuit

tests/unit/test_test_utils.py
juanjosegarcan: I added CusutomList object to test the conversion to standard types.

tests/unit/test_dry_run.py
juanjosegarcan: May throw an error when trying dry run over ngv circuit is a good idea. What do you think @WeinaJi ?

WeinaJi: Yes, you could throw an error before loading the cells. I see `self._is_ngv_run` is set at the early stage.
Maybe you can make a dry run test with the ngv config file `tests/simulations/ngv/simulation_config.json` and check `pytest.raises`. If we don't create ngv cells before the exception, it should be ok to run ngv circuit with neocortex models.

tests/unit/test_test_utils.py
```diff
@@ -167,6 +168,31 @@ def test_merge_dicts_deeply_nested():
...
+            return repr(self._items)
+
+    dic = defaultdict(lambda: defaultdict(list))
+    dic["A"]["B"] = CustomList([1, 2])
+    dic["C"] = CustomList(["x", "y"])
+    dic["D"]["E"] = 3
+
+    expected = {"A": {"B": [1, 2]}, "C": ["x", "y"], "D": {"E": 3}}
+
+    assert utils.defaultdict_to_standard_types(dic) == expected
```
cattabiani, line 193: perfect! thank you!

cattabiani (approved):
LGTM

tests/integration-e2e/test_dry_run_workflow.py
WeinaJi, line 1: Do you think that we can make the `test_dynamic_distribute` with the ringtest circuit tests/unit/test_dry_run.py? In that case, we will cover the test for `lb_mode=memory`.

juanjosegarcan: done

tests/unit/test_dry_run.py
juanjosegarcan: done

tests/unit/test_dry_run.py
```diff
@@ -2,6 +2,126 @@
 import numpy as np
 import numpy.testing as npt
 import unittest.mock
+import platform
+from pathlib import Path
+
+from tests.utils import defaultdict_to_standard_types
+from .conftest import RINGTEST_DIR
+
+NGV_DIR = Path(__file__).parent.parent.absolute() / "simulations" / "ngv"
```
cattabiani, line 11: knitpick: we could more this in convtest.py as we do for RINGTEST no?

tests/unit/test_dry_run.py
```diff
@@ -2,6 +2,126 @@
 import numpy as np
```
cattabiani, line 2: do you think that this covers everything that was in `integration-e2e/test_dry_run` plus more?

cattabiani (approved):
one knitpick + a question but not blocking

tests/unit/test_dry_run.py
juanjosegarcan: done

tests/unit/test_dry_run.py
juanjosegarcan: yes, all cases plus more checks are included

WeinaJi (approved):
Looks good! Thanks!

Issue #150, opened 2025-03-13
`NeuroGlialConnection.finalize` may be improved a lot for multi-rank runs


cattabiani:
I do not have all the data to judge this at 100% but following the comments `finalize` is a first draft for a multi-rank solution: slow but safe. We may improve this

WeinaJi:
We need to decide whether we will complete the "TODO" or stay with the fallback.
```
        # TODO: Currently it receives the base_connections object to look (bin search)
        # for the sinapse to attach to. However since target cells and Glia might be
        # distributed differently across MPI ranks, this is bound to work in a single rank.
        # For the moment we fallback to using the original synapse id.
```

https://github.com/openbraininstitute/neurodamus/blob/main/neurodamus/ngv.py#L297-L304 
And the synapse parameter is out of date if we select `USE_COMPAT_SYNAPSE_ID=False`.

evakennyobi:
Can you update the start / end dates please @cattabiani 
Or I put to the backlog if not planned to commence soon

PR #187, opened 2025-03-31
Removes obsolete PopulationID from BlueConfig connections

### LINKED ISSUE #173: configuration.py/connection_manager.py: obsolete PopulationID for BlueConfig
PopulationID

juanjosegarcan:
## Context
Since BlueConfig format is not supported anymore a clean up of deprecated functionality is needed

Fixes #173 

## Scope
Remove obsolete field PopulationID from BlueConfig

## Testing
Check all unit test still working

## Review

### DIFF
--- a/neurodamus/connection_manager.py
+++ b/neurodamus/connection_manager.py
@@ -1167,17 +1167,15 @@ class ConnectionManagerBase:
 # ##############
 
 
-def edge_node_pop_names(edge_file, edge_pop_name, src_pop_name=None, dst_pop_name=None):
-    """Find the node populations names from several edge configurations
+def edge_node_pop_names(edge_file, edge_pop_name, src_pop_name=None, dst_pop_name=None) -> tuple:
+    """Find the node populations names.
 
     Args:
         edge_file: The edge file to extract the population names from
         edge_pop_name: The name of the edge population
-    Returns: tuple of the src-dst population names. Any can be None if not available
+    Returns: tuple of the src-dst population names.
     """
-    src_dst_pop_names = _edge_meta_get_node_populations(
-        edge_file, edge_pop_name
-    ) or _edge_to_node_population_names(edge_pop_name)
+    src_dst_pop_names = _edge_meta_get_node_populations(edge_file, edge_pop_name)
     if src_dst_pop_names:
         if src_pop_name is None:
             src_pop_name = src_dst_pop_names[0]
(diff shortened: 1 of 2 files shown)


neurodamus/connection_manager.py
```diff
@@ -352,27 +351,24 @@ def _init_conn_population(self, src_pop_name, pop_id_override):
...
         cur_pop.virtual_source = (
-            self._src_cell_manager.is_virtual
-            or src_pop_name != self.src_node_population
-            or (bool(pop_id_override) and not src_pop_name)
+            self._src_cell_manager.is_virtual or src_pop_name != self.src_node_population
         )
         logging.info("Loading connections to population: %s", cur_pop)
 
-    def _compute_pop_ids(self, src_pop, dst_pop, src_pop_id=None):
+    def _compute_pop_ids(self, src_pop_name, dst_pop_name):
```
cattabiani, line 361: so src_pop_id was never used in the tests? Is there a test for this function?

cattabiani (approved):
LGTM

If the tests pass it is good for me. A question but it is not blocking

neurodamus/connection_manager.py
```diff
@@ -1128,12 +1124,12 @@ def finalize(self, base_seed=0, sim_corenrn=False, *, _conn_type="synapses", **c
         n_created_conns = 0
 
         for popid, pop in self._populations.items():
-            attach_src = pop.src_id == 0 or not pop.virtual_source  # real populations
+            attach_src = pop.src_pop_id == 0 or not pop.virtual_source  # real populations
```
WeinaJi, line 1127:
```suggestion
            attach_src = not pop.virtual_source  # real populations
```
I think we can simply here. Only read populations requires the netcon to attach to src cell.

juanjosegarcan: Sure, I just let it there, cause looks more related with default population that we agreed to remove in a later PR right?

WeinaJi: Ok then, just don't forget that 👍

WeinaJi (approved):
LGTM

neurodamus/connection_manager.py
```diff
@@ -386,7 +382,7 @@ def select_connection_set(self, src_pop_id, dst_pop_id):
         return self._cur_population
 
     # -
-    def get_population(self, src_pop_id, dst_pop_id=0):
+    def get_population(self, src_pop_id, dst_pop_id):
```
WeinaJi, line 385: I see it is called in ngv.py 
```
base_connections=base_manager.get_population(0)
```
but it looks suspicious that we use `0` as input pop id. Let me check in detail.

juanjosegarcan: I think this is more related with default pop. Coming in the next PR.

WeinaJi: https://github.com/openbraininstitute/neurodamus/blob/main/neurodamus/ngv.py#L297-L304
In fact, `base_connections` is not used because of `USE_COMPAT_SYNAPSE_ID = True`. But SONATA populations will not have the pop_id=0 so it is wrong to `base_connections=base_manager.get_population(0)` in the original code. The change here of removing the default value of dst_pop_id will break the ngv case. Can we keep `dst_pop_id=0` at the moment? We need to review the ngv part , I will open a ticket.

juanjosegarcan: done

WeinaJi: Thanks. FYI, I added a remark for the ngv todo in https://github.com/openbraininstitute/neurodamus/issues/150#issuecomment-2778779604

WeinaJi (approved):

Issue #197, opened 2025-04-09
`dry_run` is polluting `tests/simulations/ringtest`


cattabiani:
with the latest update the new `dry_run` test pollutes creating this folder:

```
tests/simulations/ringtest/output/
```

this should not happen

PR #201, opened 2025-04-10
Fix test dry_run polluting ringtest folder

juanjosegarcan:
## Context
For #197, this PR just create a temporary simulation_config file in the temporary folder to let ringtest folder clean

## Scope
Every neurodamus initialization without dry_run activated generates "output/population_offset.dat" file in the same folder of the simulation config file by default. To solve the pollution of ringtest folder a temp simulation_config file is created in the tmp test folder 

## Testing
Fix test_dry_run

## Review
* [ ] Unit test added

### DIFF
--- a/tests/unit/test_dry_run.py
+++ b/tests/unit/test_dry_run.py
@@ -98,11 +98,16 @@ def test_dry_run_distribute_cells():
     Path(("allocation_r2_c1.pkl.gz")).unlink(missing_ok=True)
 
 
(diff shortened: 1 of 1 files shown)


juanjosegarcan:
This is just a quick fix for the ringtest folder pollution, but maybe to be check the fact that by default we are damping this file to the location of simulation_config file. IMO, may it would be a good option tos set OutputRoot to a temp folder when not define by the configuration.

cattabiani (approved):
LGTM

WeinaJi (approved):

Issue #194, opened 2025-04-07
Pytest using parallel capabilities (nrank > 1)


jamesgking:
Find a method to run pytest where we exercise parallel features with neurodamus and tox

cattabiani:
just as a memo: there are 2 runs (no asserts inside, just checking that the thing works) in simulation_test.yaml. They are executed only in the CI

PR #213, opened 2025-04-22
mpi tests (proof of concept)

cattabiani:
## Context
We want to test neurodamus also with mpi #194. However, this turns out to be trickier than expected. There are multiple libraries around with pros and cons. The following list is not exaustive; it is just what I have found so far and my opinion after bashing my head on all these options:

- **pytest-mpi**: it is very non-intrusive. It just makes pytest aware that is run in mpi. The standard way to run it is `mpiexec -n 2 pytest --with-mpi ...`. The problem is that this interpheres with `--forked`. Either the processes are not forked or the mpi fails. In addition, it is ugly to see messages from various ranks interphere with each other. I did not investigate further
- **mpi-pytest**: different from pytest-mpi, newer, it makes pytest call `mpiexec` on each test that requires that. A typical call is `pytest tests/unit`. Inside `mpi-pytest` calls `mpiexec`. Unfortunately it works only with MPICH as we easily get nested `MPI_Init` calls. Changing the code to address this problem is too much IMHO
- **HBP_STEPS**: the simulator is more different than expected from our case. Over there we use `ctest` to discover the tests and we use `unittest` to test. Our problems are effectively avoided but we cannot easily replicate the same behavior in out setup.
- **pytest-isolate-mpi**: test suite to run programs with mpi in isolated environments. Ufortunately, if we do not run with mpi, the suite does not run the tests forked and the 2 libraries (`pytest-forked` and `pytest-isolate-mpi` interphere with each other when neuron and its mechanisms are involved). (chosen solution)
- **in house**: creating our own test suite is a burden and becomes also complex to maintain. We can tailor it exactly to our needs

## Scope

I picked `pytest-isolate-mpi` as it provides already a lot of what we want. The problem is that it requires a slightly different setup than what we have in the unit tests (for example). 

- [x] add a new tox environment where we can require `pytest-isolate-mpi` 
- [x] add tests to check standard MPI capabilities
- [x] add tests to check MPI + neurodamus capabilities (check also that the processes are forked) 
- [x] move `ringtest_baseconfig` to base `conftest.py`
- ~~check if and how `--basetemp` works~~ EDIT: `basetemp` is ignored when `pytest-isolate-mpi` is employed. We can still set `TMPDIR` in `setenv` in `tox.ini`
- ~~make it work out of the box even with our base setup (unit, etc)~~ EDIT: `pytest-isolate-mpi` discards `--forked` when no mpi command is employed. It is likely a bug. The workaround is to mark all the functions with `@pytest.mark.mpi(ranks=1)`. However, this is labour-intensive. Probably having 2 setups is simpler
- [x] coverage
- [x] merge #222 from @WeinaJi 
- [x] add test with empty rank (1 cell, 2 ranks) 

## Testing

Added some sample tests that check that mpi processes are properly forked and repeatible. Saving files to possibly catch race conditions too


tests/unit-mpi/test_base.py
```diff
@@ -0,0 +1,108 @@
...
+                    "dt": 1,
+                    "start_time": 0.0,
+                    "end_time": 40.0,
+                },
+            },
+        }
+    }
+], indirect=True)
+@pytest.mark.mpi(ranks=[1, 2, 4])
+def test_neurodamus(create_tmp_simulation_config_file, mpi_ranks, mpi_tmp_path):
```
WeinaJi, line 103: We can add checks here:
1. assert the local gids spread over all ranks, e.g.
```
node = n.circuits.get_node_manager("RingA")
local_gids = node.local_nodes.final_gids()
npt.assert_allclose(local_gids, ref_gids) 
```
`ref_gids` are different in each rank. 
2. check the MPI size
```
from neurodamus.core import MPI
assert MPI.size == mpi_ranks
```
3. with ranks = [4] for the coreneuron simulation,  a fake cell is expected to be added in rank id 3, so if we run  ```n = Neurodamus(create_tmp_simulation_config_file, keep_build=True)```, we should see `2004_1/2/3).dat` in the `coreneuron_input` folder.

WeinaJi: And what is `mpi_tmp_path` ? I see the `pytest-isolate-mpi` docs says it wraps the pytest `tmp_path` , are they the same value? So our `SimConig.output_root` `is_relative_to(mpi_tmp_path)` ?

WeinaJi (commented):
A big step forward! Thanks for the studies.
A general question, the `unit-mpi` test is not run in any GH actions workflow, shall we added it in `simulation_test.yml` ?

WeinaJi:
Actually I can run `unit-mpi` with a similar tox setting as `unit`:
```
[testenv:unit-mpi]
deps =
    {[testenv]deps}
    pytest-cov
    -r tests/unit-mpi/requirements.txt
    neuron-nightly
setenv =
    PYTHONPATH={toxinidir}:{env:PYTHONPATH}
    NEURON_INIT_MPI=1
allowlist_externals =
    {toxinidir}/ci/build_ndcore.sh
    sh
commands =
    sh -c ". {toxworkdir}/.envfile && pytest -x --cov=neurodamus --cov-report=term-missing {posargs:tests/unit-mpi}"
```

Then my questions are 
1. Can we run `unit` tests under pytest-isolate-mpi , with a global setting `pytest.mark.mpi(ranks=1)` in `conftest.py` ? 
2. Can we add `unit-mpi` in `coverage.yml` and publish the coverage data ?

tests/unit-mpi/test_base.py
```diff
@@ -0,0 +1,108 @@
...
+
+comm = MPI.COMM_WORLD
+rank = comm.Get_rank()
+size = comm.Get_size()
+
+
+@pytest.mark.mpi(ranks=2)
+def test_mpi_send_recv(mpi_ranks):
+    """Test basic MPI send/receive functionality."""
+    assert size == 2, "This test requires at least 2 MPI processes"
```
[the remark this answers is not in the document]
cattabiani: with the first solution there is still an hardcoded value no?
yes for the second part

tests/unit-mpi/test_base.py
cattabiani: they are not the same value since it fails if I sue the old one. I think `pytest-isolate-mpi` adds a folder on the way for collecting the same mpi run

cattabiani:
@WeinaJi 
1 it is actually harder than that because we also need to put mpi_ranks in the input of the function...
2 yes, I will try

cattabiani:
somehow this works:
```
PYTHONPATH=/home/runner/work/neurodamus/neurodamus:
```
and this does not:
```
PYTHONPATH=/home/runner/work/neurodamus/neurodamus:/home/runner/work/neurodamus/neurodamus/nrn/build/install/lib/python:
```

mgeplf:
> and this does not:

Wow, interesting.  Sounds suspiciously like whatever forks are happening between pytest and MPI, the environment is getting trashed.

It's been on my list for a while, but I've been confused as to why we need to modify the `PYTHONPATH` - part of the point of `tox` is for the package under test to be installed a (somewhat) hermetic virtualenv; why it's either not being installed right or, for some other reason, it's very strange that tinkering w/ the `PYTHONPATH` is required.

neurodamus/core/coreneuron_configuration.py
```diff
@@ -326,6 +326,7 @@ def psolve_core(self, coreneuron_direct_mode=False):
         # Model is already written to disk by calling pc.nrncore_write()
         coreneuron.skip_write_model_to_disk = True
         coreneuron.model_path = f"{self.datadir}"
+        logging.warning("AAAAAAAAA CoreNEURON simulation starting")
```
mgeplf, line 329: prob. old debug string?

tests/unit-mpi/test_base.py
```diff
@@ -0,0 +1,120 @@
...
+            },
+        }
+    }
+], indirect=True)
+@pytest.mark.mpi(ranks=2)
+def test_neurodamus(create_tmp_simulation_config_file, mpi_ranks):
+    """Test Neurodamus/neuron with and without MPI."""
+    from neurodamus import Neurodamus
+    from neurodamus.core import MPI
+    # from neurodamus.core.configuration import SimConfig
```
mgeplf, line 109: can be deleted?

tox.ini
```diff
@@ -33,6 +33,25 @@ commands =
     {toxinidir}/ci/build_ndcore.sh {toxinidir}/neurodamus/data {toxworkdir}
     sh -c ". {toxworkdir}/.envfile && pytest -x --cov=neurodamus --cov-report=term-missing --forked {posargs:tests/unit}"
 
+[testenv:unit-mpi]
+# Please you need to spack load neurodamus-models model=neocortex +coreneuron or use an equivalent spack virtual env
+; passenv = *
```
mgeplf, line 38: needed?

neurodamus/core/coreneuron_configuration.py
cattabiani: ah yes. thank you

tests/unit-mpi/test_base.py
cattabiani: yes. I was testing if coreneuron was failing or not

tests/unit-mpi/test_base.py
```diff
@@ -0,0 +1,120 @@
...
+                    "start_time": 0.0,
+                    "end_time": 40.0,
+                },
+            },
+        }
+    }
+], indirect=True)
+@pytest.mark.mpi(ranks=2)
+def test_neurodamus(create_tmp_simulation_config_file, mpi_ranks):
+    """Test Neurodamus/neuron with and without MPI."""
```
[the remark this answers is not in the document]
cattabiani: I had it like that because I was testing both. However, your solution is clearly more explicit. Accepted

tox.ini
cattabiani: no, again debug code. I was sloppy in removing stuff around sorry

mgeplf: np, just wanted to make sure, happens to everyone

tests/unit-mpi/test_base.py
WeinaJi, line 1: Can you add a CoreNEURON test with 1 cell in 2 ranks? That will cover the creation of fake cell in the empty rank.
You will need to create a new node set of 1 cell for either RingA or RingB.

cattabiani: if you want, yes, I can try. However this was a proof of concept not a pr to test specific parts of the code.

cattabiani: done. May you check that is what you had in mind?

tests/unit-mpi/test_base.py
```diff
@@ -0,0 +1,146 @@
...
+    {
+        "simconfig_fixture": "ringtest_baseconfig",
+        "extra_config": {
+            "target_simulator": "CORENEURON",
+            "node_set": "RingA_oneCell",
+        }
+    }
+], indirect=True)
+@pytest.mark.mpi(ranks=2)
+def test_empty_rank_with_neuron_and_coreneuron(create_tmp_simulation_config_file, mpi_ranks):
```
WeinaJi, line 138: It is fine to test only with CoreNEURON. But you can also include NEURON.
In addition, can you also check the fake cell `dat` file existing in the `coreneuron_input` folder ? I guess the fake cell file should be named as `1002_1.dat`. You will need to enable the option `keep_build=True`.

cattabiani: done. I noticed that each rank gets its own cwd. Thus, only rank 0 has the build folder. For save/restore it should still be ok because we can provide the restore path. What do you think?

I would not add an mpi test for save/restore in this pr. I think it goes out of scope... as you prefer tho. I can also add that if you think it should be in the same pr that introduces mpi to our framework

WeinaJi: Another question: why only rank0 writes? The fake cell files should be written in rank1.

cattabiani: each rank gets a different folder for cwd. I noticed that in coreneuron, only rank 0 writes the folder `build`. The fake cell files are in the folder for rank 0

cattabiani: it is a "feature" of pytest-isolate-mpi. I tried to remove it but I do not get the mpi or the forking in that case

cattabiani: I checked and these are my conclusions (I am not 100% sure that there isn't something else afoot):
- neurodamus is importend. It gets cwd and sets `CoreNeuron.datadir`
- `pytest-isolate-mpi` changes cwd (one per rank) and starts the simulations
- all the ranks write in the folder for rank 0 (as coreneuron would want to do)
- if you assert if a file is there, it will do it in its own, changed, cwd. Thus we should check files only in rank 0

I will add this in the comments as it is interesting to retain

WeinaJi (approved):
Thanks for the investigations and the implementations !

PR #227, opened 2025-04-29
improve `ringtest/create_data.py`

cattabiani:
## Context
Often, while developing, you need a ringtest that is slightly different than the base one. More cells, more connections etc. I went forward and tweaked slightly `create_data.py`

## Improvements
- Connections are not hardcoded. We can change the number of cells in ringA for example and have them still in a ring
- We do not need to specify as many mtypes and stypes as the number of cells (useful for big rings). The code will just cycle over a pattern
- moved the most common inputs above so they are easy to find
- if `Edges.connections` is an int the cells are connected in a ring pattern. Otherwise you can still specify the connections with a list

## Testing
already tested

## Review


juanjosegarcan (dismissed):
LGTM

tests/simulations/ringtest/create_data.py
```diff
@@ -59,18 +69,13 @@ class SonataAttribute:
 EDGE_TYPES = {attr.name: attr for attr in EDGE_TYPES}
 
 
-def _expand_values(attr, value, count):
+def _expand_values(value, count):
     if isinstance(value, str):
-        ds_value = [value] * count
-    elif isinstance(value, Sequence):
```
mgeplf, line 65: this is to handle the fact that there were special values required for, `"syn_type_id": [60, 104, 77]`; ie that the second value was > 100 and thus an excitatory synapse.

If you're making the ring variable, IMO, it's less clear where the excitatory synapses should be.  Right now they would be `Inh, Exc, Inh, Inh`, in the case of PopA if the count is increased to 4, and `Inh, Inh, Inh` for PopB.

tests/simulations/ringtest/create_data.py
```diff
@@ -1,24 +1,34 @@
-#!/bin/env python
```
mgeplf, line 1: please don't remove this

tests/simulations/ringtest/create_data.py
cattabiani: I do not see the problem. You write I, E, I and I am going to cycle over that as much as I need to. What is unclear? What would you envision instead?

mgeplf: > You write I, E, I and I am going to cycle over that as much as I need to. What is unclear?

It's unclear in the sense that the total length is determined by the count, but in a completely different place a position sensitive list exists; hence the check on the length to minimize errors.

> Also, notice that if you provide the correct-size vector it will place the values in the correct places, as before.

Yes, however, it's length is not checked, and that's an easy mistake to make.

> It is not feasible to write by hand an array of 60k elements

Something like this works: `[29, 129, 29]*60_000`, or you can use something from iterator; it's just that it's explicit in that case.

tests/simulations/ringtest/create_data.py
```diff
@@ -1,24 +1,42 @@
...
-
 @dataclass
 class Edges:
     src: str
     tgt: str
     type: str
-    connections: list[(int, int)]
+    connections: Union[int, list[tuple[int, int]]]
 
+    def __post_init__(self):
```
mgeplf, line 36: instead of the variable behavior based on the argument type, why not just create the Edges w/ `connections=[(i, (i + 1) % n) for i in range(n)]` in the caller (ie: line 204)?
That way it stays simple dataclass, and it's up to the caller to make the connections they want?

tests/simulations/ringtest/create_data.py
```diff
@@ -109,6 +122,7 @@ def make_edges(filename, edges, wanted_attributes):
         target_node_count=max(tgt_ids) + 1,
     )
 
+
 def make_lfp_weights():
```
mgeplf, line 126: I think if the ring size is going to be a parameter, if the LFP is going along with it, it should probably accept the same parameter

juanjosegarcan: Not necessarily, not all the neurons in the population have to be in the LFP file. And we have fix values inside to check in the tests. I think we can keep it simple and not overengineering for something we are using in one test

tests/simulations/ringtest/create_data.py
```diff
@@ -109,7 +122,8 @@ def make_edges(filename, edges, wanted_attributes):
         target_node_count=max(tgt_ids) + 1,
     )
 
-def make_lfp_weights():
+
+def make_lfp_weights(ringA_count, ringB_count):
```
juanjosegarcan, line 126: If we want to make lfp procedural, we need to generate the scaling factors based on the ring_count. If we want to go this way I would propose also to pass a list of tuples with (pop_name, count)

mgeplf:
I'm fine w/ the change; other than the `def __post_init__(self):` part, which builds in assumptions about connectivity.

PR #228, opened 2025-04-29
fix coveralls race condition

cattabiani:
## Context
coveralls sometimes fails because of race conditions:
https://github.com/openbraininstitute/neurodamus/actions/runs/14730918755/job/41344729187?pr=227

## Scope
We should do only one coverall after combining the coverages.
Fix: #229 (second option: the total is automatic in the publishing stage)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -15,31 +15,58 @@ jobs:
       fail-fast: false
 
     steps:
+      - uses: actions/checkout@v4
 
-    - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
 
-    - uses: actions/setup-python@v5
-      with:
-        python-version: '3.9'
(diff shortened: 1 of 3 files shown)


cattabiani:
let me know if you prefer solution 2: parallel runs, separate reports, one publishing step still (remove race condition)

juanjosegarcan (dismissed):
LGTM

mgeplf:
Any idea what changed?

Before, I thought, the `flag-name` allowed these to be separate, which was also a requiement; by combining them, there is no point to having the `baseline`, since the `unit` will import everything in `baseline`.

Is it addition of macos that causes the race?

cattabiani:
@mgeplf ok, I think I can implement option 2 in this case. I think it is a race condition. Sometimes it works, sometimes it fails

mgeplf:
if it's not too hard, alternately, we could maybe drop the baseline?

mgeplf (approved):

PR #239, opened 2025-05-05
SpikeManager: remove legacy support

WeinaJi:
## Context
As SONATA SPEC defines, the input spikes file for replay should be the SONATA .h5 format. This PR removes the support for legacy .bin and .dat files in `SpikeManager` class.

## Scope
Remove the relevant code in `replay.py`

## Testing
Add tests in `tests/unit/test_replay_manager.py`

## Review

### DIFF
--- a/neurodamus/replay.py
+++ b/neurodamus/replay.py
@@ -1,5 +1,7 @@
 """Stimulus implementation where incoming synaptic events are replayed for a single gid"""
 
+from pathlib import Path
+
 from .core.configuration import ConfigurationError
 from .utils.multimap import GroupedMultiMap
 from .utils.timeit import timeit
@@ -37,7 +39,7 @@ class SpikeManager:
             filename: path to spike out file. Interpret as binary or ascii according to extension
             delay: delay to apply to spike times
         """
-        if filename.endswith(".h5"):
+        if Path(filename).suffix == (".h5"):
             tvec, gidvec = self._read_spikes_sonata(filename, population)
         else:
             raise ConfigurationError("Spikes input should be a SONATA h5 file")
--- a/neurodamus/replay.py
+++ b/neurodamus/replay.py
@@ -1,5 +1,6 @@
 """Stimulus implementation where incoming synaptic events are replayed for a single gid"""
 
+from .core.configuration import ConfigurationError
 from .utils.multimap import GroupedMultiMap
 from .utils.timeit import timeit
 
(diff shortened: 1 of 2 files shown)


neurodamus/replay.py
```diff
@@ -42,17 +37,10 @@ def open_spike_file(self, filename, delay, population=None):
             filename: path to spike out file. Interpret as binary or ascii according to extension
             delay: delay to apply to spike times
         """
-        # determine if we have binary or ascii file
-        # TODO: filename should be able to handle relative paths,
-        # using the Run.CurrentDir as an initial path
-        # _read_spikes_xxx shall return numpy arrays
         if filename.endswith(".h5"):
```
cattabiani, line 40:
```suggestion
        if Path(filename).suffix == ".h5":
```

so that we move to using paths instead of strings... dunno if you want to add it. I am just presenting an option

cattabiani (dismissed):
LGTM. Just a suggestion, non-blocking

neurodamus/replay.py
WeinaJi: Good point. Thanks.  I've just changed that.

tests/unit/test_replay_manager.py
```diff
@@ -84,7 +85,7 @@ def ringtest_virtual_pop_config():
 
 @pytest.mark.forked
 def test_sonata_spikes_reader():
-    from neurodamus.replay import SpikeManager, MissingSpikesPopulationError
+    from neurodamus.replay import MissingSpikesPopulationError, SpikeManager
```
mgeplf, line 88: why do these need to be imported in a local function?  Is there state associated with them?

WeinaJi: Because they are used for this test. Do you prefer to import at the module level ?

mgeplf: It's the general consensus that they should be at the module level, instead of a function (ie: in pep8 and pylint both flag them), unless there is a circular import, or other reason (which would probably have a comment attached)

Maybe this is one of the things that we can do in another PR, globally, but it would make life easier if we didn't add to it.

WeinaJi: ok, I've moved all the imports at the module level except the NEURON wrapper.

@juanjosegarcan  @cattabiani , please be aware of that in your ongoing/future PRs. In prinple, all imports should be at the module level **except**  
```
    from neurodamus.core import NeuronWrapper as Nd
```
As we run all the tests in forked processes,  import of `from neurodamus import Neurodamus` can be done at the module level too.

mgeplf (approved):

tests/unit/test_replay_manager.py
mgeplf: thanks!

WeinaJi: Thank you for pointing it out @mgeplf . 
We didn't force that rule for the test files before, as we saw the strange neuron behaviors over multiple tests even in forked processes. But now it gets clear that we just need to import the neuron wrapper at function level.

mgeplf: > But now it gets clear that we just need to import the neuron wrapper at function level.

That's something I'm hoping to find time to look at, so we could at least make that safe and not have to worry about strange behaviors.

PR #245, opened 2025-05-07
Quell lints: N804, E731, SIM115, PERF203

mgeplf:
* "SIM108", # if-else-block-instead-of-if-exp
* convert `classmethod` to `staticmethod` since it does not use `cls`
* "PERF203",# try-except-in-loop
* "SIM115", # open-file-with-context-handler
* "E731",   # lambda-assignment
* "N804",   # invalid-first-argument-name-for-class-method


neurodamus/stimulus_manager.py
```diff
@@ -654,7 +652,7 @@ def draw_already_used_numbers(self, rng, dt):
         tstep = self.duration - dt
 
         while prev_t < self.delay - dt:
-            if prev_t + tstep < self.delay - dt:
+            if prev_t + tstep < self.delay - dt:  # noqa: SIM108
```
juanjosegarcan, line 655: I would prefer to put it this way: next_t = min(prev_t + tstep, self.delay - dt)

mgeplf: good call, will do

juanjosegarcan (approved):
LGTM

PR #247, opened 2025-05-07
Cleanup More lints

mgeplf:
* remove improperly used NotImplemented: it is a "A special value which should be returned by the binary special methods"; see documentation: https://docs.python.org/3.12/library/constants.html#NotImplemented
* PLR6301",# no-self-use
* "S404",   # suspicious-subprocess-import 
* "G004",   # logging-f-string
* N818",   # error-suffix-on-exception-name
* "PLR0913",# too-many-arguments: remove layer, fullmtype, combo_name from METypeItem; these are not being used anywhere; this also saves memory since many synapses are created
* "PLR0917",# too-many-positional-arguments


neurodamus/connection_manager.py
```diff
@@ -330,7 +330,8 @@ def _init_conn_population(self, src_pop_name):
         )
         logging.info("Loading connections to population: %s", cur_pop)
 
-    def _compute_pop_ids(self, src_pop_name, dst_pop_name):
+    @staticmethod
```
cattabiani, line 333: I do not see the reason of this change. Static method means that you can use the method without instantiating the class. The method is `_*` meaning that it should be used only inside the class (internal use). For what case is this relevant apart from the fact that we do not need to pass self and is cleaner?

neurodamus/core/_shmutils.py
```diff
@@ -43,7 +43,7 @@ def __set_node_info(MPI):  # TODO: Replace with MPI SHM communicator
         SHMUtil.node_id = MPI.py_scatter(node_info, 0)
         SHMUtil.nnodes = MPI.py_broadcast((node_info[-1] + 1) if MPI.rank == 0 else 0, 0)
 
-        subprocess.call(["/bin/rm", "-rf", shmdir])
+        shutil.rmtree(shmdir, ignore_errors=True)
```
cattabiani, line 46: @WeinaJi  suggested to leave this as it was. She said that it is faster than calling shutil.rmtree. You can add a comment prevent future developers to change this or you can argue with her

neurodamus/gap_junction.py
```diff
@@ -99,7 +99,8 @@ def finalize(self, *_, **_kw):
                     self.seclamp_current_per_gid_recorder[gid] = Nd.h.Vector()
                     self.seclamp_current_per_gid_recorder[gid].record(seclamp._ref_i)
 
-    def _finalize_conns(self, _final_tgid, conns, *_, **_kw):
+    @staticmethod
```
cattabiani, line 102: same as before

neurodamus/io/sonata_config.py
```diff
@@ -417,10 +417,12 @@ def parsedModifications(self):
...
 
-    def _dir(self, obj):
+    @staticmethod
+    def _dir(obj):
         return [x for x in dir(obj) if not x.startswith("__") and not callable(getattr(obj, x))]
 
-    def _adapt_libsonata_fields(self, rep):
+    @staticmethod
+    def _adapt_libsonata_fields(rep):
         for key in rep:
```
cattabiani, line 426: the same as before

neurodamus/io/sonata_config.py
```diff
@@ -308,7 +308,7 @@ def parsedProjections(self):
             for edge_pop_name, edge_pop_config in edge_config["populations"].items():
                 edge_pop = self._circuit_conf.edge_population(edge_pop_name)
                 pop_type = edge_pop_config.get("type", "chemical")
-                # skip unhandled synapse type
```
cattabiani, line 311: why did you remove the comment? outdated and missleading?

neurodamus/ngv.py
```diff
@@ -381,7 +381,8 @@ def __init__(self, circuit_conf, target_manager, cell_manager, src_cell_manager=
         kw.pop("load_offsets")
         super().__init__(circuit_conf, target_manager, cell_manager, src_cell_manager, **kw)
 
-    def _add_synapses(self, cur_conn, syns_params, syn_type_restrict=None, base_id=0):
+    @staticmethod
```
cattabiani, line 384: same as before

neurodamus/ngv.py
```diff
@@ -418,7 +419,8 @@ def finalize(self, base_Seed=0, *_):
                 "Missing cells to couple Glia to: %d", len(NeuroGlialConnection.neurons_not_found)
             )
 
-    def _create_synapse_ustate_endpoints(self, base_manager):
+    @staticmethod
```
cattabiani, line 422: the same as before. Once we agree on the point for one, you can close all of these

neurodamus/node.py
```diff
@@ -1081,8 +1078,9 @@ def _report_build_params(self, rep_name, rep_conf):
             rep_conf.get("Scaling"),
         )
 
+    @staticmethod
     @run_only_rank0
-    def _coreneuron_write_report_config(self, rep_conf, target, rep_params):
+    def _coreneuron_write_report_config(rep_conf, target, rep_params):
```
cattabiani, line 1083: same as before

neurodamus/node.py
```diff
@@ -1349,7 +1347,7 @@ def _coreneuron_configure_datadir(self, corenrn_restore, coreneuron_direct_mode)
 
         # Clean-up any previous simulations in the same output directory
         if self._cycle_i == 0 and corenrn_datadir_shm:
-            subprocess.call(["/bin/rm", "-rf", corenrn_datadir_shm])
+            shutil.rmtree(corenrn_datadir_shm, ignore_errors=True)
```
cattabiani, line 1350: the same point as before about removing large amounts of files

neurodamus/node.py
```diff
@@ -1471,8 +1469,8 @@ def _run_neuron(self):
         self.solve()
         logging.info("Simulation finished.")
 
-    # -
-    def _run_coreneuron(self):
+    @staticmethod
+    def _run_coreneuron():
```
cattabiani, line 1473: the point as before

neurodamus/node.py
```diff
@@ -1666,7 +1665,8 @@ def coreneuron_cleanup(self):
             # in restore, coreneuron data is a symbolic link
             data_folder.unlink()
         else:
-            subprocess.call(["/bin/rm", "-rf", str(data_folder)])
+            shutil.rmtree(data_folder, ignore_errors=True)
```
cattabiani, line 1668: bip bop, same point as before

neurodamus/node.py
```diff
@@ -1811,8 +1810,8 @@ def init(self):
         self.sim_init()
         assert self._sim_ready, "sim_init should have set this"
 
-    # -
-    def _merge_filesdat(self, ncycles):
+    @staticmethod
```
cattabiani, line 1813: same as before

neurodamus/node.py
```diff
@@ -2056,14 +2054,16 @@ def run(self, cleanup=True):
...
+    def _remove_file(file_name):
         import contextlib
 
         with contextlib.suppress(FileNotFoundError):
             os.remove(file_name)
 
+    @staticmethod
     @run_only_rank0
-    def _touch_file(self, file_name):
+    def _touch_file(file_name):
```
cattabiani, line 2067: same as before

neurodamus/utils/timeit.py
```diff
@@ -216,12 +216,13 @@ def timeit_show_stats(self):
 
             self._log_stats(timers_name, timers, avg_times, min_times, max_times, nof_hits)
 
+    @staticmethod
     @run_only_rank0
-    def _log_stats(self, timers_name, timers, avg_times, min_times, max_times, nof_hits):
+    def _log_stats(timers_name, timers, avg_times, min_times, max_times, nof_hits):
```
cattabiani, line 221: same as before

cattabiani (commented):
I left a lot of comments but they are actually only 3, repeated everywhere... it is very close to an approval imho

neurodamus/core/_shmutils.py
mgeplf: ok; interesting; thanks.

neurodamus/connection_manager.py
mgeplf: It's the correct way to do things; otherwise it's not clear why the self is there.
Whether or not it's a "private" method doesn't matter - they are orthogonal concepts; it's more that it clearly signals the intent.

neurodamus/io/sonata_config.py
mgeplf: IMO, it's clear from the code; the warning says the same thing.

neurodamus/connection_manager.py
cattabiani: mmm ok. All the conversations about this point can be resolved imo

neurodamus/core/_shmutils.py
mgeplf: I added a helper with a comment about `shutils.rmtree` being slow; that should make it more clear in the future.

cattabiani (approved):
LGTM

mgeplf:
this is part of https://github.com/openbraininstitute/neurodamus/issues/47

PR #248, opened 2025-05-07
Unit tests for modifications

### LINKED ISSUE #233: Unit tests for modification_manager

WeinaJi:
## Context
Fix #233. This PR adds the unit tests for cell modifications as defined in the "modifications" sections in the SONATA  config file. 

## Scope

- `modification_manager.py`: remove the legacy support function `compartment_cast`. No more needed for SONATA nodeset targets

## Testing
- `test/unit/test_modifications`: New unit tests using ringtest circuit to assess the functionalities of the `ModificationManager` class for `TTX` and `ConfigureAllSections` modification types 

## Review


tests/unit/test_modifications.py
```diff
@@ -0,0 +1,257 @@
...
+from neurodamus.modification_manager import ModificationManager
+from neurodamus.node import Neurodamus, Node
+
+SIMULATION_CONFIG_FILE = RINGTEST_DIR / "simulation_config.json"
+
+
+def test_applyTTX():
+    """
+    A test of enabling TTX with a short simulation.
+    As Ringtest cells dont't contain mechanisms that use the TTX concentration
```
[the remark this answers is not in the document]
WeinaJi: Ah, yes, corrected. Thx!

juanjosegarcan (approved):
LGTM

tests/unit/test_sonata_config.py
```diff
@@ -149,6 +149,7 @@ def test_parse_modifications(create_tmp_simulation_config_file):
     from neurodamus.core.configuration import SimConfig
 
     SimConfig.init(create_tmp_simulation_config_file, {})
+    assert list(SimConfig.modifications.keys()) == ["no_SK_E2", "applyTTX"]  # order preserved
```
cattabiani, line 152: are you sure that the order is preserved or that the order is required to be preserved?
```suggestion
    assert set(SimConfig.modifications.keys()) == {"no_SK_E2", "applyTTX"}
```

tests/unit/test_modifications.py
```diff
@@ -0,0 +1,257 @@
...
+    A test of enabling TTX with a short simulation.
+    As Ringtest cells don't contain mechanisms that use the TTX concentration
+    to enable/disable sodium channels, no spike change is expected.
+    Instead, we check that all sections contain TTXDynamicsSwitch after modification
+    """
+
+    # NeuronWrapper needs to be imported at function level
+    from neurodamus.core import NeuronWrapper as Nd
+
+    n = Node(str(SIMULATION_CONFIG_FILE))
```
cattabiani, line 24: why do you use Node instead of Neurodamus?

cattabiani (commented):
just a very minor question + a tweak

tests/unit/test_sonata_config.py
WeinaJi: Yes, the order is required to be preserved. The modifications should be done in order as in the simulation config file in case of multiple modifications on the same cells.

tests/unit/test_modifications.py
WeinaJi: Because I want to check the cell objects before and after `enable_modifications` . `Neurodamus` class will directly execute `enable_modifications`.

tests/unit/test_sonata_config.py
cattabiani: yes but in this case you cannot have a dict for SimConfig.modifications. it is not granted that it preserves order. Any cycling on that may spark errors

WeinaJi: It is preserved from python 3.7.

WeinaJi: Right. We don't have the deletion case. I will resolved this conversation and go ahead to merge this PR.

PR #252, opened 2025-05-08
astrocytes: tests + refactoring

### LINKED ISSUE #148: `ngv.Astrocyte` refactoring
There are some functions that need imrpovements:

- `_truncated_cone_surface_areas`: renaming, docstrings, generalization, move to somewhere else (simple geometry is not related to ngv cells)
- `_truncated_cone_volumes`: renaming, docstrings, generalization, move to somewhere else (simple geometry is not related to ngv cells)
- `_mcd_section_parameters`: docstrings
- `_secparams_as_hoc`: removal? Is it still necessary?
- `_init_cell` removal? Is it still necessary?

In addition: what should this class do? What do its members do? Unit tests are suggested to cover this class (if we need it)

### LINKED ISSUE #251: ngv `set_pointers` does not take into account endfeet
Consider [this function](https://github.com/openbraininstitute/neurodamus/blob/1774a831ffd351f668af288e86e10fe82a36bb23/neurodamus/ngv.py#L193)

We want to re-set the pointers. `c.all` provides all the sections WITHOUT the endfeet. This means that, if an endfoot was inserted, the index for the soma is offset by the endfeet and we are picking up a standard GlutReceive (not the soma one) when we want to re-set the pointers.

### LINKED ISSUE #257: `GlutReceiveSoma` shouldn't be pointed at by `cadifus`
- `cadifus` points to `GlutReceive` in general
- `GlutReceiveSoma` is only used for reporting

This is an example:

https://github.com/openbraininstitute/neurodamus/blob/774baa192dcfa7b29ce5e54348bbd03933797744/neurodamus/ngv.py#L204

probably there are more in the code

cattabiani:
## Context
Astrocytes may be tested better. The implementation can be largely improved. There will be additional prs. I am breaking the refactoring in manageable pieces so reviewing them is easier.

### Astrocyte
- [x] incorporate `_init_cell` in `__init__`
- [x] removed `glut_list` are a real member variable. We use now `sections_glut` and `soma_glut` variables. `glut_list` is a `@property` for compatibility with metabolism. This simplifies and makes the code more robust against off by 1 errors. 
- [x] fix #251
- [x] ~~remove hoc from `create_endfeet`~~ EDIT: moved to #258
- [x] add issue for `_nseg_warnings` #261
### MorphIOwrapper
- [x] remove `section_index2name_dict` from morphIOwrapper. replaced with `section_names`
- [x] fix #148  
- [x] add a new `@dataclass` to handle section names: `SectionName`. It has also a very handful `__str__` member function to auto-convert it in standard neuron/hoc code

### Astrocyte
- [x] `GlutReceiveSoma` should be just used for reporting. It should not interact with `cadifus`. Atm we effectively override `GLutReceive` in the soma section with this one for what concerns `cadifus`. We stop with that. fix #257

### Astrocyte
- [x] move astrocyte testing to new file. GlutList included
- [x] add testing for `GlutList`
- [x] change reference in old ngv tests since they were generated with a bug in the code
- [x] rearrange reference in tests as numpy expects the reference as last (second) element of a comparison... Otherwise it fails pointing at inverted ACTUAL and CURRENT values
- [x] change test so that there is 1 astrocyte with 2 endfeet
- [x] full comparison all Rad and R0pas in old radii test 
## MorphIOwrapper
- [x] test `get_section_names()`. Compare it with printnames? check the types

## Review


neurodamus/ngv.py
```diff
@@ -191,22 +234,27 @@ def _show_mcd(sec):
...
-        glut_list = self._glut_list
         c = self._cellref
-        index = 0
+        # the endfeet are not included in all as they are added later.
+        # I still do not know exactly when the pointers need to be
+        # reassigned and which ones are stale. the endfeet may be already
+        # up-to-date
+        all_secs = chain(c.all, self.endfeet)
+
+        # just a safety check
```
mgeplf, line 244: `assert` suggests it's a safety check, it would be more helpful to add a message to the: `assert predicate, "mismatch of of cells to glut_list"` or whatever

neurodamus/ngv.py
```diff
@@ -191,22 +234,27 @@ def _show_mcd(sec):
     #        )
 
     def set_pointers(self):
-        glut_list = self._glut_list
         c = self._cellref
```
mgeplf, line 237: since lots is being changed, can we also change this one letter variable, or drop it all together and just use `self._cellref`?

neurodamus/ngv.py
```diff
@@ -303,17 +351,20 @@ def ustate_event_handler2(syn_gid):
                 if syn_gid is None:
                     continue
 
-            glut_obj = glut_list[int(syn_params.astrocyte_section_id)]
+            glut_idx = int(syn_params.astrocyte_section_id)
+            # check that we do not connect to the GlutSynapseSoma
```
mgeplf, line 355: I'm not sure how this checks that it's not connected to `GlutSynapseSoma` (GlutSynapseSoma is not mentioned anywhere else in the file, I think) is it an invariant that it's at the end?
could also add the message to the assert

neurodamus/ngv.py
```diff
@@ -19,19 +19,136 @@
 from .utils.pyutils import append_recarray, bin_search
 
 
+class GlutList:
```
mgeplf, line 22: if it's indeed list-like, it's worth inheriting from the relevant https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes, imo.
Could also look into `UserList`.

In my experience, creating data types that are builtin-like usually biting people, b/c of their subtle differences from the real thing, which catch people out later on.

For instance:
```
> a=[1,2,3,4,5]
> a[1:3]
[2, 3]

b=GlutList([1,2,3,4], 5)
b[1:3]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[15], line 1
----> 1 b[1:3]

File ~/src/neurodamus/neurodamus/ngv.py:53, in GlutList.__getitem__(self, index)
     52 def __getitem__(self, index):
---> 53     index = self._sanitize_index(index)
     54     return self._body[index] if index != len(self) - 1 else self._tail

File ~/src/neurodamus/neurodamus/ngv.py:46, in GlutList._sanitize_index(self, index)
     44 total_len = len(self)
     45 og_index = index
---> 46 if index < 0:
     47     index += total_len
     48 if index < 0 or index >= total_len:

TypeError: '<' not supported between instances of 'slice' and 'int'
```

Would it be possible to have a class with a two members body, tail (but w/ better names, describing why the tail is special: maybe `sections` and `soma` or something similar?), and then a method like `iterall`?

neurodamus/ngv.py
```diff
@@ -19,19 +19,136 @@
...
+        return [*self._body, self._tail]
+
+    def __eq__(self, other):
+        if isinstance(other, GlutList):
+            return self._body == other._body and self._tail == other._tail
+        if isinstance(other, list):
+            return self.to_list() == other
+        return NotImplemented
+
+    @property
```
mgeplf, line 86: why not just make `tail` "public" instead of a property?

neurodamus/ngv.py
```diff
@@ -19,19 +19,136 @@
...
+
 class Astrocyte(BaseCell):
-    __slots__ = ("_glut_list", "_nseg_warning", "_secidx2names")
+    __slots__ = ("_glut_list", "_nseg_warning", "section_names")
 
     def __init__(self, gid, meinfos, circuit_conf):
-        """Instantiate a new Cell from node info."""
+        """Initialize an Astrocyte cell.
+
+        - Load and attach morphology to the cell.
```
mgeplf, line 104: not sure this is needed here, since the comments say the same thing

neurodamus/ngv.py
```diff
@@ -19,19 +19,136 @@
...
+        # Recalculate number of segments and sections
+        self._cellref.geom_nseg_fixed()
+        self._cellref.geom_nsec()
+
+        logging.debug("Instantiating NGV cell gid=%d", gid)
+
+        # Insert mechanisms and glutamate receptors in each section
+        for sec in self._cellref.all:
+            if sec.nseg > 1:
+                self._nseg_warning = 1
```
mgeplf, line 135: what sort of warning is this? it also seems to be used by AstrocyteManager, so should it be public?

neurodamus/ngv.py
```diff
@@ -191,22 +234,27 @@ def _show_mcd(sec):
     #        )
 
     def set_pointers(self):
-        glut_list = self._glut_list
         c = self._cellref
-        index = 0
+        # the endfeet are not included in all as they are added later.
+        # I still do not know exactly when the pointers need to be
```
mgeplf, line 239: is this a note that should be acted on before this is merged?

neurodamus/ngv.py
```diff
@@ -191,22 +234,27 @@ def _show_mcd(sec):
...
-        for sec in c.all:
-            glut = glut_list[index]
-            index += 1
+        for glut, sec in zip(self.glut_list, all_secs):
             Nd.setpointer(glut._ref_glut, "glu2", sec(0.5).cadifus)
-        soma = c.soma[0]
-        glut = glut_list[index]
-        Nd.setpointer(glut._ref_glut, "glu2", soma(0.5).cadifus)
 
     @property
```
mgeplf, line 250: why not just make `_glutlist` public?

neurodamus/morphio_wrapper.py
```diff
@@ -222,21 +221,22 @@ def _build_morph(self):
 
         self._morph = Morphology(self._morph)
 
-    def _build_sec_idx2names(self):
-        """Build section index to nrn section names mapping on top of MorphIO section.id"""
-        # soma is not accounted for in sections
-        self._sec_idx2names[0] = "soma"
+    def section_names_generator(self):
+        """Generator yielding (section_name, relative_index), starting with ('soma', 0)."""
```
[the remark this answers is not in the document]
cattabiani: dataclasses? Ahahah maybe it is not clear to me too. Do you have an example in mind?

- it is a generator because in this way it updates if the content changes. If I create a new list I need to keep it in sync. However, if you are sure that I can calculate it only once and then I can cache it, I can produce a list insted

neurodamus/morphio_wrapper.py
```diff
@@ -303,12 +303,12 @@ def morph_as_hoc(self):
         # generate sections connect + their respective 3D points commands
         for i, sec in enumerate(self._morph.sections):
             index = i + 1
-            tstr = self._sec_idx2names[index]
+            tstr = MorphIOWrapper.combined_name(*self.section_names[index])
```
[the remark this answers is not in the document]
cattabiani: why? I will change, np on that. Why is it harder to refactor?

neurodamus/ngv.py
cattabiani: it is a safety check (as I wrote just before). Anyway, done

neurodamus/ngv.py
cattabiani: yes, np with that. Done

neurodamus/ngv.py
cattabiani: in case we want to have a translation layer later. It is probably premature optimization. I will change that

cattabiani: made when I was thinking that I will need that translation layer btw. Now it turned to premature opt

neurodamus/ngv.py
cattabiani: yes, redundant. Removed

neurodamus/morphio_wrapper.py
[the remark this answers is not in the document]
mgeplf: > Why is it harder to refactor?

It makes it harder to change `combined_name` signature, or the contents of `section_names`; suppose you wanted to add another element to the return value of `section_names_generator`, for example.

neurodamus/morphio_wrapper.py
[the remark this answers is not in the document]
mgeplf: WRT dataclass:
```
@dataclass
class SectionLocation:
    section_type_name: str
    segment: int
```

> it is a generator because in this way it updates if the content changes.

I'm not sure a follow, I'm saying continue having a function `get_section_names` that returns a list, that is created on the fly each time it's called, and thus has the changed contents.

neurodamus/ngv.py
mgeplf: I'm saying that using `assert` already tells signals it's a safety check; so a more helpful comment (or, message) explains the goal of the safety check.

neurodamus/morphio_wrapper.py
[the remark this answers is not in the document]
cattabiani: ok, in this case I revert and explicitly pick the first 2 elements

neurodamus/morphio_wrapper.py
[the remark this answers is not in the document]
cattabiani: yes, that is a solution equivalent to a generator. However, if you just need to iterate over it a generator is more efficient. As you prefer, the change in efficiency is undetectable probably

neurodamus/ngv.py
cattabiani: got it. ok

neurodamus/morphio_wrapper.py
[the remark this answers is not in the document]
mgeplf: it looks like it's being used in only one place, where it's converted to a list right away, which is why I figured it's used as a list always; if you think it's going to be iterated elsewhere, no need to change it - however, that opens the question as to why `self.section_names` is stored - should the generator always be used?

neurodamus/ngv.py
cattabiani: that was my initial iteration but I was getting tangled in overloads as the class behaves differently since it is never empty. In addition this is a classic case of "composition over inheritance". If you need slices I can added them easily. I am also not fond of this class. It is here only because if I change it metabolism will probably complain... Now that I think about it I could just split this in 2 new variables and have just a small property function that returns the glut_list if needed by metabolism but is not used in the code... probably a better solution

neurodamus/morphio_wrapper.py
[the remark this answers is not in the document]
cattabiani: everywhere I use it I actually need to retrieve the value by index (so I need to convert). Yes, let me turn it to a list created on the fly

neurodamus/ngv.py
cattabiani: eheheh, this was a todo/check for me. It is superfluous now

neurodamus/ngv.py
cattabiani: there is an issue on this #261. I will takle this once I proceed on AstrocyteManager

neurodamus/ngv.py
cattabiani: no. I left if for posterity in case I figure out why these pointers need to be reassigned. Maybe creating an issue is better

cattabiani: #263

neurodamus/ngv.py
cattabiani: another case of premature opt. I was considering options added this and forgot to remove it

neurodamus/morphio_wrapper.py
```diff
@@ -222,21 +221,22 @@ def _build_morph(self):
 
         self._morph = Morphology(self._morph)
 
-    def _build_sec_idx2names(self):
-        """Build section index to nrn section names mapping on top of MorphIO section.id"""
-        # soma is not accounted for in sections
-        self._sec_idx2names[0] = "soma"
+    def section_names_generator(self):
```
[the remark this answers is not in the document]
cattabiani: probably we could use that however, it is simpler in this way imho. Also, this does not rely on cached values or maps so it is less likely to give desync problems

neurodamus/morphio_wrapper.py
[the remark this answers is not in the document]
cattabiani: About `relative_index` I would keep it as a pair `string, int` as a new class to me is premature opt. I can improve the comment to explain relative_index ... relative to what

neurodamus/morphio_wrapper.py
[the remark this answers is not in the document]
WeinaJi: But then where is the output of `_build_sec_typeid_distrib()` used ?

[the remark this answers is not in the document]
cattabiani: `morph_as_hoc` in the same class

neurodamus/ngv.py
cattabiani: done. We use `sections_glut` and `soma_glut` to record this stuff. `glut_list` is a property for compatibility reasons with metabolism

neurodamus/ngv.py
mgeplf: ok

neurodamus/ngv.py
mgeplf: I think it needs to be figured out before this is merged because I'm not sure how this is comment actionable

neurodamus/morphio_wrapper.py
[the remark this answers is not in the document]
mgeplf: the nice thing about a dataclass, which is basically just meant to hold data and have no methods, is that it helps a lot with readability and understanding what is contained, what the type is, and gives a place for a documentation.  It's not an optimization, it helps with readabilty and understanding of the code.

neurodamus/ngv.py
cattabiani: I do not know how to do something about it at the moment. The reason is probably buried in stdinit. I opened an issue to ask

neurodamus/morphio_wrapper.py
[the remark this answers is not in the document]
cattabiani: to me it is clearer to have a pair of string and int but I understand that it is my personal opinion. If you insist, I will proceed with the change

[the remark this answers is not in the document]
mgeplf: I think I'm misunderstanding what you mean by clearer?
The dataclass makes the types explicit, and makes it easy for a human to look at what the expected types are, while a tuple does neither of those things.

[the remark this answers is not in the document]
cattabiani: to me a tuple(str, int) is very clear. The dataclass adds a level of indirection. I need to check what is this class to find out that it is a tuple of str and int

neurodamus/ngv.py
mgeplf: Tough one, I guess we go w/ this for now.

mgeplf (approved):

juanjosegarcan (approved):
LGTM just two comments on unneeded files update

tests/simulations/ngv/edges.h5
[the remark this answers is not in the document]
cattabiani: I did not add it. It was create_data.py. what do you mean it is not needed?

PR #254, opened 2025-05-09
remove hocify

mgeplf:
* to speed up simulation loading, there was an attempt to load all the morphologies, and dump them as hoc.  This process was called `hocify`.
* this, however, uses many files, and we had an inode exhaustion problem, which limits the usability


mgeplf:
Jim: "this was a good effort, and people learned a lot, but now we can archive it"

Issue #225, opened 2025-04-28
It appears `ConnectionManagerBase::reenable*` are unused


mgeplf:
It looks safe to remove, it's untested and not referred to by any documentation outside of autocreated documentation.

Can they be deleted?

@WeinaJi / @cattabiani / @juanjosegarcan / @jplanasc / @jamesgking ?

mgeplf:
There are a couple tests for `delete/delete_group` in `tests/unit/test_conn_manager.py` which is why I didn't remove them.

From what I can tell, this all looks like code from 2018/2019, and hasn't really been touched since then, which is another indication that it is unused code.

mgeplf:
any input @jamesgking ?

jamesgking:
Looking at my email history, these kinds of functions were added as convenience functions at the request of Oren for his phd research. I can write him to give him a heads up that we think to remove them and get his thoughts.

@WeinaJi you were helping him with his gap junction code, and these functions weren't being used there, right?

WeinaJi:
> Looking at my email history, these kinds of functions were added as convenience functions at the request of Oren for his phd research. I can write him to give him a heads up that we think to remove them and get his thoughts.
> 
> [@WeinaJi](https://github.com/WeinaJi) you were helping him with his gap junction code, and these functions weren't being used there, right?

No, these functions are not needed for applying the gj junction corrections. Maybe they were used for computing the gj corrections?

jdcourcol:
Don't waste time with that. It is anyway in the git history in the unlikely case we have to resurrect this. just remove it.

mgeplf:
I updated the commit; I kept `ConnectionManagerBase::_unlock_all_connections` because it looks like it's related to something else, but I'm not entirely sure.

mgeplf:
PR merged.

PR #265, opened 2025-05-13
Remove unused methods in `EngineBase` and `ConnectionBase::{enable,disable}`

mgeplf:
Remove `ConnectionBase::{enable,disable}`, seemingly unused
    * remove the plumbing needed for this, as well (ie: `skip_disabled` and
      thus the dependency in the `finalize` methods on `sim_corenrn`)

Remove unused methods of `EngineBase`

Remove unused methods in `ConnectionSet`


cattabiani:
the removal would clearly simlify the code however... this is a feature that is working now. How can we be sure that this is really unused?

mgeplf:
Which feature?

mgeplf:
Also, I would add, that we don't actually know if the feature is working - it's untested, and not used anywhere in the code.

mgeplf:
I think this goes back to the discussion in https://github.com/openbraininstitute/neurodamus/issues/225 and it's basically the same thing - code that isn't used internally in neurodamus.

neurodamus/core/_engine.py
```diff
@@ -77,23 +77,3 @@ class EngineBase(metaclass=_EngineMeta):
...
-
-    @classmethod
-    def new_cell_manager(cls, circuit_conf, target_parser, run_conf):
-        if not cls.CellManagerCls:
-            raise RuntimeError("missing CellManagerCls")
-        return cls.CellManagerCls(circuit_conf, target_parser, run_conf)
-
-    def new_synapse_manager(cls, circuit_conf, target_manager, cell_manager):
-        """Routine responsible for creating connection managers."""
-        if not cls.SynapseManagerCls:
```
WeinaJi, line 89: Can you also update the docstring of class `EngineBase` ?  `SynapseManagerCls` -> `InnerConnectivityCls`

mgeplf: sure, will do

mgeplf:
> we don't actually know if the feature is working

I just noticed that the `NeuroGlialConnection` isn't respecting the `_disabled` property, so that's more evidence that the implementation is uneven.

WeinaJi (approved):

neurodamus/core/_engine.py
```diff
@@ -77,23 +77,3 @@ class EngineBase(metaclass=_EngineMeta):
     """A dict of the new connection types and associated Manager class"""
     CircuitPrecedence = 1
```
mgeplf, line 78: @WeinaJi: I wonder if this precendence should be removed?
I noticed that the `METypeEngine` sets it to 0, but NGV doesn't use it.
It appears to be used to make iteration deterministic:
https://github.com/openbraininstitute/neurodamus/blob/ad08b95c7d69a2d8ed7184dad42b422636464a23/neurodamus/core/configuration.py#L693

However, I think is already done since `__engines` is a dictionary, so it's based on insertion time.
Thoughts?

WeinaJi: It ensures that the biophysical population get instantiated first in case that they are in a random order in `circuit_config.json`. I think it matters for the load balance (as we are loading only 1 non-virtual population for computing the load balance).

mgeplf: does it matter that there may be more than one biophysical population?

PR #269, opened 2025-05-14
ngv refactoring + tests

### LINKED ISSUE #149: ngv: `USE_COMPAT_SYNAPSE_ID = True` find out if it is still required
Having `USE_COMPAT_SYNAPSE_ID = True` in the middle of the classes (not in a class) is very un-pythonic. Do we need this? What is it for? It is [here](https://github.com/openbraininstitute/neurodamus/blob/6cca80a5cdd2a09ea9493ab69f4f8a7624344d33/neurodamus/ngv.py#L226) and it looks like an optimization not yet fully implemented in ngv. What were the plans here? What do we want to do? @jamesgking

### LINKED ISSUE #259: check `_connect_endfeet`
`GlioVascularManager._connect_endfeet` should be tested

### LINKED ISSUE #261: warnings in `_nseg_warning` looks very strange
As @mgeplf pointed out, this way of collecting warnings is very strange and probably can be refactored. To be done while refactoring `AstrocyteManager` and after refactoring `Astrocyte` to have the full picture.

Example of use of this variable:

https://github.com/openbraininstitute/neurodamus/blob/774baa192dcfa7b29ce5e54348bbd03933797744/neurodamus/ngv.py#L233

### LINKED ISSUE #262: `AstrocyteManager` refactoring

### LINKED ISSUE #263: why `set_pointers` is needed?
[here](https://github.com/openbraininstitute/neurodamus/blob/774baa192dcfa7b29ce5e54348bbd03933797744/neurodamus/ngv.py#L193)

there is a line about caching and the pointers that need to be reassigned. I still do not geet exactly which pointers need to be reassigned but I am sure that is needed for the `c.all` sections. The endfeet? I dunno

### LINKED ISSUE #267: is endoplasmic reticulum a thing in ngv?
I see this in the ngv code:

```
        # Configure endoplasmic reticulum and section parameters
        self._cellref.execute_commands(self._er_as_hoc(morph))
        self._cellref.execute_commands(self._secparams_as_hoc(morph))
```

do we have this endoplasmic reticulum? These commands are empty. I wonder if I can remove this stuff

### LINKED ISSUE #272: `vasculature` shouldn't be optional
The ngv code treats the `vasculature` for the astrocytes as optional. It should not be the case. This can simplify the code. @dkeller9 confirmed this (thx Dan!).

### LINKED ISSUE #276: obsolete `post_stdinit`
after some investigation it turns out that the only cell distributor that was really using [here](https://github.com/openbraininstitute/neurodamus/blob/1583edb87dab809b6a2fc29ad5f625486afedb5b/neurodamus/node.py#L1210) was ngv. We found out that the main part (reassign pointers) is not needed anymore so it just collects a warning about resizing sections to 1 compartment (that to me is not that important). Do we still want this legacy code or can I remove everything?

### LINKED ISSUE #285: `GlutReceive` should be placed only where needed
At the moment we place `GlutReceive` everywhere because `cadifus` needed a valid pointer `glu2`. With the addition of https://github.com/openbraininstitute/neurodamus-models/pull/8 this is no longer needed. We should place `GlutReceive` mechanisms
only where needed. Notice that there may be multiple NetCons connected to the same GlutReceive. In that case we need to connect to the same `GlutReceive` point process.

cattabiani:
## Context
ngv can be refactored and tested more

## Scope
- [x] rework `_connect_endfeet` and dump the construction and endfeet connection to the astrocyte in the astrocyte class. It belongs there. Fix #259
- [x] remove endoplasmic reticulum functions. The functions were unused
- [x] add in the docs what we were doing with the endoplasmic reticulum. Partial fix #267
- [x] add `_init_section` and `_init_endfoot_section`: it is more SMART
- [x] fix #261: better warning for section resizing
- [x] renamed `*glut` lists in `glut_all`, `glut_endfeet`, and `glut_soma` to underline where they are from. `glut_list` remains for compatibility with metabolism
- [x] simplified `R0pas` extraction. I do not think an endfoot without a vasculature has much sense
- [x] fix #262
- [x] fix #263 not necessary anymore. From Michael Hines 
>I do believe all ref_vars are automatically updated. CVode.cache_efficient now does nothing. In version 9 all range variable memory layout is Structure of Array. The substance of cache efficiency is now merely a  permutation.
Replaced with the more pythonic: `sec(0.5).cadifus._ref_glu2 = glut._ref_glut`. 
- [x] fix #276. No answer in the ticket. I decide: we can emit a normal warning when we encounter the problem and avoid the whole `post_stdinit` function which, in turn, can be safely removed now.

There is a test to check that it still works as a pointer
- [x] fix #272 make vasculature mandatory
- [x] `GlioVascularManager` should use proper `ConfigurationError`, not `Exception` 
- [x] skip refactoring of `NeuroGlialConnection` as it requires unblocking of #149 and this fix is likely to be a PR by itself based on its complexity 
- [x] fix #269
- [x] fix #149: we remove the semi-implemented the feature
- [x] fix #285: `GlutReceive` is placed only when needed. `get_glut` ensures that we place, if needed, at most 1 `GlutReceive` per section
- [x] added a new `_emit_init_cell_warnings` that collects and emits warnings from cell inits in rank 0. Testing this is going to be complicated because I need to create a new cell type, a manager that uses that and then I need to somehow init it. I think to init that I need a config object and a population which requires various files and configuration data. Maybe it is just an overkill just to see if this function correctly reports the warnings. @WeinaJi what is your opinion?
- [x] remove `netcon.record` debug messages

## Testing
- [x] add real unit tests for the `Astrocyte`

## Review


docs/source/ngv.rst
```diff
@@ -106,6 +106,19 @@ There was an initial attempt to incorporate the ER into NGV. This effort was tem
 
 Ultimately, the feature was removed at the time, as partially implemented functionality added complexity without delivering tangible benefits.
 
+Optimized Synapse GID Assignment
```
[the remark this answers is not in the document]
cattabiani: ah no! thx for catching this

neurodamus/ngv.py
```diff
@@ -10,65 +9,98 @@
...
 
+    def _init_basic_section(self, sec):
+        """Init a normal section
+
+        Initialize sec with mechanisms and point processes
+        (store PP to avoid GC).
+        """
+        # resize if necessary
+        if sec.nseg > 1:
+            logging.warning(
```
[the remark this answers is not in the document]
cattabiani: that requires a `post_stdinit` which is a function that became obsolete apart from this warning. Maybe I can try to warn once but not in the post_stdinit?

neurodamus/ngv.py
```diff
@@ -187,24 +157,15 @@ def getVersion():
 
 
 class AstrocyteManager(CellDistributor):
-    # Cell Manager is the same as CellDistributor, so it's able to handle
-    # the same Node formats and Cell morphologies.
-    # The difference lies only in the Cell Type
-    CellType = Astrocyte
-    _sonata_with_extra_attrs = False
+    """Manages Astrocyte cells, extending CellDistributor with post-stdinit handling.
```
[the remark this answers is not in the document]
cattabiani: yes, that is old

neurodamus/ngv.py
```diff
@@ -391,89 +291,61 @@ def finalize(self, base_Seed=0, *_):
...
+                )
+                # set the v-gid (that is actually an synapse id) to the netcon. Useful for
+                # reporting and debugging
+                pc.cell(synapse_gid, netcon)
+                if GlobalConfig.verbosity >= LogLevel.DEBUG:
+                    netcon.record(
+                        lambda tgid=conn.tgid, synapse_gid=synapse_gid: print(
+                            f"[gid={tgid}] Ustate netcon event. Spiking via v-gid:",
+                            synapse_gid,
+                        )
```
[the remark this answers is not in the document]
cattabiani: true

tox.ini
```diff
@@ -35,6 +35,24 @@ commands =
     {toxinidir}/ci/build_ndcore.sh {toxinidir}/neurodamus/data {toxworkdir}
     sh -c ". {toxworkdir}/.envfile && pytest -x --cov=neurodamus --cov-report=term-missing --forked {posargs:tests/unit}"
 
+[testenv:removeme]
```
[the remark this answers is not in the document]
cattabiani: [in a soothing voice] "this is not the piece of code that you are looking for..." [cit.]

apart the joke, I just forgot to remove this debugging code. thx!

tests/unit-ngv/test_astrocyte.py
```diff
@@ -1,55 +1,61 @@
...
+
+    # glut_list tests
+    glut_list = astro.glut_list
+    len(glut_list) == len(astro.glut_all) + len(astro.glut_endfeet) + 1
+    glut_receive_type = type(glut_list[0])  # Assuming the first element is GlutReceive
+    glut_receive_soma_type = type(glut_list[-1])  # Assuming the last element is GlutReceiveSoma
+
+    # Check that all elements in GlutList are GlutReceive except the last one
+    assert all(isinstance(glut, glut_receive_type) for glut in glut_list[:-1])
+    assert isinstance(glut_list[-1], glut_receive_soma_type)
```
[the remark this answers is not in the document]
cattabiani: mmm yes, that is possible

tests/unit-ngv/test_astrocyte.py
```diff
@@ -1,55 +1,61 @@
...
+
+    for glut, sec in zip(astro.glut_list, chain(astro.all, astro.endfeet)):
+        sec_glut = sec(0.5).point_processes()[-1]
+        assert glut.same(sec_glut)
```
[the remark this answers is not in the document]
cattabiani: 3 point processes? GlutReceive, GlutReceiveSoma and? cadifus is a mechanism. Anyway, for the first 2 it exists and I verified that it works as expected

tests/unit-ngv/test_astrocyte.py
```diff
@@ -1,55 +1,61 @@
...
+    npt.assert_allclose(R0passes, [sec(0.5).vascouplingB.R0pas for sec in astro.endfeet])
+
+    # glut_list tests
+    glut_list = astro.glut_list
+    len(glut_list) == len(astro.glut_all) + len(astro.glut_endfeet) + 1
+    glut_receive_type = type(glut_list[0])  # Assuming the first element is GlutReceive
+    glut_receive_soma_type = type(glut_list[-1])  # Assuming the last element is GlutReceiveSoma
+
+    # Check that all elements in GlutList are GlutReceive except the last one
+    assert all(isinstance(glut, glut_receive_type) for glut in glut_list[:-1])
```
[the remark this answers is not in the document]
cattabiani: yes, I am going to explore the hname route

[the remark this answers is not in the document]
cattabiani: done

tests/unit-ngv/test_astrocyte.py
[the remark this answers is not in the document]
WeinaJi: ok, no problem now.

neurodamus/ngv.py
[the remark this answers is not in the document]
WeinaJi: I see. Then we just warn once per cell instead of per section. How about that ?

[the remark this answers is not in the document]
cattabiani: good! I will do it tomorrow

docs/source/ngv.rst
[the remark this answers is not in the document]
WeinaJi: OK, then we can keep it here

neurodamus/ngv.py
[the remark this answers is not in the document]
cattabiani: done

tests/unit-ngv/test_astrocyte.py
```diff
@@ -1,55 +1,67 @@
-import pytest
-
-from neurodamus import Neurodamus
+import numpy.testing as npt
+from neurodamus.ngv import Astrocyte
 from tests.conftest import NGV_DIR
+from unittest.mock import MagicMock
+
+meinfos = MagicMock()
```
mgeplf, line 6: how hard is it to create real MEInfos/circuit_conf as opposed to using MagicMock?
In my experience, using mocks like this is fragile, as whatever is being mocked, which isn't indicated here, may change during a refactor, but than this code will still pass the unit tests.

neurodamus/ngv.py
```diff
@@ -81,99 +129,41 @@ def gid(self, val: int):
...
-            "probably some sections are unaccounted for"
-        )
-
-        for glut, sec in zip(self.sections_glut, all_secs):
-            Nd.setpointer(glut._ref_glut, "glu2", sec(0.5).cadifus)
+        for sec, parent_id, length, diameter, R0pas in zip(
+            self.endfeet, parent_ids, lengths, diameters, R0passes
+        ):
+            is_resized = self._init_endfoot_section(sec, parent_id, length, diameter, R0pas)
+            assert not is_resized, "Endfeet sections should be created with only 1 compartment"
```
WeinaJi, line 161: Do we assert this currently in the main branch? 
If it needs to raise errors, it seems efficient to assert earlier before the execution in `_init_endfoot_section` ?

cattabiani: this should never happen. If you want I can remove the check. The thing is that for `all` the function returns if there were resizes and that is relevant. Here I just have the info: `is_resized` and should always be false.

neurodamus/ngv.py
```diff
@@ -11,65 +10,114 @@
...
+        - resize sections if necessary
+        - add cadifus, GlutReceive (they will be connected later)
+        - record section_names for creating connections later
+        """
         super().__init__(gid, meinfos, None)
 
-        # Compose the path to the morphology file
-        morph_file = (
+        # Create the cell
+        self._cellref = Nd.Cell(gid)  # cell instantiated with Cell.hoc
```
jamesgking, line 36: The morphology loading in Cell.hoc has been made more robust compared to when first working with earlier ngv development. Maybe a future PR can see if there are still changes necessary in order to simply call Cell( gid, morphfile)?

tests/unit-ngv/test_astrocyte.py
cattabiani: ok, I can try. I hope it is not entangled too much in the rest of the code. If I need to use the whole neurodamus then it becomes an integration test

neurodamus/ngv.py
```diff
@@ -11,65 +10,114 @@
...
+        sec_list = getattr(self._cellref, section_name.name)
+        return sec_list[section_name.id]
+
+    @staticmethod
+    def _init_basic_section(sec):
+        """Init a normal section
+
+        Initialize sec with mechanisms and point processes
+        (store PP to avoid GC).
+        """
```
WeinaJi, line 83: Can you add what is returned in the docstring? Becuase it is not clear to me from the function name.

neurodamus/ngv.py
```diff
@@ -11,65 +10,114 @@
...
-        # connected to other point processes or mechanisms
+        # assigned later
+        self._gluts = {}
+        is_resized_secs = False
+        for sec in self.all:
+            is_resized = self._init_basic_section(sec)
+            is_resized_secs = is_resized_secs or is_resized
+
+        if is_resized_secs:
+            logging.warning(
```
jamesgking, line 57: As discussed, this *might* show many warnings. We could reduce it more if it is worth the effort? I won't press too much at this time, though.

neurodamus/ngv.py
WeinaJi: Then we can remove this check, and `_init_endfoot_section` takes `is_resized` from `_init_basic_section` but doesn't return it. What do you think?

neurodamus/ngv.py
```diff
@@ -11,65 +10,114 @@
...
+    @staticmethod
+    def _init_basic_section(sec):
+        """Init a normal section
+
+        Initialize sec with mechanisms and point processes
+        (store PP to avoid GC).
+        """
+        # resize if necessary
+        is_resized_sec = sec.nseg > 1
+        if is_resized_sec:
```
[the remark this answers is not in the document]
cattabiani: if nseg == 1 we would asign is_resized_sec to 1. I think you meant:

if is_resized_sec := (sec.nseg > 1):

neurodamus/ngv.py
cattabiani: good point. Done

neurodamus/ngv.py
```diff
@@ -255,95 +230,40 @@ def finalize(self, astrocyte, base_Seed, *, base_connections=None, **kw):
...
-            glut_obj = sections_glut[glut_idx]
+            glut_obj = astrocyte.get_glut(glut_idx)
             netcon = pc.gid_connect(syn_gid, glut_obj)
-            netcon.delay = 0.05
-
-            netcon.record(ustate_event_handler2(syn_gid))
-
+            netcon.delay = self.netcon_delay
+            netcon.record(
+                lambda syn_gid=syn_gid: print(f"GOOD netcon event 2. Spiking via v-gid: {syn_gid}")
```
jamesgking, line 267: These printed messages seem like they were for debugging. Can @dkeller9 confirm if they should be removed in production?

jamesgking: I see below a similar construct is used, but checks for
`if GlobalConfig.verbosity >= LogLevel.DEBUG:`
Maybe that should also be employed here?

tests/unit-ngv/test_astrocyte.py
cattabiani: magicMock removed

neurodamus/ngv.py
cattabiani: ok, np. Done

neurodamus/ngv.py
cattabiani: I did not know that was a possibility. I can open an issue

neurodamus/ngv.py
cattabiani: I think a natural development is that we accumulate the warnings and emit at the end of init instead of an `ad hoc` function `post_stdinit`. I can investigate that

neurodamus/ngv.py
cattabiani: #286

neurodamus/ngv.py
```diff
@@ -190,25 +178,42 @@ def getVersion():
...
     CellType = Astrocyte
     _sonata_with_extra_attrs = False
 
-    def post_stdinit(self):
-        nseg_warning = 0
-        for cell in self.cells:
-            cell.set_pointers()
-            nseg_warning += cell._nseg_warning
+    def _emit_resized_section_warnings(self):
+        """TODO"""
```
WeinaJi, line 191: What is still on the "TODO" list?

cattabiani: done

PR #296, opened 2025-05-30
Remove more unused code

mgeplf:


mgeplf:
Did a pass with vulture (https://github.com/jendrikseipp/vulture - hat-tip @jdcourcol ) and got rid of some more unnecessary code.

neurodamus/core/_neuron.py
```diff
@@ -86,24 +86,6 @@ def section_in_stack(self, sec):
         yield
         self.h.pop_section()
 
-    @classmethod
-    def run_sim(cls, t_stop, *monitored_sections, **params):
```
cattabiani, line 90: was this used in bbp_workflow? Is this a part that JD suggested to remove?

neurodamus/core/configuration.py
```diff
@@ -216,10 +213,8 @@ class _SimConfig:
     buffer_time = 25
     save = None
     restore = None
-    coreneuron_outputdir = None
```
cattabiani, line 219: good catch! I checked and apparently I forgot to remove this line when I reworked the configuration.py. Probably due to the high-intesity back-and-forth review with Weina

neurodamus/metype.py
```diff
@@ -315,24 +298,6 @@ def _make_coord_map_matrix(position, rotation, scale):
         m[:, 3] *= scale
         return m
 
-    def local_to_global_coord_mapping(self, points):
```
cattabiani, line 318: this is used in multiscale run (if only ngv was working):

https://github.com/openbraininstitute/MultiscaleRun/blob/45c493da4e6c9ec604c672864b6927a3b5e34977/multiscale_run/neurodamus_manager.py#L191

neurodamus/utils/progressbar.py
```diff
@@ -122,14 +122,6 @@ def iter(cls, iterable, end=None, start=0, **kw):
...
         return cls(end, start, **kw)(iterable, end or None, start)
 
-    @classmethod
-    def itervalues(cls, iterable, **kw):
-        return cls.iter(iterable.values(), len(iterable), **kw)
-
-    @classmethod
-    def iteritems(cls, iterable, **kw):
-        return cls.iter(iterable.items(), len(iterable), **kw)
-
```
cattabiani, line 132: I think at least one of this is used in multiscale run:
https://github.com/openbraininstitute/MultiscaleRun/blob/45c493da4e6c9ec604c672864b6927a3b5e34977/multiscale_run/simulation.py#L211

tests/unit/test_coords.py
```diff
@@ -1,29 +0,0 @@
-import neurodamus.metype as metype
```
cattabiani, line 1: commected to the comment before: I think the function is used

tests/unit/test_multimap.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_merge_grouped():
+    d = GroupedMultiMap(np.array([3, 4, 3], "i"), [1, 2, 3])
+    d += GroupedMultiMap(np.array([2, 4, 3], "i"), ["x", "y", "z"])
+    assert d[3] == [1, 3, "z"]
+
+    d = GroupedMultiMap(np.array([3, 4, 3], "i"), [1, 2, 3])
+    d += GroupedMultiMap(np.array([2, 4, 3], "i"), ["x", "y", "z"])
+    assert d[3] == [1, 3, "z"]
```
cattabiani, line 53: I would add a small test to check that the vals are sorted (if we want that property) or remain unsorted (if we want that). Atm you are adding the values already sorted so these 2 are both valid at the same time

cattabiani: a test for an empty GroupedMultiMap?

cattabiani: is it required to have an np.array as the keys as input? Can I use a simple list of ints instead? If not, why?

cattabiani (commented):
I think some (2) of these functions may be used in MultiscaleRun. I would not remove them before restoring multiscaleRun to a usable state

neurodamus/core/configuration.py
mgeplf: no problem

neurodamus/utils/multimap.py
```diff
@@ -127,26 +93,19 @@ def _duplicates_to_list(np_keys, values):
...
     def __iadd__(self, other):
-        MultiMap.__iadd__(self, other)
+        """Inplace add (incorporate other)"""
+        self._keys, self._values = self.sort_together(
+            np.concatenate((self._keys, other._keys)), self._concat(self._values, other._values)
+        )
         self._keys, v_list = self._duplicates_to_list(self._keys, self._values)
         self._values = [reduce(add, subl) for subl in v_list]
         return self
-
```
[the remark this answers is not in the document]
mgeplf: Which, `__iadd__` or `flat_values`?  I'm removing an unused baseclass here, so the diff looks messy.

neurodamus/utils/progressbar.py
mgeplf: I don't see any usage of `itervalues` or `iteritems`; the `__iter__` is still there, though.

tests/unit/test_multimap.py
mgeplf: > a test for an empty GroupedMultiMap?

You mean like on line 42? I can add a name to make it more explicit.

> Atm you are adding the values already sorted so these 2 are both valid at the same time

Frankly, I wish this had better documentation/tests for what the expected behavior is.
I didn't change anything, afaik, so it's whatever existed before.
I'm not sure which is "right", or if it matters.

> is it required to have an np.array as the keys as input? Can I use a simple list of ints instead? If not, why?

I'm keeping the same interface that existed before, so I can't really say.

neurodamus/utils/progressbar.py
cattabiani: ah in this case, forget this comment

neurodamus/metype.py
mgeplf: Ok, I've put them back, and created a ticket: https://github.com/openbraininstitute/neurodamus/issues/320

tests/unit/test_coords.py
mgeplf: re-added.

neurodamus/utils/compat.py
```diff
@@ -24,10 +24,6 @@ def __add__(self, other):
         array.extend(self, other)
         return self
 
-    def as_hoc(self):
```
cattabiani, line 27: I have seen this used somewhere... anyway, if the CI passes I am fine

neurodamus/utils/multimap.py
```diff
@@ -7,24 +7,24 @@
...
             values: The array of the values, can be any indexable, but better if numpy
         """
         assert len(np_keys) == len(values), "Keys and values must have the same length"
-        if presorted:
-            self._keys = np_keys
-            self._values = values
-        else:
-            self._keys, self._values = self.sort_together(np_keys, values)
+        self._keys, self._values = self.sort_together(np_keys, values)
+        self._keys, self._values = self._duplicates_to_list(self._keys, self._values)
```
cattabiani, line 27: while you are at that I think duplicates_to_list may be made more pythonic:

```
from itertools import groupby

@staticmethod
def _duplicates_to_list(np_keys, values):
    grouped_keys = []
    grouped_values = []

    for key, group in groupby(zip(np_keys, values), key=lambda x: x[0]):
        grouped_keys.append(key)
        grouped_values.append([val for _, val in group])

    return np.array(grouped_keys), grouped_values
```

cattabiani: also docstrings for sort_together and _duplicates_to_list. It is not immediately clear what they are supposed to do

cattabiani (approved):
LGTM

just a few comments (if you want to tackle them). Anyway, not-blocking

PR #304, opened 2025-06-04
Remove `classproperty`, as it doesn't work as expected

mgeplf:
* as discovered here: https://github.com/openbraininstitute/neurodamus/pull/295#discussion_r2120863540
* see https://github.com/python/cpython/issues/89519#issuecomment-1093931546 and https://github.com/python/cpython/issues/89519#issuecomment-1397556641d

### DIFF
--- a/neurodamus/core/_neuron.py
+++ b/neurodamus/core/_neuron.py
@@ -8,7 +8,6 @@ import os.path
 from contextlib import contextmanager
 
 from .configuration import GlobalConfig, NeuronStdrunDefaults, SimConfig
-from neurodamus.utils.pyutils import classproperty
 
 
 #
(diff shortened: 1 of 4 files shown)


neurodamus/core/_neuron.py
```diff
@@ -24,7 +23,7 @@ class _Neuron:
     # No new attributes. __setattr__ can rely on it
     __slots__ = ()
 
-    @classproperty
+    @property
```
juanjosegarcan, line 26: Is not property expecting an instance?

mgeplf: Yeah, it's read only on instances, which is all we can do.

mgeplf: (see the discussions I linked in the commit message)

mgeplf: > I saw it, then is not better to change to def h(self) to be more clear

What sort of change are you thinking of?

juanjosegarcan (approved):
LGTM

cattabiani (approved):
LGTM

PR #333, opened 2025-07-03
do not allow empty population names in CellManagerBase

mgeplf:
* be explicit about what populations are to be managed


cattabiani (approved):
LGTM if we want this. Does anybody know if some scientists still want the functionality that is going to be removed? @jamesgking @WeinaJi

PR #334, opened 2025-07-04
Fix CI for MacOS-15

WeinaJi:
In `simulation_test.yml`,   upgrade setuptools when building h5py, to avoid build error from `project.license`.
This is because Neuron installs requires an older setuptools in the previous step but it doesn't work for h5py.

### DIFF
--- a/.github/workflows/simulation_test.yml
+++ b/.github/workflows/simulation_test.yml
@@ -199,9 +199,9 @@ jobs:
           export HDF5_INCLUDEDIR=/usr/include/hdf5/mpich
           export HDF5_LIBDIR=/usr/lib/x86_64-linux-gnu/hdf5/mpich
         fi
-        # Upgrade setuptools again, to avoid build error of `project.license`
-        # It should have been done in the previous step, no idea why
-        pip install --upgrade setuptools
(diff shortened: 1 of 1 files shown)


.github/workflows/simulation_test.yml
```diff
@@ -200,6 +199,9 @@ jobs:
           export HDF5_INCLUDEDIR=/usr/include/hdf5/mpich
           export HDF5_LIBDIR=/usr/lib/x86_64-linux-gnu/hdf5/mpich
         fi
+        # Upgrade setuptools again, to avoid build error of `project.license`
+        # It should have been done in the previous step, no idea why
```
mgeplf, line 203: Was it because there was a cache hit, so the change in the step wasn't executed?
Perhaps the previous commit probably worked, just the cache needed to be removed first?

WeinaJi: The caches for MacOS haven't been built yet in `main` because of the failure when pushing my PR into `main`. 
In fact, it is NEURON that requires an older setuptools previously which doesn't work for h5py. 
It is complicated when using caches. NEURON uses its cache, and venv has another cache. I wasn't able to see this failure at the end of my PR probably because of using NEURON cache which wouldn't revert the setuptools version at some point.

Issue #146, opened 2025-03-11
ReportingBufferSize is legacy, could be replaced by a CLI option


WeinaJi:
This BlueConfig parameter is used to control how much data to buffer before writing/flushing to disk. We tried to balance frequency of data write for better IO throughput versus how much memory we consumed.
 
However, there isn't any key defined in the SONATA config file for such usage. We may define a CLI option to alter it.

WeinaJi:
At the moment, we've decided to remove it as there is no way to set it via the config file.
Just a reminder, 
1. For neuron, it is set via `self._sonatareport_helper.set_max_buffer_size_hint(reporting_buffer_size)` which calls the function `sonata_set_max_buffer_size_hint` in libsonata-report
2. For Coreneuron, `SimConfig.corenrn_buff_size` is written in `report.conf` and transferred to coreneuron.

WeinaJi:
Implemented by @jamesgking

PR #349, opened 2025-07-30
Updates to add command line argument for changing in memory buffer size for reports

jamesgking:
## Context
Mentioned in https://github.com/openbraininstitute/neurodamus/issues/146
Users should be able to change the memory used for reports to hold data between writes to disk

## Scope
Added element to command.py to receive the value and node.py should receive and apply the value for both classic NEURON and CoreNEURON executions

## Testing
Integration test added which verifies that the reportinglib detects the requested max buffer size

## Review


tests/integration-e2e/test_cli_opts.py
```diff
@@ -165,3 +165,24 @@ def test_cli_output_path():
...
+    custom_env["SPDLOG_LEVEL"] = "debug"
+    result = subprocess.run(
+        ["neurodamus", CONFIG_FILE_MINI, f"--report-buffer-size=64"],
+            check=True,
+            cwd=test_folder_path,
+            capture_output=True,
+            text=True,
+            env=custom_env
+        )
+    assert f"Max Buffer size: 67108864" in result.stdout
```
mgeplf, line 188: no need for an `f` string

tests/integration-e2e/test_cli_opts.py
```diff
@@ -165,3 +165,24 @@ def test_cli_output_path():
     assert not (test_folder_path / simconfig_output_path).is_dir(), \
            f"Directory '{simconfig_output_path}' should NOT exist."
     assert (test_folder_path / output_path).is_dir(), f"Directory '{output_path}' not found."
+
+def test_cli_report_buff_size():
```
mgeplf, line 169: I wonder if this can be a unit test; doesn't look like it needs anything but a circuit; could just be the ringtest, no?

tests/integration-e2e/test_cli_opts.py
```diff
@@ -165,3 +165,24 @@ def test_cli_output_path():
...
+    test_folder = tempfile.TemporaryDirectory("cli-test-report-buff-size")  # auto removed
```
mgeplf, line 170: please don't use `tempfile`; use the `tmp_path` fixture; even better, there is the `create_tmp_simulation_config_file` fixture that does this for you, I think

neurodamus/commands.py
```diff
@@ -67,6 +67,9 @@ def neurodamus(args=None):
         --num-target-ranks=<number>  Number of ranks to target for dry-run load balancing
         --coreneuron-direct-mode     Run CoreNeuron in direct memory mode transfered from Neuron,
                                      without writing model data to disk.
+        --report-buffer-size=<number> Override the size in MB each rank will allocate for each
```
mgeplf, line 70: where is the current default being set?  Could that be built into the logic of the command line, and also in the validation?  That way whoever has to chase it down in the future will find it more easily.

tests/integration-e2e/test_cli_opts.py
jamesgking: This is using libsonatareport for its terminal output to check the buffer size is set, so I'll keep it here for now. In the future it would be better to have a simpler test and put that with the unit tests

tests/integration-e2e/test_cli_opts.py
jamesgking: I'll keep this construct for now to be consistent with the other existing tests. Maybe we can do a proper switch in the future

neurodamus/commands.py
jamesgking: It's a good point that of the several 'default' values in the code, it is not clear which is the one actually used. I'll make another ticket to fix this with some info on my recent code browsing. What I would like is that libsonatareport makes its default accessible to outsize applications so they can query it

tests/integration-e2e/test_cli_opts.py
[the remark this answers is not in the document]
jamesgking: Added

cattabiani: I think you do not need libsonatareport. You can just check SimConfig after neurodamus is called in python. This test is still useful to be sure that that value is really used in the end. 

TLDR: I argue for 2 tests, one unit test to check SimConfig value (probably you need to add 1 or 2 lines to an existing test) and one with libsonatareport to check that really, at the end, the value is used.

tests/integration-e2e/test_cli_opts.py
mgeplf: ah right, good point - forgot that was the dependency.

tests/integration-e2e/test_cli_opts.py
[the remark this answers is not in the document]
cattabiani: this can be tested by checking `report.conf`. No need for an integration test. Something similar to what is in `integration-e2e/reports.py`. However, since you do not need the `run` step you can do it in the unit tests. Otherwise you can follow what is in `unit/test_corenrn_save_restore.py`

tests/integration-e2e/test_cli_opts.py
cattabiani: totally agree. You can check `unit/test_ringcells.py` for inspiration. Or this:

```
@pytest.mark.parametrize("create_tmp_simulation_config_file", [
    {
        "simconfig_fixture": "ringtest_baseconfig",
        "extra_config": {
            "target_simulator": "CORENEURON",
            "node_set": "Mosaic"
        }
    }
], indirect=True)
def test_coreneuron(create_tmp_simulation_config_file):
```

tests/integration-e2e/test_cli_opts.py
mgeplf: ah right, good point - forgot that was the dependency.

tests/integration-e2e/test_cli_opts.py
```diff
@@ -165,3 +165,24 @@ def test_cli_output_path():
     assert not (test_folder_path / simconfig_output_path).is_dir(), \
```
cattabiani, line 165: This comes after reading the test. I think you can check `unit/test_corenrn_save_restore.py` for inspiration. Something like this:

```
@pytest.mark.parametrize("create_tmp_simulation_config_file", [
    {
        "simconfig_fixture": "ringtest_baseconfig",
        "extra_config": {
            "target_simulator": "CORENEURON",
            "inputs": {
                "Stimulus": {
                    "module": "pulse",
                    "input_type": "current_clamp",
                    "delay": 5,
                    "duration": 50,
                    "node_set": "RingA",
                    "represents_physical_electrode": True,
                    "amp_start": 10,
                    "width": 1,
                    "frequency": 50
                }
            },
        }
    }
], indirect=True)
def test_file_placement_base(create_tmp_simulation_config_file):
    command = ["neurodamus", "simulation_config.json"]
    subprocess.run(command, check=True, capture_output=True)

    # Check the output directory: output + save files
    check_dir_content("output", output_content)
    assert not Path("build").exists()
```
- the fixture can change the `simulation_config.json`
- with the subprocess you can add keys to the call
- after you can check the files (useful for coreneuron and `report.conf`

cattabiani (commented):
I think you should add also a way to change the buffer size in the python function. Something here:

https://github.com/openbraininstitute/neurodamus/blob/7e57fdfcd27b99d8a8bcd6a0d2dd2d25474caf13/neurodamus/node.py#L1687

This would allow you to create a simpler unit test too. Something that does not rely on subprocess. The subprocess test is still needed, ok, but another test to test this way of overriding the buffer size should be useful

neurodamus/commands.py
mgeplf: >  What I would like is that libsonatareport makes its default accessible to outsize applications so they can query it

I'm not sure that's necessary; we can just have a default in ND, and always set it to that unless the user requests a different value - basically ignore that that libsonatareport has a default.

cattabiani: why does libsonatareport have a default? It should be set by coreneuron/neurodamus no?

tests/integration-e2e/test_cli_opts.py
jamesgking: Latest commit has adapted to the structure listed

tests/integration-e2e/test_cli_opts.py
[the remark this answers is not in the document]
jamesgking: CoreNEURON test moved to unit test and checks content of report.conf

neurodamus/core/configuration.py
```diff
@@ -944,6 +945,18 @@ def _keep_coreneuron_data(config: _SimConfig):
...
+    user_config = config.cli_options
+    if user_config.report_buffer_size is not None:
+        report_buff_size = int(user_config.report_buffer_size)    
+    else:
+        return
+    
+    assert report_buff_size > 0, "Report buffer size must be > 0"
+    config.report_buff_size = report_buff_size
+
+
```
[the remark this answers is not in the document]
jamesgking: Made consistent

neurodamus/node.py
```diff
@@ -1157,6 +1157,7 @@ def _reports_init(self, pop_offsets_alias):
             CoreConfig.write_spike_filename(file_name)
         else:
             # once all reports are created, we finalize the communicator for any reports
+            self._sonatareport_helper.set_max_buffer_size_hint(SimConfig.report_buffer_size)
```
cattabiani, line 1160: where was this before? We just had this function without using it?

tests/unit/test_cli.py
```diff
@@ -155,3 +156,30 @@ def test_cli_output_path(create_tmp_simulation_config_file):
...
+           },
+       }
+    }
+], indirect=True)
+def test_cli_report_buff_size(create_tmp_simulation_config_file):
+    command = ["neurodamus", create_tmp_simulation_config_file, "--report-buffer-size=64", "--keep-build"]
+    subprocess.run(command, check=True, capture_output=True)
+    
+    report_confs = utils.ReportConf.load("build/report.conf")
+    assert report_confs.reports["soma_v.h5"].buffer_size == 64
```
cattabiani, line 185: we are not testing for:

```
Neurodamus(create_tmp_simulation_config_file, report_buffer_size=64)
```

for example. Other commands got their own test file (like direct mode and save restore). Maybe this one is simple enough that can stay in here... or add a new file. As you prefer

In addition, we could test at least the cli a little more. What happens if:
- buffer size is a prime number? like 11
- buffer size is very big? is everything ok? We never dump data to file in that case?
- is 0? We should check that, indeed, the error is raised
- is negative? same as before
- is empty?
- is not an int/number

cattabiani (commented):
we are very close but I would add a few more tests

tests/integration-e2e/test_cli_opts.py
```diff
@@ -39,3 +40,36 @@ def test_cli_disable_reports(tmp_path):
...
+                    "dt": 0.1,
+                    "start_time": 0.0,
+                    "end_time": 40.0
+                }
+           },
+        }
+    }
+], indirect=True,
+)
+def test_cli_report_buff_size(create_tmp_simulation_config_file):
```
WeinaJi, line 63: Can you also add a integration test for CORENEURON?

neurodamus/core/configuration.py
```diff
@@ -944,6 +945,18 @@ def _keep_coreneuron_data(config: _SimConfig):
     log_verbose("delete_corenrn_data = %s", config.delete_corenrn_data)
 
 
+@SimConfig.validator
+def _report_buffer_size(config: _SimConfig):
```
WeinaJi, line 949: In the case of `--restore`, we will reuse the `report.conf` from save, so `--report-buffer-size` will not work.  Shall we add a warning about that ?

neurodamus/node.py
jamesgking: It was used in stand-alone CoreNEURON

neurodamus/core/configuration.py
jamesgking: I'll simply add the warning now. Changing the buffer size after a restore is not a priority and I can imagine it is not s o straight forward

tests/integration-e2e/test_cli_opts.py
jamesgking: Can I do this in a loop? I don't see anymore in the existing tests to switch between the two modes in a single test. Or should I simply do two functions, one for CoreNEURON, one for NEURON?

tests/integration-e2e/test_cli_opts.py
```diff
@@ -39,3 +40,36 @@ def test_cli_disable_reports(tmp_path):
...
+                    "sections": "soma",
+                    "dt": 0.1,
+                    "start_time": 0.0,
+                    "end_time": 40.0
+                }
+           },
+        }
+    }
+], indirect=True,
+)
```
cattabiani, line 62:
```suggestion
    "create_tmp_simulation_config_file", [
    {
        "simconfig_fixture": "ringtest_baseconfig",
        "target_simulator": "NEURON",
        "extra_config": {
            "reports": {
                "soma_v": {
                    "type": "compartment",
                    "variable_name": "v",
                    "sections": "soma",
                    "dt": 0.1,
                    "start_time": 0.0,
                    "end_time": 40.0
                }
           },
        }
    },
{
        "simconfig_fixture": "ringtest_baseconfig",
        "target_simulator": "CORENEURON",
        "extra_config": {
            "reports": {
                "soma_v": {
                    "type": "compartment",
                    "variable_name": "v",
                    "sections": "soma",
                    "dt": 0.1,
                    "start_time": 0.0,
                    "end_time": 40.0
                }
           },
        }
    }
], indirect=True,
)
```

cattabiani: this creates effectively 2 tests. 1 for neuron, one for coreneuron

tests/integration-e2e/test_cli_opts.py
cattabiani: answer just above :D

neurodamus/core/configuration.py
WeinaJi: If needed, probably should be updated here.
https://github.com/openbraininstitute/neurodamus/blob/387307511431a93fbb29c86bab153bddd0b75973/neurodamus/core/coreneuron_configuration.py#L131
It is not urgent as it is only us that play with the buffer size at the moment. I will let you decide @jamesgking .

cattabiani: that usually deals with the updated tstop. Maybe you can add a check there. I think the safest would be to complain before: if you specify `buffer-size` and `restore` you get an error. I would not use a warning as they are rarely seen

WeinaJi: >that usually deals with the updated tstop.

The function name and the docstring indicate that it could update other parameters in the report.conf, e.g. buffer_size. Right? 

In my previous comment, I meant if we need to update the buffer_size in restore, probably it is in this function.

cattabiani: > In my previous comment, I meant if we need to update the buffer_size in restore, probably it is in this function.

ah I did not understand. I thought it was about the warning since Jim mentioned it just before. 

btw, I am planning to change the way we deal with report.conf because it is a little clunky and prone to errors

tests/integration-e2e/test_cli_opts.py
jamesgking: Yes, seems to do the trick

>tests/integration-e2e/test_cli_opts.py::test_cli_buff_size[create_tmp_simulation_config_file0] PASSED
>tests/integration-e2e/test_cli_opts.py::test_cli_buff_size[create_tmp_simulation_config_file1] PASSED

tests/unit/test_cli.py
cattabiani: yes, that was mainly my point

jamesgking: Added test to confirm error handling of negative and 0 values

neurodamus/core/configuration.py
jamesgking: Added Configuration Error when both are present

cattabiani (approved):
LGTM

PR #357, opened 2025-08-13
move e2e cli tests into unit tests where posible

mgeplf:
* only one that requires report is kept


cattabiani (approved):
LGTM if the CI passes. I see that you kept the report one here. Good because I have a ton of changes for that one and without libsonata it cannot work

PR #366, opened 2025-08-25
disable affected corenrn tests (probably all of them)

cattabiani:
## Context & Scope

`report.conf` has been updated (coreneuron side). #337 has the updates but it is a big pr; it will take time to review. Until then, we temorarly disable the affected tests.


WeinaJi:
You can use `@pytest.mark.skip` to skip the specific test.

cattabiani:
coverall will fail probably because we reduce coverage. I think it should be disregarded in this case

Issue #362, opened 2025-08-18
Investigate change in spike count under certain simulation conditions


jamesgking:
As reported in ticket

https://github.com/openbraininstitute/prod-platform-architecture/issues/103

The recent simulations run have different number of spikes. Determine which changes may have caused, correct, and add tests to avoid future regressions.

cattabiani:
so, I checked a little the changelog and I do not see anything that interpheres with the spike report directly. Only #294 (hash: 5335c3f) may be suspicious if there are some unintended consequences with `_translate_dict` since that was touched and it may change stuff around. What am I referring to in particular? Now we loop over the attrs excluding: `if not x.startswith("__") and not callable(getattr(libsonata_obj, x))`. Before we were looping over everything. Maybe there are some `snake_to_camel` changes that were automatically applied every time. 

In this case we should be able to see changes also for smaller circuits

jamesgking:
I was able to reproduce with a single gid, 198323, the first one that had a new spike 

The change comes from a rogue line when introducing compartment sets where instead of getting only the center of a section for a stimulus, it returned all compartments, increasing the number of stimulus objects created. This impacted large cells (e.g. L5) substantially more.

I will create a PR to fix that line and make a test which add a stimulus object in a many compartment cell and confirm that it is the only one that exists (with correct values) when the sim runs

cattabiani:
this was my fault. A typo when I duplicated this function for the compartment sets removed part of this line.

jamesgking:
For a brief update, I am updating out tests to include a cell with a larger soma (3 compartments)
When inserting stimulus, it should insert only 1 instance and I will give enough current for a set number of APs. Then also test the count stays consistent

PR #369, opened 2025-08-26
Fix when retrieving points in a section when using default values

jamesgking:
For issue https://github.com/openbraininstitute/neurodamus/issues/362

Update bigcell to have larger some which would have more than 1 compartment Test that expected number of stimuli are added when using bigcell

JGK

## Review


juanjosegarcan (commented):
LGTM aside from one comment

tests/simulations/ringtest/morphologies/asc/cell_big.asc
```diff
@@ -1,9 +1,9 @@
 ("CellBody"
     (Color Red)
     (CellBody)
-	(16.5   0.0     0.0)
```
cattabiani, line 4: why this change?

tests/simulations/ringtest/hoc/TestCell.hoc
```diff
@@ -97,6 +98,16 @@ proc biophys(){
 
 proc biophys_inhomo(){}
 
+proc geom_nseg_fixed(/* chunkSize */) { local secIndex, chunkSize
```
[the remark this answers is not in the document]
jamesgking: Thanks, removed secIndex. I keep loop on all secs because send also gets divided into more segments which is more consistent with what we typically run

tests/unit/test_ringcells.py
```diff
@@ -143,3 +145,51 @@ def test_coreneuron(create_tmp_simulation_config_file):
     assert coreneuron_data.is_dir() and not any(coreneuron_data.iterdir()), (
         f"{coreneuron_data} should be empty."
     )
+
+
+@pytest.mark.parametrize(
```
cattabiani, line 150: can we apply the stimulus to something that is not the soma? In that case may we have a test on another section and test the same? The soma is often special. 

Otherwise, discard this comment

cattabiani (commented):
just 2 questions

tests/simulations/ringtest/morphologies/asc/cell_big.asc
jamesgking: The previous soma did not divide into more than 1 compartment

tests/unit/test_ringcells.py
jamesgking: We can add an additional test in the near future. I also want more tests on the number of spikes when using this big soma. I didn't want to drag out getting the fix in for reverting the simulation behavior

cattabiani:
uhm, intersting. Inadvertitely you changed the shm test:

```
        else:
>           assert shm_transfer_message_warning in captured.out
E           AssertionError: assert 'Unknown SHM directory for model file transfer in CoreNEURON.' in 'As of Oct 2020 Neurodamus uses Random123 as default\n'
E            +  where 'As of Oct 2020 Neurodamus uses Random123 as default\n' = CaptureResult(out='As of Oct 2020 Neurodamus uses Random123 as default\n', err='').out
```

at least I can confirm that this is localized to this branch and main works. 

I say that because this morning I did a new libsonata release (I needed new features). However, it looks like the problem is related to this branch. I let you debug

WeinaJi (approved):

tests/simulations/ringtest/hoc/TestCell.hoc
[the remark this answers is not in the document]
juanjosegarcan: Cause in the next method (geom_nsec) we are fixing apical, basal and axonal to 2 segs.

juanjosegarcan (approved):
LGTM

PR #375, opened 2025-09-03
enable `cell_permute` for coreneuron

cattabiani:
## Context
`corenrn` can run with `cell-permute = {0 (cpu, gpu), 1 (cpu, gpu), 2 (gpu)}`. Neurodamus at the moment is not leveraging this. We should allow users to set this. 

**Note:** this is different from `cell-permute` for neuron (9+ only). I may enable that in the future. In that case this will become useless legacy code. At any rate, neuron 9 is new and not fully used everywhere. Proceeding step by step, enabling `cell-permute` in coreneuron is the next logical step

## Scope
TODO

## Testing

Any serious test that involves reports is **blocked** because reports with `cell-permute =/= 0` are currently bugged in coreneuron. There is already a pr that solves this: https://github.com/neuronsimulator/nrn/pull/3587

TODO

## Review


neurodamus/commands.py
```diff
@@ -71,6 +71,11 @@ def neurodamus(args=None):
         --report-buffer-size=<number> Override the size in MB each rank will allocate for each
                                       report buffer to hold data. When the buffer is full, the
                                       ranks will aggregate data for writing to disk. Default: 8 MB
+        --cell-permute=[0, 1]   Cell permutation [default: 0].
```
mgeplf, line 74: instead of using numbers, why not use the actual name?
IE: `--cell-permute=[unpermuted, node-adjacency]`

neurodamus/utils/pyutils.py
```diff
@@ -5,11 +5,37 @@
...
+    __invalid__ = None
+
+    @classmethod
+    def from_string(cls, s: str):
+        if not s:
+            return cls(cls.__default__)
+        mapping = dict(cls.__mapping__)
+        return cls(mapping.get(s.lower(), cls.__invalid__ if cls.__invalid__ is not None else s))
+
+    def to_string(self) -> str:
```
mgeplf, line 27: I hadn't noticed this before, but these methods are strangely named.
`__str__`'s job is `to_string`; and it looks to me that `__str__` is actually `__repr__` in this case.
Doesn't need to be fixed in this PR, just surprised by it.

mgeplf: Let's not change how StrEnumBase works in this PR.

However, the fact that `StrEnumBase` inherits from `IntEnum` is already weird, IMO; but again, separate issue.

mgeplf: I would have the "nice" names everywhere, except at the single point where the permutation is set for neuron, where the is a dict that maps the "nice" version to the required number.

That keeps the code easier to read, imo, and it's only at the interface where the conversion to the neuron specified requirement happens.

cattabiani: but it is like that. the `to_string` is to get the string for neuron compatibility. Do you want a better name? Or you think the neuron one is the nice one? Should I override the str to return something like repr?

neurodamus/commands.py
mgeplf: yeah

neurodamus/core/configuration.py
```diff
@@ -54,7 +54,19 @@ class Feature(Enum):
     LoadBalance = 5
 
 
+class CellPermute(StrEnumBase):
```
WeinaJi, line 57: Why is it based on `StrEnum` not `IntEnum` or even a simple `Enum`?

WeinaJi: ok, thanks for the answer.

PR #378, opened 2025-09-05
compartment sets, stimulus

cattabiani:
## Review


neurodamus/stimulus_manager.py
```diff
@@ -76,7 +95,7 @@ class BaseStim:
 
     IsNoise = False
 
-    def __init__(self, _target, stim_info: dict, _cell_manager):
+    def __init__(self, _tpoints, stim_info: dict, _cell_manager):
```
mgeplf, line 98: I assume `tpoints` means `target_points`? I could only guess that because the previous name was `target`.
If they are going to be renamed, is there a reason not to call them `locations`?

neurodamus/target_manager.py
```diff
@@ -143,6 +143,14 @@ def clear_simulation_data(self):
         # deference SectionRefs to sections
         self._section_access.clear()
 
+    def get_compartment_set(self, name: str):
```
mgeplf, line 146: would it make sense for `TargetManager` to be specialized for compartment instead of having an `_compartment_sets` member that is only used for some compartment sets.

Then the decision to include or not the specified `tpoints` can be delegated to the `TargetManager` without having to detect it by using `stim_info.get("CompartmentSet")` and dispatching differently there.

I hope idea makes sense, and I conveyed it clearly.

neurodamus/stimulus_manager.py
```diff
@@ -48,12 +47,32 @@ def interpret(self, target_spec, stim_info):
...
+        tpoints = self.get_point_list(target_spec, stim_info, cell_manager)
+        stim = stim_t(tpoints, stim_info, cell_manager)
         self._stimulus.append(stim)
 
+    def get_point_list(self, target_spec, stim_info: dict, cell_manager):
+        """Return points from target, using a specified compartment set if given.
+
+        Args:
+            target: Object providing points.
+            stim_info (dict): May contain "CompartmentSet".
```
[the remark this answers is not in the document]
cattabiani: done

neurodamus/stimulus_manager.py
```diff
@@ -48,12 +47,32 @@ def interpret(self, target_spec, stim_info):
...
+            cell_manager: Required for target's point methods.
+
+        Returns:
+            list of points.
```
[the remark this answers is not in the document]
cattabiani: done. I also improved the docs for the TPointList. To appease the type checker I had to move TPointList to the top of the file

neurodamus/stimulus_manager.py
```diff
@@ -48,12 +47,32 @@ def interpret(self, target_spec, stim_info):
...
+        """
+        target = self._target_manager.get_target(target_spec)
+        compartment_set_name = stim_info.get("CompartmentSet")
+        if compartment_set_name:
```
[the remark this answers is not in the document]
cattabiani: I have always found the walrus operator a little confusing because it is an assignment inside an if (ususally used like that) so you do not see it as an assignment at first glance. It is an if. Anyway, changed

neurodamus/target_manager.py
mgeplf: > I find that it increases complexity without really bringing anything useful.
> Python does not enforce input types so you are still able to put the wrong class everywhere as long as you do not touch the parts that are not there.

What I'm trying to convey is that there appears to be a common functionality between the various ways of getting "points", but right now they are merged into a single class that means the consumer has to know which method they are trying to use: either with or without `compartment_sets`.  Instead, I thought, it would be helpful to have an abstraction that does handles the two cases without the consumer needing to know which it is (this is what I mean by the `having to detect it by using stim_info.get("CompartmentSet")`)

>  In addition the target_manager is conceived as a singleton.

Ah, right, that's annoying.  I guess that's the main problem w/ trying to change it around.  A re-architecting of this is beyond this PR, so I'll learn to live with it.

neurodamus/stimulus_manager.py
mgeplf: > tpoints are probably taget points. They are named everywhere as TPoints and stuffed in a list. A TPointList.

Hmm, good to know.  It's too bad that knowledge is gone, because it feels relevant.  Thanks for the explanation.

>  I can rename then as locations

No, that's fine, leave it as it is.

mgeplf (commented):
I can't review it anymore, if @WeinaJi is happy, I'm happy.
If @WeinaJi sheds light on whether `tpoint` == `target_point`, it would be nice to at least replace the `tpoints` arguments to `target_points` (or, if t stands for something else, whatever it stands for :))

neurodamus/stimulus_manager.py
mgeplf: > ah c'mon the knowledge is not gone. I just explained (by re-reading the code) and Weina did this (I think)

Perhaps gone is too harsh; but it requires guessing, at least.

neurodamus/target_manager.py
WeinaJi: I am trying to understand the ideas. Currently in this PR we have
```
        target = self._target_manager.get_target(target_spec)
        if compartment_set_name := stim_info.get("CompartmentSet"):
            compartment_set = self._target_manager.get_compartment_set(compartment_set_name)
            return target.get_point_list_from_compartment_set(
                cell_manager=cell_manager, compartment_set=compartment_set
            )
        return target.get_point_list(cell_manager=cell_manager)
```
`target` is a `NodeSetTarget` that owns `get_point_list_from_compartment_set` and `get_point_list`. Do you want to separate them? That means a `NodeSetTarget` and a `CompartmentSetTarget` and self._target_manager.get_target(target_spec) would pick the right xxxTarget according to the nodeset or compartmentset name?

mgeplf: > That means a NodeSetTarget and a CompartmentSetTarget and self._target_manager.get_target(target_spec) would pick the right xxxTarget according to the nodeset or compartmentset name?

Yes! It was something like that I was thinking of.

neurodamus/stimulus_manager.py
WeinaJi: >I assume tpoints means target_points?

Yes, I think we could name that. It refers to the points of a target.

neurodamus/core/nodeset.py
```diff
@@ -160,6 +160,16 @@ def __init__(self, gids=None, gid_info=None):
...
     max_gid = property(lambda self: self._max_gid)
 
+    def __repr__(self):
+        gids = self.gids(raw_gids=True)
+        n = len(gids)
+        return (
+            f"SelectionNodeSet(n={n}, "
+            f"offset={self.offset}, "
+            f"population={self.population_name}, "
+            f"raw gids={self.gids(raw_gids=True)})"
```
WeinaJi, line 170:
```suggestion
            f"raw gids={gids})"
```
The value of `gids` has already been assigned above.

cattabiani: right. Done

neurodamus/stimulus_manager.py
```diff
@@ -76,7 +103,7 @@ class BaseStim:
 
     IsNoise = False
 
-    def __init__(self, _target, stim_info: dict, _cell_manager):
+    def __init__(self, _tpoints, stim_info: dict, _cell_manager):
```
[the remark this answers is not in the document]
cattabiani: sorry, I forgot to do this. I will do it after lunch

WeinaJi (approved):
LGTM, thanks!

PR #413, opened 2025-10-20
add test to check that direct mode produces the same files as file mode

cattabiani:
## Context
direct mode was never tested properly, with integration tests. It was just tested in the unit tests, with a fake libsonatareport. 

## Scope
Test that coreneuron direct mode returns the same results as file mode. They should be the same but it is nice to check


cattabiani:
@WeinaJi ah now I see. However, this test is 50/50 report/reports. Do you think it is more appropriate to place it there?

tests/integration-e2e/test_reports.py
```diff
@@ -465,3 +465,5 @@ def test_reports_cell_permute(create_tmp_simulation_config_file):
 
         assert r.allclose(r_reference, **(loose_tols if ref_file.name in loose_tol_files else {})), f"The reports differ:\n{file}\n{ref_file}"
 
+
+
```
cattabiani:
```suggestion
```

Issue #414, opened 2025-10-22
check environment at runtime


cattabiani:
We had a long and windy problem on aws recently #393 stemmed from the fact that neurodamus <-> neuron compatibility broke in the recent past. Thus:

- old neuron <-> new neurodamus 
- new neuron <-> old neurodamus

are not compatible anymore. In particular, new neurodamus is not compatible to any neuron below `9.0a-1485-g0d990513b`. 

Unfortunately we cannot change the past so the new thing in these 2 cases must detect the problem. Neuron already complains because the file `report.conf` is malformed. The tricky part is to make new neurodamus complain if old neuron is present. It does not automatically fail because the malformed file produced by neurodamus can still be partially read by neuron and fail to generate reports. 

Thus, we need to check the neuron version (or if you want, the environment) of neurodamus. At least when the program is running. If you want to be super strict you can check only when printing the malformed file (`report.conf`) but I would be a little more generic and check it at runtime, when `Node` is imported.

mgeplf:
What do you mean by `fully interrupting error that states`?  As in the process exits?

Was it added in this:
https://github.com/neuronsimulator/nrn/pull/3641

or some other commit?

cattabiani:
in fact, Judit already encountered this case and neuron correctly died with an error. Unfortunately, for her it was not very communicating and I was on holiday. She wasted 2 or 3 days on this but the error was already there and working as intended. That is the reason I added more context and info to the error in neuron. 

Notice that this is the inverse case of the problem faced by Judit just now. 

This case is: new neuron, old neurodamus
the problem Judit encountered was: new neurodamus, old neuron

As stated before, I cannot add checks in the old versions. I cannot change the past. Thus the new version (either neuron or neurodamus) need to address this problem properly

mgeplf:
> Notice that this is the inverse case of the problem faced by Judit just now.

Ah, that's what I missed; sorry for the noise.

PR #415, opened 2025-10-22
check environment

cattabiani:
## Context
#414 

## Scope
We want to check that neuron is compatible with neurodamus at runtime. How? We:
- create a class `GitVersion` that parses the neuron version
- we compare vs a hardcoded value

## Testing
- add a small unit test to check `GitVersion` and `check_environment`

## To the Reviewers
Let me know if you prefer to put this check somewhere else. The position in `Node.__init__` sounds good to me but I dunno...

### DIFF
--- a/neurodamus/utils/version.py
+++ b/neurodamus/utils/version.py
@@ -16,19 +16,6 @@ from functools import total_ordering
 from packaging import version
 
 
-def check_environment():
-    """Ensure programs in the environment (like neuron) are compatible with Neurodamus."""
-    import neuron
-
-    # Neurodamus requires NEURON >= 9.0a-1485-g0d990513b because
-    # the layout of `report.conf` changed in this commit.
-    req_version = GitVersion("9.0a-1485-g0d990513b")
-
-    curr_neuron_version = GitVersion(neuron.__version__)
-    if curr_neuron_version < req_version:
-        raise RuntimeError(f"NEURON >= {req_version} required, found {curr_neuron_version}")
-
-
 @total_ordering
 class GitVersion:
     """Represents a Git-style version string: base[-commits-gHASH]. Comparison considers
--- a/neurodamus/core/_neuron.py
+++ b/neurodamus/core/_neuron.py
@@ -7,6 +7,7 @@ import os
 from contextlib import contextmanager
 
 from .configuration import GlobalConfig, SimConfig
+from neurodamus.utils.version import GitVersion
 
 
 # Singleton, instantiated right below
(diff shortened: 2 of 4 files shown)


neurodamus/utils/version.py
```diff
@@ -0,0 +1,62 @@
...
+  supporting comparison operators.
+- check_program_version: function to verify a program meets a minimum Git-version requirement.
+
+Intended for use with programs that use Git-describe-style versioning, such as NEURON.
+"""
+
+import re
+from functools import total_ordering
+
+from packaging import version
```
mgeplf: this ads a new dependency, why not just use the regex to pull the first part?

neurodamus/utils/version.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __eq__(self, other):
+        return (self.base, self.commits) == (other.base, other.commits)
+
+    def __lt__(self, other):
+        if self.base != other.base:
+            return self.base < other.base
+        return self.commits < other.commits
+
+    def __hash__(self):
```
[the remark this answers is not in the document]
cattabiani: good question. Ask ruff. either this or I add a no ruleXXX. I thought it was more elegant to add this. It does not hurt

neurodamus/utils/version.py
```diff
@@ -0,0 +1,62 @@
...
+
+import re
+from functools import total_ordering
+
+from packaging import version
+
+
+def check_environment():
+    """Ensure programs in the environment (like neuron) are compatible with Neurodamus."""
+    import neuron
```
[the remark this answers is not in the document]
cattabiani: I want to check at runtime. If this gets imported to do neurodamus --version or something else and neuron is not there it fails even if you want and can get the answer without neuron

neurodamus/utils/version.py
cattabiani: either I lose correct handling of alpha/beta/release candidate suffixes (9.0a, 9.0b1) or I reinvent the wheel to do exactly what they doin packaging. If you prefer, yes, we can do that

neurodamus/utils/version.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Neurodamus requires NEURON >= 9.0a-1485-g0d990513b because
+    # the layout of `report.conf` changed in this commit.
+    req_version = GitVersion("9.0a-1485-g0d990513b")
```
WeinaJi: After NEURON 9 is released, the versioning is 9.0.0-xxx-xxxx. Is "9.0.0" > "9.0a" in the comparison?

neurodamus/utils/version.py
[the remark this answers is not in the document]
mgeplf: ah, it's triggered on classes w/ __eq__; I guess we keep it.

neurodamus/utils/version.py
mgeplf: > Should I add to requirements.txt?

Additional dependencies should go in pyproject.toml.
I'm not sure why there is a requirements.txt; it's not doing anything.

pyproject.toml
```diff
@@ -8,17 +8,18 @@ build-backend = "hatchling.build"
 [project]
 name="neurodamus"
 authors = [
-    {name = "Blue Brain Project, EPFL"},
```
jplanasc: I wonder whether we should not keep BBP as author and add OBI as both author and maintainer?

cattabiani: @mgeplf  @jdcourcol ?

jdcourcol: yes

Issue #343, opened 2025-07-17
"CurrentDir" is legacy parameter and never set in SONATA, to be removed.


WeinaJi:
At the moment `SimConfig.current_dir` is always using the default value, i.e. the folder that contains the simulation config file. That means even if the user launches a simulation in a different directory, it will still write in the directory where the simulation config file is.

```
@SimConfig.validator
def _current_dir(config: _SimConfig, run_conf):
    curdir = run_conf.get("CurrentDir")

    if curdir is None:
        log_verbose("CurrentDir using simulation config path [default]")
        curdir = config.simulation_config_dir
```

PR #416, opened 2025-10-23
remove current_dir since it is unused after swtich to sonata

cattabiani:
## Context
fix: #343 

## Scope
Remove current_dir and refer to `simulation_config_dir`

## Testing
All the tests already do ample use `simulation_config_dir` and if I unassign something incorrectly they complain

### DIFF
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -252,6 +252,10 @@ class _SimConfig:
         cls._parsed_run = cls._config_parser.parsedRun
         cls._simulation_config = cls._config_parser  # Please refactor me
         cls.simulation_config_dir = os.path.dirname(os.path.abspath(config_file))
+        log_verbose(
+            "SimulationConfigDir using directory of simulation config file: %s",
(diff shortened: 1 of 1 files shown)


neurodamus/core/configuration.py
```diff
@@ -803,31 +796,6 @@ def _randomize_gaba_risetime(config: _SimConfig):
...
-                    "Setting CurrentDir to '.' is discouraged and "
-                    "shall never be used in production jobs."
-                )
-            else:
-                raise ConfigurationError("CurrentDir: Relative paths not allowed")
-            curdir = os.path.abspath(curdir)
-        if not os.path.isdir(curdir):
-            raise ConfigurationError("CurrentDir doesnt exist: " + curdir)
-
-    log_verbose("CurrentDir = %s", curdir)
```
WeinaJi: Shall we keep this verbose log somewhere in order to know the path to find the simulation data, e.g. coreneuron data dir? Maybe print the log where `simulation_config_dir` is determined.

PR #422, opened 2025-11-14
Support Allen's point neuron model

WeinaJi:
## Context
Support simulations with the Allen's V1 circuit:
- nodes types: biophysical and point_process
- edges types: Exp2Syn_synapse and point_process

To allow for external customer synapse models (e.g. https://github.com/openbraininstitute/neurodamus-models/tree/main/Allen_V1), Neurodamus removes the check of BBP synapse models `ProbAMPANMDA_EMS.mod`, warning instead of error. Removes the obsolete BGLIBPY_MOD_LIBRARY_PATH env var. 

## Scope
In `allen_point.py`:
- `AllenPointEngine` to handle Allen's point process nodes and edges
- `Exp2SynSynapseRuleManager` and `Exp2SynSynapseReader` to read the Exp2Syn_synapse edges type which uses different synapse parameters and synapse models from the BBP circuit. Currently Neurodamus treats Exp2Syn_synapse edge as projections rather than inner connectivity because of the limitation in sonata_config.py:make_circuit (its node type is still biophysical so it's passed to METypeEngine). But it doesn't change anything for the underlying netcons.

## Testing
- New test env `unit-allen-v1` is created in tox.ini
- New ringtest circuit for allen-v1 model, nodes, edges and synapse replay inputs
- `tests/unit-allen-v1/test_ringcells_allen_v1.py`: check cell states for biophysical cell and point cells.

## Review

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,26 @@
 Change Log
 ==========
 
+Version 4.2.0
+=============
+**23 March 2026**
+
+*New Features*
+  * Add SEClamp voltage levels (#481)
+  * Implement support for condition:modification added to sonata-extension spec (#486)
+  * Add filtering based on ``section_local_ids`` for ``get_point_list`` (#490)
+
+*Improvements*
+  * SONATA spec update: change modifications from CamelCase to snake_case (#484)
+  * ``SonataConfig.parsedConnections`` rework (#447)
+  * Remove ``check_is_subset`` (#488)
+  * Rework of ``modification_manager.py``: unify section type definitions, rename modification classes for clarity (#499)
+
+*Bug Fixes*
+  * Fix offset 0-based (#487)
+  * Fix stale section counts in ``metype.py`` (#499)
+
+
 Version 4.1.0
 =============
 **12 February 2026**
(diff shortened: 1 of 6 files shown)


neurodamus/io/sonata_config.py
```diff
@@ -15,6 +15,15 @@ class ConnectionTypes(str, Enum):
...
     GlioVascular = "GlioVascular"
+    PointProcess = "PointProcess"
+    Exp2Syn = "Exp2Syn"
+
+
+NODE_TYPE_ENGINE_MAP = {
+    "astrocyte": "NGV",
+    "point_process": "AllenPoint",
+    "biophysical": "METype",
+}
```
cattabiani: this violates the plugin style registry: now, if we want to add a new engine, we need to add the class AND change this dict. 

You should find a way to construct this in the metaclass during registration. The class itself, when registering, should know to what nodes it should be associated to

cattabiani (commented):
I have mostly nitpicks apart from my comment on `NODE_TYPE_ENGINE_MAP`. In addition:
- missing the issue you want to close in the description

cattabiani:
@jplanasc no, I still get a circular dep. Jus reinstalled. It works in main. I install with spack. This is a regression

PR #425, opened 2025-12-01
Implement extracellular stimulus: module "spatially_uniform_e_field"

WeinaJi:
## Context
Implements the extracellular stimulus of pattern spatially_uniform_e_field. It allows the summation of multiple `fields` in the same stimulus block, as well as the combination of multiple spatially_uniform_e_field stimulus blocks.

## Scope
1. New class `SpatiallyUniformEField` in `stimulus_manger.py`
2. New class `ElectrodeSource` in `stimuli.py`
3. Function `get_all_segment_points` in `metype.py` to compute the global coordinates of all the segment points

## Testing
1. new unit test in `test_metype.py` for `get_all_segment_points`
2. fix `test_get_section_id` in `test_metype.py` which should assert the section ids of the test cell.
3. new unit tests for the new electric field stimulus
- unit/test_single_extracellular_stimulus.py (single stimulus block)
- unit/test_multiple_extracellular_stimulus.py (multiple stimulus blocks)
- unit-mpi/test_extracellular_stimulus.py (mpi test)
- scientific/test_extracellular_stimulation (for reporting)
4. And the rotation to the current ring test big-cell nodes file, to enable the computation of segments' global coordinates. Should have no impact on the other unit tests.

## Review

## Docs

https://github.com/openbraininstitute/sonata-extension/blob/master/source/sonata_simulation.rst

### DIFF
--- a/neurodamus/stimulus_manager.py
+++ b/neurodamus/stimulus_manager.py
@@ -922,7 +922,7 @@ class SpatiallyUniformEField(BaseStim):
 
     def parse_check_all_parameters(self, stim_info: dict):
         self.duration = float(stim_info["Duration"])  # duration [ms]
-        self.dt = float(SimConfig.run_conf["Dt"])
+        self.dt = float(SimConfig.run_conf.dt)
         self.delay = float(stim_info["Delay"])  # start time [ms]
         if not (np.isclose(self.delay % self.dt, 0) or np.isclose(self.delay % self.dt, self.dt)):
             self.delay = np.ceil(self.delay / self.dt) * self.dt
--- a/tests/unit/test_multiple_extracellular_stimuli.py
+++ b/tests/unit/test_multiple_extracellular_stimuli.py
@@ -818,10 +818,10 @@ def test_neuron_report_with_efields(create_tmp_simulation_config_file, ref_peak)
     n.run()
 
     # Write ASCII reports
-    write_ascii_reports(ascii_recorders, n._run_conf["OutputRoot"])
+    write_ascii_reports(ascii_recorders, n._run_conf.output_root)
 
     # Read ASCII reports
-    report = Path(n._run_conf["OutputRoot"]) / ("compartment_v.txt")
+    report = Path(n._run_conf.output_root) / ("compartment_v.txt")
     assert report.exists()
     data = read_ascii_report(report)
     cell0_voltage_vec = [vec[3] for vec in data if vec[0] == 0]
--- a/tests/unit/test_single_extracellular_stimulus.py
+++ b/tests/unit/test_single_extracellular_stimulus.py
@@ -657,10 +657,10 @@ def test_neuron_report_with_efields(create_tmp_simulation_config_file, ref_peak)
     n.run()
 
     # Write ASCII reports
-    write_ascii_reports(ascii_recorders, n._run_conf["OutputRoot"])
+    write_ascii_reports(ascii_recorders, n._run_conf.output_root)
 
     # Read ASCII reports
-    report = Path(n._run_conf["OutputRoot"]) / ("compartment_v.txt")
+    report = Path(n._run_conf.output_root) / ("compartment_v.txt")
     assert report.exists()
     data = read_ascii_report(report)
     cell0_voltage_vec = [vec[3] for vec in data if vec[0] == 0]


neurodamus/core/configuration.py
```diff
@@ -451,16 +451,6 @@ def _check_params(  # noqa: C901
                 f"simulation config param value is invalid: [{section_name}] {param} = {val}"
             )
 
-    for param, deprecated in (deprecated_values or {}).items():
```
mgeplf: hurray, getting rid of deprecated things

neurodamus/metype.py
```diff
@@ -274,6 +275,69 @@ def local_to_global_coord_mapping(self, points):
...
+            ]
+            ans = get_local_seg_extremes(
+                sec.nseg,
+                ll,
+            )
+            if loc2glob:
+                ans = np.array(
+                    loc2glob(ans),
+                    dtype=float,
+                    order="C",
```
mgeplf: is defining the `order` required? I think one can save a copy by using `np.column_stack` instead of `np.transpose`, if the order makes a big difference; probably worth making a note after benchmarking

tests/simulations/ringtest/neuromodulation/edges_AB.h5
mgeplf: did the edges change?

cattabiani:
uh I did not see that you asked a review on this. I will do it on monday

neurodamus/core/configuration.py
```diff
@@ -524,11 +514,12 @@ def _stimulus_params(config: _SimConfig):
         "Width",
         "Lambda",
         "Weight",
-        "NumOfSynapses",
```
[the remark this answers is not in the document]
WeinaJi: Yes, it was marked as "deprecated" in the legacy BlueConfig [link](https://sonata-extension.readthedocs.io/en/latest/blueconfig.html#blueconfigsection-2). And it is not included in the SONATA spec.

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,143 @@ def parse_check_all_parameters(self, stim_info: dict):
         self.rs = float(stim_info.get("RS", 0.01))  # series resistance [MOhm]
         if self.delay > 0:
             logging.warning("%s ignores delay", self.__class__.__name__)
+
+
+@StimulusManager.register_type
+class SpatiallyUniformEField(BaseStim):
+    """Inject a temporally-oscillating extracellular potential field.
+    The potential field is defined as the sum of multiple potential fields which vary cosinusoidally
+    in time, and whose gradient (i.e., E field) is constant.
```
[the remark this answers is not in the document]
WeinaJi: Done

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,143 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+    """
```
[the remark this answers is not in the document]
WeinaJi: Done

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,87 @@ def attach_to(self, section, position=0.5):
...
+        )
+        self.base_position = base_position
+
+    def add_cosines(self, total_duration, fields, step, **kw):
+        """Add multiple cosinusoidal signals
+        Args:
+            total_duration: total duration, in ms, including ramp-up and ramp-down periods
+            fields: dict of the stimulus fields parameters
+            step: the time step, in ms
+        Returns: a list of sine signal vectors
```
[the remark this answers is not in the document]
WeinaJi: Done

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,87 @@ def attach_to(self, section, position=0.5):
...
+        self.dt = dt
+        self.ramp_up_time = ramp_up_time
+        self.ramp_down_time = ramp_down_time
+        self.signals = self.add_cosines(
+            duration + self.ramp_up_time + self.ramp_down_time,
+            self.fields,
+            step=self.dt,
+            delay=delay,
+        )
+        self.base_position = base_position
```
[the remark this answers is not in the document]
WeinaJi: Done

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,87 @@ def attach_to(self, section, position=0.5):
...
+        self.fields = fields
+        self.dt = dt
+        self.ramp_up_time = ramp_up_time
+        self.ramp_down_time = ramp_down_time
+        self.signals = self.add_cosines(
+            duration + self.ramp_up_time + self.ramp_down_time,
+            self.fields,
+            step=self.dt,
+            delay=delay,
+        )
```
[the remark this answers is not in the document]
WeinaJi: Agree. I have replace `self.signals` with  `signals` on-the-fly inside def `attach_to`. It looks better now.

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,88 @@ def attach_to(self, section, position=0.5):
...
+class ElectrodeSource(SignalSource):
+    """Constructs an extracellular potential field as the sum of multiple user-defined e-fields,
+    and applies the resulting signal to the segment's e_extracellular.
+
+    Args:
+        fields: list of user-defined electric field components (e.g. cosinuoid fields)
+        duration: duration of the signal, not including ramp up and ramp down.
+        ramp_up_time: duration during which the signal amplitude ramps up linearly to 0, in ms
+        ramp_down_time: duration during which the signal amplitude ramps down linearly to 0, in ms
+        base_position: coordinate [x, y, z] of the ground point where potential is 0
```
cattabiani: nitpick: maybe we can point out that it is usually the soma baricenter if not specified? Is it the soma baricenter?

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,87 @@ def attach_to(self, section, position=0.5):
...
-# in extracellular electrode voltage. At this stage only step changes can be used. Gradual,
-# i.e. sinusoidal changes will be implemented in the future
-# After every step has been defined, you have to call initElec() to perform the frequency dependent
-# transformation. This transformation turns e_electrode into e_extracellular at distance d=1 micron
-# from the electrode. After the transformation is complete, NO MORE STEPS CAN BE ADDED!
-# You can then use inject() to first scale e_extracellular down by a distance dependent factor
-# and then vector.play() it into the currently accessed compartment
-#
-# TODO: 1. more stimulus primitives than step. 2. a dt of 0.1 ms is hardcoded. make this flexible!
+class ElectrodeSource(SignalSource):
```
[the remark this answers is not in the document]
cattabiani: lgtm in general. 2 small nitpicks, not blocking

tests/scientific/test_extracellular_stimulation.py
```diff
@@ -0,0 +1,95 @@
...
+                "inputs": {
+                    "ex_efields": {
+                        "input_type": "extracellular_stimulation",
+                        "module": "spatially_uniform_e_field",
+                        "delay": 0,
+                        "duration": 10,
+                        "node_set": "Mosaic",
+                        "fields": [
+                            {"Ex": 50, "Ey": -25, "Ez": 75, "frequency": 100},
+                            {"Ex": 100, "Ey": -50, "Ez": 50, "frequency": 0},
```
cattabiani: nitpick: what about a 3rd frequency, very low like 0.001? So we can check if the results are basically the same as 0 and there are not jumps in the results. Just an idea, if you want

neurodamus/stimulus_manager.py
```diff
@@ -80,7 +84,15 @@ def get_point_list(
             return target.get_point_list_from_compartment_set(
                 cell_manager=cell_manager, compartment_set=compartment_set
             )
-        return target.get_point_list(cell_manager=cell_manager)
+        if stim_info["Pattern"] == "SpatiallyUniformEField":
```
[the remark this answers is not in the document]
cattabiani: yes. May you create an issue so we do not forget?

cattabiani (dismissed):
I have only nitpicks this time. I would prefer to have them addressed but they are not blocking. LGTM

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,104 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+            seg_index = int(np.floor((len(sec_seg_points) - 1) * x))
+            return sec_seg_points[seg_index]
+        return None
+
+    @staticmethod
+    def interp_axon_positions(x, axon_index, soma_position):
+        """Interpolate positions of the axon segment for the given x,
+        because of no 3d point for the new axons.
+        Assume that the axon is oriented along the z-axis from soma, 30 um displaced for 1st axon,
+        60 um for 2nd axon, the same x- and y-coordinates as soma.
```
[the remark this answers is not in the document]
WeinaJi: yes, exactly

neurodamus/metype.py
WeinaJi: ok, I will change to np.column_stack

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,143 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+        self.parse_check_all_parameters(stim_info)
+
+        # apply stim to each point in target_points
+        for target_point_list in target_points:
+            gid = target_point_list.gid
+            cell = cell_manager.get_cell(gid)
+            all_seg_points = cell.compute_segment_global_coordinates()
+            soma = cell.CellRef.soma[0]
+            # soma position is the average of all its 3d points
+            soma_position = np.array(all_seg_points[soma.name()]).mean(axis=0)
```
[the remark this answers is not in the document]
WeinaJi: Ah, I overlooked this comment before. What was your suggestion? 
And I am using the same name as the key of the list `all_seg_points` . I could probably change to sec_id as the key.

neurodamus/core/stimuli.py
WeinaJi: Yes, I can add that in the comment. But there is no default value of base_position, user has to specify it.

neurodamus/stimulus_manager.py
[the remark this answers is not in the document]
WeinaJi: https://github.com/openbraininstitute/neurodamus/issues/429 Done

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,87 @@ def attach_to(self, section, position=0.5):
...
+        self.apply_ramp(stim_vec_sum, self.dt)
+        self.stim_vec.append(stim_vec_sum)
+        self._add_point(self._base_amp)  # Last point
+
+        section.insert("extracellular")
+        seg = section(x)
+        self.stim_vec.play(seg.extracellular._ref_e, self.time_vec, 1)
+        self.signals.clear()  # clear the list to free memory
+
+    def apply_ramp(self, signal_vec, step):
```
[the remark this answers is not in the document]
cattabiani: ok, if you think it is clear it is fine.

joseph-tharayil:
Hi all, I've been testing this code more extensively, and have found that for a single-column-sized simulation, it's using far too much RAM, even for an HPC system to handle. I believe that there are two main reasons for this: the first is the creation of a new `ElectrodeSource` object for each neuron, and the second is the dt of the stimulus. I've addressed both of those issues [here](https://github.com/joseph-tharayil/neurodamus/tree/weji/stimulus_electrodes).

While these workarounds do allow the simulation to run, the increase in the stimulus dt is a nontrivial limitation, since it means we can't use stimulus frequencies above ~50 Hz. Does anyone have any ideas for how to handle this better?

joseph-tharayil:
> Hello @joseph-tharayil , thanks for reporting the issue and proposing the changes.
> 
> How many cells does the single-column-sized simulation contain that shows the memory issue? How much memory does it consume? What is the memory capacity of your system?

The simulation has 30'000 cells. It consumes at least 4 TB of memory (this is the largest system on which I've tested it, but I try as much as possible to avoid using this system and instead use a local machine with 1 TB). 
 
> > the first is the creation of a new `ElectrodeSource` object for each neuron,
> 
> As we inject the extracellular potential on the segment object, i.e.`self.stim_vec.play(seg.extracellular._ref_e, self.time_vec, 1)`, and each segment has different amplitudes (`stim_vec`) because of the its position. I am currently storing all segments' `time_vec` and `stim_vec` to avoid garbage collection. `time_vec` should be common for all so we certainly can save memory for saving it once, just as in your implementation. I see you save `self.extracellulars.append(out)` which is equivalent to save `stim_vec` according to @jamesgking's test.
> 
> > the second is the dt of the stimulus
> 
> I see you fix `dt=2.5` in your implementation. Is that the minimal dt to fit the memory limit?

I haven't tested this extensively, but I believe the minimal dt is between 1 and 2 (on my 1 TB system)

> Meanwhile, I am still working on this PR to add James’s requirement that multiple extracellular stimulus blocks should be additive. This requires summing stim_vec from multiple extracellular stimulus blocks before injecting it into seg.extracellular._ref_e.

joseph-tharayil:
> Hi @joseph-tharayil ,
> 
> One thing to note with having one ElectrodeSource per neuron is that we need to handle the cleanup of `self.time_vec` and `self.stim_vec` for each segment. In your current implementation, they are not cleared, which causes them to grow across segments, see [here](https://github.com/joseph-tharayil/neurodamus/blob/63dac79bbb667729642481a9fb44ca7a71fa2fb2/neurodamus/core/stimuli.py#L529) and [there](https://github.com/joseph-tharayil/neurodamus/blob/63dac79bbb667729642481a9fb44ca7a71fa2fb2/neurodamus/core/stimuli.py#L547). Therefore, the injection on each segment is using the wrong vectors. I will try to solve that.

Thanks for catching this! I looked at my old code, and realized I had a better implementation [here](https://github.com/joseph-tharayil/neurodamus/blob/stimulus-electrodes-testing/neurodamus/core/stimuli.py). It creates a sinusoid when the ElectrodeSource object is initialized, but for each segment, it clones that sinusoid, rescales it, and plays in the clone. I think that should work, right?

WeinaJi:
This PR is ready for review again @cattabiani @mgeplf @jamesgking, thanks!
Now it combines multiple stimulus blocks into a single `SpatiallyUniformEField` instance, where each cell owns an `ElectrodeSource` stimulus object. This object contains a `time_vec` shared by all cell segments and a `segs_stim_vec` dictionary holding the stim_vec for each segment. In the case of multiple stimulus blocks, the cell’s `ElectrodeSource` objects are combined when their node_sets overlap.

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,106 @@ def attach_to(self, section, position=0.5):
...
+        self.duration = duration
+        self.base_position = base_position
+        self.dt = dt
+        self.ramp_up_time = ramp_up_time
+        self.ramp_down_time = ramp_down_time
+        self.segs_stim_vec = {}  # Map of {segment: stimulus_vector} for each cell segment
+        # for delay, add cur_t as the first point, then advance cur_t
+        if delay > 0:
+            self.time_vec.append(self._cur_t)
+            self._cur_t = delay
```
cattabiani: just a question: you say that:

> for delay, add cur_t as the first point, then advance cur_t

however, after, you append cur_t, not pre-pend. I wonder if this is a typo

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,106 @@ def attach_to(self, section, position=0.5):
...
+    """
+
+    def __init__(
+        self, base_amp, delay, duration, fields, ramp_up_time, ramp_down_time, dt, base_position
+    ):
+        self.time_vec = Nd.h.Vector()  # Time points for stimulus waveform
+        self._cur_t = 0
+        self._base_amp = base_amp
+        self._delay = delay
+        self.fields = fields
```
cattabiani: knitpick: fields could be a dict of `FieldConfig` (you can pick also another name) objects so we are sure of what should be inside and what is not. 

Also the inputs could be passed from another config object. So it is clear where we do parsing and where we consume the data.

The correct place to implement this is libsonata. The second best is in `sonata_config.py` or `configuration.py`. 

It is not necessary for this pr but it would be better to move away from passing dictionaries of variables around

neurodamus/metype.py
```diff
@@ -268,12 +270,64 @@ def local_to_global_coord_mapping(self, points):
                 "run neurodamus with `enable_coord_mapping=True`"
             )
         if self.local_to_global_matrix is None:
-            raise Exception("Nodes don't provide all 3d position/rotation info")
+            raise Exception(f"Nodes {self.gid} don't provide all 3d position/rotation info")
```
cattabiani: could we use an error that is less generic? Exception is not a great error to raise

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,282 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+
+            if gid in self.stimList:
+                # Consolidate with existing stimulus
+                cur_es = self.stimList[gid]
+                self.merge_stimuli(cur_es, es)
+            else:
+                # Add new stimulus
+                self.stimList[gid] = es
+
+    def merge_stimuli(self, cur_es, new_es):
```
cattabiani: this should be directly in the class. You could even implement `+=` if you want

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,282 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+        """Update the current ElectrodeSource with a new object
+        1. combine the time_vec
+        2. combine the segs1_stim_vec, if the time over laps, the stim_vec should be summed
+        """
+        t1_vec = cur_es.time_vec.as_numpy()
+        t2_vec = new_es.time_vec.as_numpy()
```
cattabiani: you could keep them as numpy all the time in the class and convert to neuron Vector just when we play, at the end. Way less conversions

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,282 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+
+        if ranges_overlap(t1_vec, t2_vec):
+            # remove the last point which is for reverting signal back to base_amp
+            last_t1, t1_vec = t1_vec[-1], t1_vec[:-1]
+            last_s1, stim1_vec = stim1_vec[-1], stim1_vec[:-1]
+
+            last_t2, t2_vec = t2_vec[-1], t2_vec[:-1]
+            last_s2, stim2_vec = stim2_vec[-1], stim2_vec[:-1]
+
+            combined_time_vec = np.union1d(t1_vec, t2_vec)
```
cattabiani: this is dangerous with floats. consider using a tolerance

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,282 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+    @classmethod
+    def apply_all_stimuli(cls):
+        """Apply all consolidated stimuli to their segments"""
+        if cls._instance:
+            for es in cls._instance.stimList.values():
+                for segment, stim_vec in es.segs_stim_vec.items():
+                    section = segment.sec
+                    if not section.has_membrane("extracellular"):
+                        section.insert("extracellular")
+                    stim_vec.play(segment.extracellular._ref_e, es.time_vec, 1)
```
cattabiani: here you could turn to `Vector` and roto-translate. Only once

tests/unit/test_multiple_extracellular_stimulus.py
```diff
@@ -0,0 +1,697 @@
...
+    0.0,
+]
+
+
+def test_combine_time_stim_vectors():
+    # case 1, overlap, no delay
+    t1_vec = [0, 1, 2, 3, 4, 5, 5]
+    stim1_vec = [10, 10, 10, 10, 10, 10, 0]
+    t2_vec = [3, 4, 5, 6, 7, 7]
+    stim2_vec = [100, 100, 100, 100, 100, 0]
```
cattabiani: in general, use floats for t_vec and stim1_vec. I suspect your implementation is fragile with float comparisons. If you use ints the fragility does not appear

cattabiani (changes requested):
very nice! However, I think we should:

- create a real config object to pass around
- make the `merge_stimuli` more robust (avoid float comparisons)
- reduce the number of roto-translations required
- move the merging in `simuli.py`
- cover some more edge cases in your merge_stimuli tests

cattabiani:
I just noticed another edge case:

first uniform:
tvec= [0, 0.1, 0.2]

second uniform:
tvec [10, 10.1, 10.2]

third uniform:
tvec = [5, 5.1, 5.2]

in `merge_stimuli` we cannot, in general, assume that one of the 2 tvec is contiguous. My toy example may change in this case:

```
def merge_sparse_stimuli(t1_vec, stim1_vec, t2_vec, stim2_vec, dt):
    """
    Merge two (possibly sparse) signals on the same uniform grid efficiently.
    Only the indices corresponding to actual times are stored.
    """

    t_ref = min(t1_vec[0], t2_vec[0])  # any common reference works
    # Map times to integer indices
    i1 = np.rint((t1_vec - t_ref) / dt).astype(int)
    i2 = np.rint((t2_vec - t_ref) / dt).astype(int)

    # Only take the union of indices that actually exist
    all_idx = np.union1d(i1, i2)

    combined_time_vec = t_ref + dt * all_idx
    combined_stim_vec = np.zeros(len(all_idx))

    # Map original indices to positions in all_idx
    pos1 = np.searchsorted(all_idx, i1)
    pos2 = np.searchsorted(all_idx, i2)

    combined_stim_vec[pos1] += stim1_vec
    combined_stim_vec[pos2] += stim2_vec

    return combined_time_vec, combined_stim_vec
    
 ```

File: neurodamus/stimulus_manager.py [hunk 24]
```diff
@@ -812,3 +824,282 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+
+    def parse_check_all_parameters(self, stim_info: dict):
```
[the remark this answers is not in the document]
WeinaJi: as above

tests/unit/test_multiple_extracellular_stimulus.py
```diff
@@ -0,0 +1,697 @@
...
+    )
+    npt.assert_allclose(res_time_vec, [11, 12, 13, 14, 15, 15, 30, 40, 50, 60, 70, 70])
+    npt.assert_allclose(res_stim_vec, [10, 10, 10, 10, 10, 10, 0, 100, 100, 100, 100, 100, 0])
+
+    # case 6, no overlap, t2_vec before t1_vec, with delay
+    t1_vec = [0, 30, 40, 50, 60, 70, 70]
+    stim1_vec = [0, 100, 100, 100, 100, 100, 0]
+
+    t2_vec = [0, 1, 2, 3, 3]
+    stim2_vec = [10, 10, 10, 10, 0]
```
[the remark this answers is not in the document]
WeinaJi: Let's make the decision, dt must be the same as how it is now. Otherwise the merging algorithm will fall.

tests/unit/test_multiple_extracellular_stimulus.py
WeinaJi: You are right, I have changed the combine algorithm based on the time steps. Please check again.

neurodamus/metype.py
WeinaJi: Done

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,282 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+            combined_time_vec = np.append(combined_time_vec, last_t)
+            combined_stim_vec = np.append(combined_stim_vec, last_amp)
+        elif t1_vec[-1] < t2_vec[0]:
+            combined_time_vec = np.append(t1_vec, t2_vec)
+            combined_stim_vec = np.append(stim1_vec, stim2_vec)
+        else:
+            combined_time_vec = np.append(t2_vec, t1_vec)
+            combined_stim_vec = np.append(stim2_vec, stim1_vec)
+
+        if (is_delay1 or is_delay2) and combined_time_vec[0] > 0:
```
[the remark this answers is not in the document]
WeinaJi: Good point, done!

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,282 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+            t2_vec = t2_vec[1:]
+            stim2_vec = stim2_vec[1:]
+
+        def ranges_overlap(t1, t2):
+            """Check if two time ranges overlap."""
+            t1_min, t1_max = t1[0], t1[-1]
+            t2_min, t2_max = t2[0], t2[-1]
+            return not (t1_max < t2_min or t2_max < t1_min)
+
+        if ranges_overlap(t1_vec, t2_vec):
```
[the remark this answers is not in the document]
WeinaJi: Done!

neurodamus/stimulus_manager.py
```diff
@@ -812,3 +824,282 @@ def parse_check_all_parameters(self, stim_info: dict):
...
+        self.stimList = {}  # map of {gid: ElectrodeSource}
+        self.add_new_stimuli(target_points, stim_info, cell_manager)
+        self._initialized = True
+
+    def add_new_stimuli(self, target_points, stim_info, cell_manager):
+        """Process stimulus block for target cells.
+        Creates ElectrodeSource for new cells or merges with existing stimuli.
+        """
+        # parse parameters for the current stimus block
+        self.parse_check_all_parameters(stim_info)
```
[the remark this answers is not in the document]
WeinaJi: Here I follow the BaseStim routine `parse_check_all_parameters`. Sure, in the future, we can refactor it to another place.

tests/unit/test_multiple_extracellular_stimuli.py
```diff
@@ -0,0 +1,697 @@
...
+
+    t2_vec = [0, 1, 2, 3, 3]
+    stim2_vec = [10, 10, 10, 10, 0]
+
+    res_time_vec, res_stim_vec = SpatiallyUniformEField.combine_time_stim_vectors(
+        t1_vec, stim1_vec, t2_vec, stim2_vec, is_delay1=True, is_delay2=False
+    )
+    npt.assert_allclose(res_time_vec, [0, 1, 2, 3, 3, 30, 40, 50, 60, 70, 70])
+    npt.assert_allclose(res_stim_vec, [10, 10, 10, 10, 0, 100, 100, 100, 100, 100, 0])
+
```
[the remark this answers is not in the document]
cattabiani: noice!

cattabiani (commented):
I like thorough testing! I have a couple of suggestions and then it is LGTM!

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,226 @@ def attach_to(self, section, position=0.5):
...
+        # need to introduce any point to avoid interpolation.
+        self._cur_t += duration
+        return self
+
+    def add_cosines(self):
+        """Add multiple cosinusoidal signals
+        Returns: a list of cosine signal vectors
+        """
+        total_duration = self.duration + self.ramp_up_time + self.ramp_down_time
+        tvec = Nd.h.Vector()
```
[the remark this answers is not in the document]
WeinaJi: Because of the delay case when self.time_vec is prepended with [0]. And I need also to apply the ramp up/down to the cosine stim_vec corresponding to the actual `tvec` not including the `t=0` in case of delay. I think that I can not avoid creating an temporary tvec.

pyproject.toml
```diff
@@ -22,13 +22,14 @@ dependencies = [
     "docopt-ng",
     "libsonata",
     "psutil",
-    "packaging"
```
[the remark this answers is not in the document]
WeinaJi: Because of the local-to-global rotation requires `from scipy.spatial.transform import Rotation`
For the extracellular simulation and the multiscale run, we need to either install neurodamus[full] or add scipy to the standard dependency. I remember we made it optional for a specific version of scipy before, but now it seems working fine.

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,205 @@ def attach_to(self, section, position=0.5):
...
+            t2_vec = t2_vec[1:]
+            if not (np.isclose(t2_vec[0] % dt, 0) or np.isclose(t2_vec[0] % dt, dt)):
+                raise ValueError(
+                    f"ElectrodeSource time vector must be divisible by dt {dt}, "
+                    f"check the delay parameter {t2_vec[0]}"
+                )
+            efields2 = efields2[:, 1:]
+
+        # Convert time -> integer ticks
+        t1_ticks = np.round(t1_vec / dt).astype(np.int64)
```
cattabiani: I am not sure that `searchsorted` works also if there are releated values. May you check if either:
- searchsorted works with [0, 1, 2, 2, 2, 3] values?
- or we are sure that there are no repeating ticks? Maybe an assert before may be enough

cattabiani: same as t2_ticks obv

neurodamus/core/stimuli.py
```diff
@@ -491,17 +493,205 @@ def attach_to(self, section, position=0.5):
...
+            combined_time_ticks = np.concatenate((t2_ticks, t1_ticks))
+            combined_efields = np.concatenate((efields2, efields1), axis=1)
+
+        # Convert ticks -> float time
+        combined_time_vec = combined_time_ticks.astype(float) * dt
+
+        if combined_time_vec[0] > 0:
+            # in case of delay add back t=0 stim=0
+            combined_time_vec = np.insert(combined_time_vec, 0, 0.0)
+            combined_efields = np.insert(combined_efields, 0, 0.0, axis=1)
```
cattabiani:
```suggestion
            combined_efields = np.concatenate(
    [np.zeros((combined_efields.shape[0], 1)), combined_efields],
    axis=1
)
```

This works, but it reallocates and can surprise with dtype promotion in other contexts. Better to be explicit

tests/unit/test_multiple_extracellular_stimuli.py
```diff
@@ -0,0 +1,822 @@
...
+    # Read ASCII reports
+    report = Path(n._run_conf["OutputRoot"]) / ("compartment_v.txt")
+    assert report.exists()
+    data = read_ascii_report(report)
+    cell0_voltage_vec = [vec[3] for vec in data if vec[0] == 0]
+    assert len(cell0_voltage_vec) == 21 * 23  # 21 time steps * 23 compartments
+    peaks_pos = find_peaks(cell0_voltage_vec, prominence=1)[0]
+    npt.assert_allclose(peaks_pos, ref_peak)
+    n.clear_model()  # clear up the reporting vector, required for the next run.
```
cattabiani: aaah, this is a nice amount of tests :D

neurodamus/core/stimuli.py
WeinaJi: Done!

cattabiani (dismissed):
LGTM

PR #428, opened 2025-12-11
Use full python version in the CI cache key

WeinaJi:
We retrieve the actual python version in the CI, and use the full version numbers as the cache keys. 
In this way, we don't need to pin any specific python version for caching. And the caches will be rebuilt when the python version is updated in `actions/setup-python`.


mgeplf (approved):

mgeplf:
very nice!

PR #437, opened 2025-12-22
make `neurodamus/node.py` 0 based

mgeplf:


cattabiani:
@mgeplf I suspect it is not strictly a problem. 1-based gids is a subset of 0-based gids. The only problem would arise if somewhere in the code we assume that they are sequential and starting from 0 (when they start from 1). 
>why are we providing the GIDs

I think it is because we need these gids and giving a sequence of numbers is the simplest way. Probably, other distribution methods such as `range(0, 2*n_gids, 2)` would work too

WeinaJi:
>I wonder if SONATA nodesets aren't being applied to virtual populations; so the mismatch in GIDs wouldn't matter in that case.

Yes, that's the case. Neurodamus always creates a`VirtualCellPopulation` with the whole virtual population to compute the offset. I guess gids_vec = [1...1001] is still safe for offsetting the next population, although the correct gids should be [0...1000]. It is better to correct it as this PR does. Thanks! Could you please also update the PR title ?

neurodamus/node.py
```diff
@@ -123,12 +122,9 @@ def register_node_manager(self, cell_manager):
 
     def _new_virtual_node_manager(self, circuit):
         """Instantiate a new virtual node manager explicitly."""
-        storage = libsonata.NodeStorage(circuit.CellLibraryFile)
-        pop_name = circuit.PopulationName
-        node_size = storage.open_population(pop_name).size
-        gid_vec = list(range(1, node_size + 1))
-        virtual_cell_manager = VirtualCellPopulation(pop_name, gid_vec)
-        self.virtual_node_managers[pop_name] = virtual_cell_manager
+        pop = libsonata.NodeStorage(circuit.CellLibraryFile).open_population(circuit.PopulationName)
```
WeinaJi: A nitpick:
how about keeping `pop_name = circuit.PopulationName` instead of of calling `circuit.PopulationName` every time ?

mgeplf: to me it seems more clear, instead of having to look up what `pop_name` is

I think this should go away, anyways, so we handle all the connectivity the same; instead of creating VirtualPopulations out of thin air.

PR #449, opened 2026-01-06
Cleanup rng

mgeplf:
* `RNGConfig` class is never used
* `rngMode` isn't something that can be set through the CLI or SONATA config, so keeping track of it isn't necessary.
* Nearly 2026, no need to explain the 2020 Random123 change any more
* RNGSettings::init hoc function is no longer needed, removed


jamesgking (approved):
LTGM

Issue #460, opened 2026-01-07
`modifications`: name clashes change behavior


cattabiani:
## Description
`test_modifications`

this setup:
```
        {
            "simconfig_fixture": "ringtest_baseconfig",
            "extra_config": {
                "target_simulator": "NEURON",
                "conditions": {
                    "modifications": [
                        {
                            "name": "no_SK_E2",
                            "node_set": "RingA:oneCell",
                            "type": "ConfigureAllSections",
                            "section_configure": "%s.gnabar_hh *= 11; %s.e_pas *= 0.1",
                        }
                    ]
                },
            },
        }
```

gives different results than this:
```
        {
            "simconfig_fixture": "ringtest_baseconfig",
            "extra_config": {
                "target_simulator": "NEURON",
                "conditions": {
                    "modifications": [
                        {
                            "name": "no_SK_E2",
                            "node_set": "RingA:oneCell",
                            "type": "ConfigureAllSections",
                            "section_configure": "%s.e_pas *= 0.1",
                        },
                        {
                            "name": "no_SK_E2",
                            "node_set": "RingA:oneCell",
                            "type": "ConfigureAllSections",
                            "section_configure": "%s.gnabar_hh *= 11",
                        }

                    ]
                },
            },
        }
```

why? because internally we turn modifications from a list to a dict. The keys are the `name` field. What happens if there is a `name` clash? the second modification overrides the first one and the first one is effectively lost. The docs:

https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#parameters-required-for-modifications

Do not require to have different names for different modifications. 

## Solution

We should just use a list. Better if we leverage `sonata` and use the same list. 

This depends on the question: do we need to edit this list after loading with sonata? Some tests do that but I suspect it is just a hack done for testing.

WeinaJi:
Would it be better to check for name clashes in libsonata parser? I am afraid it could also happen to `connection_overrides` which is also a list.

cattabiani:
I thought names where mostly for the users. In this case we can remove them alltogether

cattabiani:
We can report the full block or mention it by position: block 3 block 5 etc. 

Anyway, the name can stay. I understand the reasons more. Do you want a list or a dict? We can do with both. Where do you want to enforce this? In libsonata?

Issue #458, opened 2026-01-07
`SonataConfig.parsedModifications`: dict -> config object


cattabiani:
## Description

- Fix #452 for `parsedModifications`

## Acceptance criteria
- [ ] `sonata_config.parsedModifications` should not be a dict. It should be the libsonata connections object

PR #461, opened 2026-01-07
`SonataConfig.parsedModifications` rework

cattabiani:
## Context

modifications is now a list of dicts

Fix: #458 
Fix: #460 
Fix: #462

## Scope

- make `modifications` a list
- remove validator: it was validating just stuff coming from libsonata. Libsonata has the fields by construction
- adapt configuration and tests
- register modification types using an enum, not the name of the class

## Testing
- add test for #460

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,7 +7,7 @@ SHELL ["/bin/bash", "-c"]
 
 ARG LIBSONATA_TAG=master
 ARG LIBSONATAREPORT_TAG=master
-ARG NEURON_TAG=master
+ARG NEURON_TAG=9.0.1
 ARG NEURON_COMMIT_ID
 ARG WORKDIR=/opt/software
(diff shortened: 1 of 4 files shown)


jamesgking (approved):
LGTM

tests/unit/test_modifications.py
```diff
@@ -1,48 +1,10 @@
...
-
-from neurodamus.core.configuration import ConfigurationError, SimConfig
+from neurodamus.core.configuration import ConfigurationError
 from neurodamus.modification_manager import ModificationManager
 from neurodamus.node import Neurodamus, Node
-
-SIMULATION_CONFIG_FILE = RINGTEST_DIR / "simulation_config.json"
-
-
-def test_applyTTX():
```
WeinaJi: Can we keep the unit test for TTX ?

Issue #457, opened 2026-01-07
`SonataConfig.parsedReports`: dict -> config object


cattabiani:
## Description

- Fix #452 for `parsedReports`

## Acceptance criteria
- [ ] `sonata_config.parsedReports` should not be a dict. It should be the libsonata connections object

PR #463, opened 2026-01-08
`SonataConfig.parsedReports` rework

cattabiani:
## Context

Fix: #457

## Scope

- remove validator in Node. Libsonata does the hard lifting. No need to do it again. We leave the checks about start and end time down the line (check report parameters) as consistency is not checked in libsonata
- we still need `ReportParameters` since some variables (start and end time) can change based on simulator time and cli
- use original libsonata enums throughout the code. No need to reinvent them. `ReportType, Compartment, Sections, Scaling` local enums have been removed because of this
- no translation in sonata config. We just collect to dict. Why is it not a dict in libsonata? 
- add registration method like in the rest of neurodamus and `ReportManager`
- adapt of parsing of `report_on`
- remove test on report parsing as there is no parsing of libsonata config. Everything is used directly

## Tests
Everything was already tested before

### DIFF
--- a/neurodamus/report.py
+++ b/neurodamus/report.py
@@ -9,15 +9,17 @@ from .utils.pyutils import cache_errors
 
 class ReportManager:
     """Registry and factory for Report subclasses based on libsonata report types."""
-    
+
     _report_types = {}
 
     @classmethod
     def register_type(cls, report_type):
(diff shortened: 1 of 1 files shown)


jamesgking (approved):
LGTM

PR #493, opened 2026-03-10
In some rare cases metadata can return None

cattabiani:
## Context

Fix: #494 

We should avoid it otherwise coopt does not exit correctly in case of --version and code exec continues

## Scope

updated `_metadata.py`

### DIFF
--- a/neurodamus/_metadata.py
+++ b/neurodamus/_metadata.py
@@ -2,8 +2,10 @@ import importlib.metadata
 
 try:
     __version__ = importlib.metadata.version("neurodamus")
+    if __version__ is None:
+        __version__ = "devel"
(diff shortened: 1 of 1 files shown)


mgeplf:
> some rare cases 

What are the rare cases?

mgeplf:
> does not change that this is also a neurodamus fragility. It assumes that version is not None

From what I can tell, though, it shouldn't return None for a properly installed package - it seems to be an invariant that is being broken - seemingly only with spack.

I don't know if we should be doing work around for spack if it's not going to install things properly.

To me, the bigger question is why is spack doing this - is the installation recipe wrong? Do other packages that use `pyproject.yaml` and `hatch` not get the right version?

PR #507, opened 2026-04-08
New CLI option to allow dry run and memeory load balancing based on heap memory

### LINKED ISSUE #363: Implement metype size compute based on heap size
## Context

Implement metype size estimation based on the heap size following `memray` approach

## Assets

Neurodamus simulator

## Technical implementation & specs

Resume work from https://github.com/openbraininstitute/neurodamus/tree/metype_heap_size
1. Measure the heap memory of various metype instances in the circuit using `memray` at dry run
2. Load the heap memory with `--memory` load balance mode
3. Local test: Run simulation with the `--memory` load balance and check the memory balance info from the log, e.g.`Memusage (RSS) per task [MB]: Max=xx, Min=xx, Mean(Stdev)=xx`.
4. Run on AWS SSCx circuit hex0 (30K), compare the memory metrics from different loadbalance modes WholeCell vs. Memory on Grafana

## Screenshots
As shown below, the heap size reflects the correct memory consumption curve w.r.t the resident size
Image

## Acceptance criteria

- [ ] Replace the current memory measurement based on the system RSS with the precise heap memory
- [ ] Reasonable load balance in a large simulation
- [ ] Test at scale (30K )on AWS, attach the figures of memory metrics from Grafana

WeinaJi:
## Context
Fix #363, improve memory load balancing to match dry-run memory estimates
1. In the `--dry-run` mode, cells and synapses are included into the estimated memory, and neurodamus produces the allocation file `allocation_r<No_Ranks>_c1.pkl.gz` as well as `cell_memory_usage.json`. 
In the `--lb-mode=Memory` mode, if the allocation file is missing,  run the cell distribution on-the-fly just as in dry run mode. This PR fixes an issue in the `--lb-mode=Memory` mode, that is , when neither allocation file or `cell_memory_usage.json` exists, distributing cells on the fly should include the memory estimates for synapses, just as in dry run.
2. Enable memory measurement by RSS or Heap size via CLI option `--memory-tracker`.
3. When distributing cells across ranks, distribute cells from the most memory consuming to the least. 

## Scope
1. In `memory.py:_memory_mode_load_balancing` function: create synapse manager for all the connections when estimate memory for each metype combination.
2. In `export_cell_memory_usage` function: to reuse `cell_memory_usage.json` to generate the allocation file without instantiating cells and synapses, dump the data dict `metype_cell_syn_average` and `pop_metype_gids` into `cell_memory_usage.json`, together with the existing data `metype_memory`.
4. in `memory.py: distribute_cells`, sort `metype_memory_usage` by its values `total_mem_per_cell` in descending order, and distribute cells from most memory consuming to the least. 
5. CLI option `--memory-tracker` to either measure rss or use memray to measure the heap size per cell metype. Measuring rss is faster, our default option. Measuring heap requires additional dependency (memray), but it would bring indentical measurement from run to run, used for testing.

## Testing
New test `test_lb_mode_memory_from_scratch` in `tests/unit-mpi/test_dry_run.py`

## Review


neurodamus/cell_distributor.py
```diff
@@ -302,47 +305,50 @@ def _instantiate_cells_dry(self, cell_type, skip_metypes, **_opts):
...
+            if SimConfig.memory_tracker == MemoryTracker.HEAP:
+                import memray
+
+                tmp_file = Path(f".tmp_{metype}.bin")
+                dst = memray.FileDestination(tmp_file, overwrite=True, compress_on_exit=True)
+                with memray.Tracker(destination=dst, memory_interval_ms=10000):
+                    _load_cells(cells, cell_type, cell_offset)
+                reader = memray.FileReader(tmp_file)
+                peak_memory = reader.metadata.peak_memory / 1024
+                tmp_file.unlink()
```
cattabiani: this will leave stale files around if an exception occurs in between when you create the files and when you unlink. A try -> except -> finally could solve this

tests/unit-mpi/test_dry_run.py
```diff
@@ -56,15 +54,16 @@ def test_dry_run_memory_use(create_tmp_simulation_config_file, mpi_ranks):
         }
         assert nd._dry_run_stats.metype_counts == expected_metypes_count
         assert nd._dry_run_stats.suggested_nodes > 0
+    assert nd._dry_run_stats.stats_preloaded == False
```
cattabiani:
```suggestion
    assert not nd._dry_run_stats.stats_preloaded
```

ruff could flag this ... I think? Anyway, more pythonic

tests/unit-mpi/test_dry_run.py
```diff
@@ -123,7 +122,54 @@ def test_dry_run_dynamic_distribute(create_tmp_simulation_config_file, mpi_ranks
...
+    Path("allocation_r2_c1.pkl.gz").unlink(missing_ok=True)
+
+    nd = Neurodamus(create_tmp_simulation_config_file, lb_mode="Memory", memory_tracker="heap")
+
+    expected_metypes_count = {
+        'MTYPE1-ETYPE1': 2, 'MTYPE0-ETYPE0': 1,
+        'MTYPE2-ETYPE2': 1, 'MTYPE0-ETYPE1': 1
+    }
+    assert nd._dry_run_stats.metype_counts == expected_metypes_count
+    assert nd._dry_run_stats.stats_preloaded == False
```
cattabiani:
```suggestion
    assert not nd._dry_run_stats.stats_preloaded
```

tests/unit/test_dry_run.py
```diff
@@ -49,6 +51,15 @@ def test_dry_run_memory_use(create_tmp_simulation_config_file):
     assert nd._dry_run_stats.metype_counts == expected_metypes_count
     assert nd._dry_run_stats.suggested_nodes > 0
 
+    assert Path(nd._dry_run_stats._ALLOCATION_FILENAME +"_r2_c1.pkl.gz").exists()
+    assert Path(nd._dry_run_stats._MEMORY_USAGE_FILENAME).exists()
+    assert nd._dry_run_stats.stats_preloaded == False
```
cattabiani:
```suggestion
    assert not nd._dry_run_stats.stats_preloaded
```

docs/source/memory_load_balance.rst
```diff
@@ -21,7 +21,7 @@ Dry Run
 
 This will run the simulation in dry-run mode and collect the memory usage data for
 each gid, for both cells and synapses.
-The data will be stored in the `memory_per_metype.json`, `cell_memory_usage,json` and
+The data will be stored in the `cell_memory_usage,json` and
```
cattabiani: additional curiosity: what happened to this file?
```suggestion
The data will be stored in the `cell_memory_usage.json` and
```

neurodamus/utils/memory.py
```diff
@@ -312,15 +314,31 @@ def collect_all_mpi(self):
...
         with open(self._MEMORY_USAGE_FILENAME, encoding="utf-8") as fp:
-            self.metype_memory = json.load(fp)
+            data = json.load(fp)
+        self.metype_memory = data["metype_memory"]
+        self.pop_metype_gids = {
+            pop: {metype: np.array(gids, dtype="uint32") for metype, gids in metype_gids.items()}
+            for pop, metype_gids in data["pop_metype_gids"].items()
+        }
+        self.metype_cell_syn_average = Counter(data["metype_cell_syn_average"])
+        self.stats_preloaded = True
```
cattabiani: I think cell_memory_usage deserves to be a fully implemented dataclass with to_json/from_json. In this way you segregate the logic and properties of this from memory. In addition:

- The export/import pair must agree on the same keys and conversion logic (numpy→list, list→numpy, dict→Counter). A dataclass puts both directions in one place, so they can't drift apart.
- DryRunStats has ~15 attributes. Only these 3 get serialized. A dataclass makes that boundary visible instead of implicit.
- The tests already assert dryrun_data[key] == getattr(nd._dry_run_stats, key) for each key individually — with a dataclass you'd just compare the whole object.
stats_preloaded is really "did I load this group of 3 fields from disk" — it becomes a natural property: whether the dataclass instance exists or not.

TLDR: the dataclass is more SOLID

neurodamus/utils/memory.py
```diff
@@ -486,6 +504,11 @@ def distribute_cells(
             syns_mem = SynapseMemoryUsage.get_memory_usage(self.metype_cell_syn_average[metype])
             metype_memory_usage[metype] = metype_mem + syns_mem
 
+        # Sort by total memory
+        metype_memory_usage = dict(
+            sorted(metype_memory_usage.items(), key=operator.itemgetter(1), reverse=True)
+        )
+
         def assign_cells_to_bucket(rank_allocation, rank_memory, batch, batch_memory):
```
cattabiani, line 556: may you add a docstring here and report this:

"in the order from the most memory consumer to the least"

It useful to know when you first encounter the function

neurodamus/node.py
```diff
@@ -504,35 +504,50 @@ def _memory_mode_load_balancing(self):
...
-                self.clear_model()
 
             alloc, _, _ = self._dry_run_stats.distribute_cells_with_validation(
                 MPI.size, SimConfig.modelbuilding_steps
             )
+
+            # reset since we instantiated
+            Nd.t = 0.0  # Reset time
+            self.clear_model()
+
```
cattabiani, line 550: shouldn't you reset before distributing with validation? Maybe I am still locked into the old implementation. Why did you move this after `distribute_cells_with_validation`

cattabiani (commented):
The PR is already very solid. I have a few improvements a couple of nitpicks

docs/source/memory_load_balance.rst
WeinaJi: You mean the  original file`memory_per_metype.json`? I don't know. This file doesn't exist in the main branch, and `memory_per_metype` is writen in `cell_memory_usage.json`.  Perhaps it was replaced by `cell_memory_usage.json` and the doc is not updated.

cattabiani: ok

neurodamus/utils/memory.py
WeinaJi: Good point. I've found it was already mentioned the in doc string of function `distribute_cells`, line 482 above 
```
Cells with higher memory load are distributed first.
```
Is that sufficient? 
Interesting it wasn't implemented in the main branch but the doc string already says so.

cattabiani: it is sufficient for someone that operates in `distribute_cells`. However, if someone is stumbling on just this function he/she will not see the comment there. I would add a small docstring here too. Nothing fancy eh, something short

neurodamus/node.py
WeinaJi: I think it doesn't matter whether to `clear_model()` before or afte the distribution. In the `dry-run` workflow, in `run()`,  the cell distribution takes place first, then when all is done, we call `self.cleanup()`. Here I just want to be consistent with dry-run.

neurodamus/cell_distributor.py
WeinaJi: Good point. `try...finally` is added. Please check.

tests/unit-mpi/test_dry_run.py
WeinaJi: The test scripts are not checked by ruff in the CI. But I have adapted your suggestion.

tests/unit-mpi/test_dry_run.py
WeinaJi: same as above

tests/unit/test_dry_run.py
WeinaJi: same as above

neurodamus/utils/memory.py
WeinaJi: Good point. I have added a dataclass `CellMemoryUsage`. Please check.

neurodamus/utils/memory.py
WeinaJi: I just noticed that you want to add docstring to the function `assign_cells_to_bucket` ? That's a different thing, it is to assign a batch of cells (input) to the bucket with the lowest total memory via `heapq.pop`. And the batch of cells are passed to `assign_cells_to_bucket` from the most memory consumer to the least by looping over sorted `metype_memory_usage`. 
I've reordered the lines a bit and added a docstring to `assign_cells_to_bucket`. Please check if it is clear now.

cattabiani (dismissed):
LGTM

cattabiani (approved):
LGTM ^2

PR #508, opened 2026-04-14
Pin NEURON to tag 9.0.1 instead of commit hash

cattabiani:
## Context

@jplanasc suggested to pin neuron to 9.0.1 not the current commit


jplanasc (approved):
LGTM, thanks!

PR #515, opened 2026-04-30
Katta/accept simulation config directly

cattabiani:
## Context

BlueRecording needs to compute electrode weights starting from a circuit config alone, without requiring a full simulation config file on disk. Currently neurodamus only accepts a file path, which forces callers to write temporary files even when the simulation config can be trivially constructed in memory (e.g. via `libsonata.SimulationConfig(json_string, base_path)`).

## Scope

Allow `Neurodamus` / `Node` to receive a pre-built `libsonata.SimulationConfig` object as an alternative to a file path. The object is propagated through `_SimConfig.init`, `_init_config_parser`, and `SonataConfig` without any file I/O. When no file path is available, the `.SUCCESS` marker is skipped and `simulation_config_dir` falls back to `SimulationConfig.base_path`.

## Tests
added some unit tests to check this

### DIFF
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -262,28 +262,17 @@ class _SimConfig:
         from . import NeuronWrapper as Nd
 
         Nd.init()
-
-        import libsonata
-        if isinstance(config_file, libsonata.SimulationConfig):
-            sim_conf = config_file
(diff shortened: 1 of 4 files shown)


cattabiani:
there is a better pr coming

PR #517, opened 2026-05-01
adapt to new libsonata

cattabiani:
## Context

The new libsonata brings some new small but useful features like:
- rejects duplicate modification names
- `list_input_names` now preserves order

This allows us to:
- accept a `libsonata.SimulationConfig` object instead of the path of a file. Useful for last minute patches and dummy simulation_configs. No more juggling with temp files
- simplification in a couple of places of the logic. Stuff is already handled by libsonata. Less bookkeping

## Scope

- Use `libsonata.SimulationConfig.list_input_names` directly instead of re-reading the JSON file to get input order (`_stable_inputs_order` removed)
- `SonataConfig`, `Node`, and `Neurodamus` now accept a `libsonata.SimulationConfig` object in addition to a file path
- Removed double config parsing: `Node.__init__` no longer parses the config separately — `SimConfig.init` handles everything
- Let `libsonata.SonataError` propagate naturally instead of wrapping it in `ConfigurationError`
- Skip `.SUCCESS` file creation when config is passed as an object (no file path available)

## Very small improvements/cosmetics:

- Let `libsonata.SonataError` propagate naturally instead of wrapping it in `ConfigurationError`
- Renamed `config_file` → `config_path_or_obj` with type hints and docstrings

## Testing

- Dropped `test_configure_all_sections_AugAssign_name_clash` — libsonata now rejects duplicate modification names
- Updated `test_modification_wrong_target_multiple_types` to expect `libsonata.SonataError` directly
- Added test: `test_sonata_config_from_simulation_config_object`
- Added test: `test_neurodamus_accepts_simulation_config_object`

## Review

### DIFF
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -262,7 +262,7 @@ class _SimConfig:
         from . import NeuronWrapper as Nd
 
         if isinstance(config_file, str):
-            if not os.path.isfile(config_file):
+            if not Path(config_file).is_file():
                 raise ConfigurationError("Config file not found: " + config_file)
             cls.config_file = config_file
             cls.simulation_config_dir = os.path.dirname(os.path.abspath(config_file))
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -261,18 +261,24 @@ class _SimConfig:
         # Import these objects scope-level to avoid cross module dependency
         from . import NeuronWrapper as Nd
 
-        Nd.init()
-        if not os.path.isfile(config_file):
-            raise ConfigurationError("Config file not found: " + config_file)
-        logging.info("Initializing Simulation Configuration and Validation")
+        if isinstance(config_file, str):
+            if not os.path.isfile(config_file):
+                raise ConfigurationError("Config file not found: " + config_file)
+            cls.config_file = config_file
+            cls.simulation_config_dir = os.path.dirname(os.path.abspath(config_file))
+        else:
+            cls.config_file = None
+            cls.simulation_config_dir = config_file.base_path
 
-        log_verbose("ConfigFile: %s", config_file)
-        log_verbose("CLI Options: %s", cli_options)
-        cls.config_file = config_file
         cls._config_parser = cls._init_config_parser(config_file)
+
+        use_color = (cli_options or {}).pop("use_color", True)
+        log_file = cls._config_parser._sim_conf.output.log_file
+        Nd.init(log_filename=log_file, log_use_color=use_color)
+
+        logging.info("Initializing Simulation Configuration and Validation")
         cls._parsed_run = cls._config_parser.parsedRun
         cls._simulation_config = cls._config_parser  # Please refactor me
-        cls.simulation_config_dir = os.path.dirname(os.path.abspath(config_file))
         log_verbose(
             "SimulationConfigDir using directory of simulation config file: %s",
             cls.simulation_config_dir,
@@ -370,16 +376,11 @@ class _SimConfig:
 
     @classmethod
     def _init_config_parser(cls, config_file):
-        if not config_file.endswith(".json"):
+        if isinstance(config_file, str) and not config_file.endswith(".json"):
             raise ConfigurationError(
                 "Invalid configuration file format. The configuration file must be a .json file."
             )
-        try:
-            config_parser = SonataConfig(config_file)
-        except Exception as e:
-            msg = f"Failed to initialize SonataConfig with {config_file}"
-            raise ConfigurationError(msg) from e
-        return config_parser
+        return SonataConfig(config_file)
 
     @classmethod
     def _init_hoc_config_objs(cls):
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -284,6 +284,8 @@ class _SimConfig:
         cls._config_parser = SonataConfig(sim_config)
 
         logging.info("Initializing Simulation Configuration and Validation")
+        log_verbose("ConfigFile: %s", config_file)
+        log_verbose("CLI Options: %s", cli_options)
         cls._parsed_run = cls._config_parser.parsedRun
         cls._simulation_config = cls._config_parser  # Please refactor me
         log_verbose(
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -257,39 +257,30 @@ class _SimConfig:
     cell_requirements = property(lambda self: self._cell_requirements)
 
     @classmethod
-    def init(cls, config_file, cli_options):
-        # Import these objects scope-level to avoid cross module dependency
+    def init(
+        cls, sim_config: libsonata.SimulationConfig, cli_options, config_file: str | None = None
+    ):
+        """Initialize the simulation configuration.
+
+        Args:
+            sim_config: A ``libsonata.SimulationConfig`` instance.
+            cli_options: A dictionary of CLI options.
+            config_file: The original config file path, or None if a
+                SimulationConfig object was passed directly.
+        """
+        # Import scope-level to avoid cross module dependency
         from . import NeuronWrapper as Nd
 
-        # 1. Resolve to a libsonata.SimulationConfig object
-        if isinstance(config_file, str):
-            if not config_file.endswith(".json"):
-                raise ConfigurationError(
-                    "Invalid configuration file format."
-                    " The configuration file must be a .json file."
-                )
-            if not Path(config_file).is_file():
-                raise ConfigurationError("Config file not found: " + config_file)
-            cls.config_file = config_file
-            cls.simulation_config_dir = str(Path(config_file).resolve().parent)
-            sim_config_obj = libsonata.SimulationConfig.from_file(config_file)
-        elif isinstance(config_file, libsonata.SimulationConfig):
-            cls.config_file = None
-            cls.simulation_config_dir = config_file.base_path
-            sim_config_obj = config_file
-        else:
-            raise ConfigurationError(
-                f"Invalid config_file type: {type(config_file)}. "
-                "Expected a file path (str) or libsonata.SimulationConfig."
-            )
+        # Ensure NEURON is initialized (no-op if already done by Node)
+        Nd.init()
 
-        # 2. Set up logging (before any parsing that may log)
-        use_color = (cli_options or {}).pop("use_color", True)
-        log_file = sim_config_obj.output.log_file
-        Nd.init(log_filename=log_file, log_use_color=use_color)
+        cls.config_file = config_file
+        cls.simulation_config_dir = (
+            str(Path(config_file).resolve().parent) if config_file else sim_config.base_path
+        )
 
-        # 3. Parse the full config (now logging is available)
-        cls._config_parser = SonataConfig(sim_config_obj)
+        # Parse the full config
+        cls._config_parser = SonataConfig(sim_config)
 
         logging.info("Initializing Simulation Configuration and Validation")
         cls._parsed_run = cls._config_parser.parsedRun
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -205,7 +205,6 @@ class LoadBalanceMode(Enum):
 class _SimConfig:
     """A class initializing several HOC config objects and proxying to simConfig"""
 
-    config_file = None
     cli_options = None
     run_conf = None
     output_root = None
@@ -257,34 +256,24 @@ class _SimConfig:
     cell_requirements = property(lambda self: self._cell_requirements)
 
     @classmethod
-    def init(
-        cls, sim_config: libsonata.SimulationConfig, cli_options, config_file: str | None = None
-    ):
+    def init(cls, sim_config: libsonata.SimulationConfig, cli_options):
         """Initialize the simulation configuration.
 
         Args:
             sim_config: A ``libsonata.SimulationConfig`` instance.
             cli_options: A dictionary of CLI options.
-            config_file: The original config file path (e.g.
-                ``"simulation_config.json"``), or None if ``sim_config``
-                was built from a JSON string via
-                ``libsonata.SimulationConfig(json_str, base_path)``.
         """
         # Import scope-level to avoid cross module dependency
         from . import NeuronWrapper as Nd
 
         # Ensure NEURON is initialized (no-op if already done by Node)
         Nd.init()
-        cls.config_file = config_file
-        cls.simulation_config_dir = (
-            str(Path(config_file).resolve().parent) if config_file else sim_config.base_path
-        )
+        cls.simulation_config_dir = sim_config.base_path
 
         # Parse the full config
         cls._config_parser = SonataConfig(sim_config)
 
         logging.info("Initializing Simulation Configuration and Validation")
-        log_verbose("ConfigFile: %s", config_file)
         log_verbose("CLI Options: %s", cli_options)
         cls._parsed_run = cls._config_parser.parsedRun
         cls._simulation_config = cls._config_parser  # Please refactor me
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -266,9 +266,14 @@ class _SimConfig:
                 raise ConfigurationError("Config file not found: " + config_file)
             cls.config_file = config_file
             cls.simulation_config_dir = os.path.dirname(os.path.abspath(config_file))
-        else:
+        elif isinstance(config_file, libsonata.SimulationConfig):
             cls.config_file = None
             cls.simulation_config_dir = config_file.base_path
+        else:
+            raise ConfigurationError(
+                f"Invalid config_file type: {type(config_file)}. "
+                "Expected a file path (str) or libsonata.SimulationConfig."
+            )
 
         cls._config_parser = cls._init_config_parser(config_file)
 
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -261,25 +261,35 @@ class _SimConfig:
         # Import these objects scope-level to avoid cross module dependency
         from . import NeuronWrapper as Nd
 
+        # 1. Resolve to a libsonata.SimulationConfig object
         if isinstance(config_file, str):
+            if not config_file.endswith(".json"):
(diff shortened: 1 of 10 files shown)


neurodamus/core/configuration.py
```diff
@@ -261,18 +261,24 @@ def init(cls, config_file, cli_options):
...
-        Nd.init()
-        if not os.path.isfile(config_file):
-            raise ConfigurationError("Config file not found: " + config_file)
-        logging.info("Initializing Simulation Configuration and Validation")
+        if isinstance(config_file, str):
+            if not Path(config_file).is_file():
+                raise ConfigurationError("Config file not found: " + config_file)
+            cls.config_file = config_file
+            cls.simulation_config_dir = os.path.dirname(os.path.abspath(config_file))
+        else:
```
mgeplf: I think it would be better to have an `elif` with the other type that `config_file` can be and then an `else` with an exception.

neurodamus/io/sonata_config.py
```diff
@@ -93,24 +93,19 @@ class SonataConfig:
...
-            if inputs := json.load(fd).get("inputs", None):
-                self._stable_inputs_order = tuple(inputs.keys())
-            else:
-                self._stable_inputs_order = ()
+    def __init__(self, config_path_or_obj: str | libsonata.SimulationConfig):
+        """Initialize SonataConfig from a file path or a SimulationConfig object.
 
+        Args:
+            config_path_or_obj: Path to a SONATA simulation config JSON file,
+                or a pre-built ``libsonata.SimulationConfig`` object.
```
mgeplf: why the double ``?
I think `a pre-built libsonata.SimulationConfig object` is `an libsonata.SimulationConfig instance`

neurodamus/node.py
```diff
@@ -334,33 +334,34 @@ class Node:
...
-            )
-        import libsonata
-
-        conf = libsonata.SimulationConfig.from_file(config_file)
-        Nd.init(log_filename=conf.output.log_file, log_use_color=options.pop("use_color", True))
+        if isinstance(config_path_or_obj, str):
+            assert config_path_or_obj, "`config_path_or_obj` cannot be empty"
+            if config_path_or_obj.endswith("BlueConfig"):
+                raise ConfigurationError(
+                    "Legacy format BlueConfig is not supported, please migrate to SONATA config"
```
mgeplf: I think we can drop this testing for `BlueConfig`s; they are long gone.

neurodamus/node.py
```diff
@@ -334,33 +334,34 @@ class Node:
...
+                )
 
         # This is global initialization, happening once, regardless of number of
         # cycles
         log_stage("Setting up Neurodamus configuration")
+        SimConfig.init(config_path_or_obj, options)
```
mgeplf: are there potentially issues with the init ordering? For instance, does the `Nd.execute("cvode = new CVode()")` need to come before for the configuration to successfully do its NEURON related checks?

tests/unit/test_sonata_config.py
```diff
@@ -285,3 +285,41 @@ def test_parse_inputs(create_tmp_simulation_config_file):
         "PercentLess": 50.0
     }
     assert SimConfig.stimuli[2] == utils.merge_dicts(SimConfig.stimuli[2], expected_input_subthreshold)
+
+
+def test_sonata_config_from_simulation_config_object():
+    """SonataConfig accepts a pre-built libsonata.SimulationConfig object."""
+    import json
```
mgeplf: why not do this import at the top?

tests/unit/test_sonata_config.py
```diff
@@ -285,3 +285,41 @@ def test_parse_inputs(create_tmp_simulation_config_file):
...
+
+    assert raw_conf.parsedRun.base_seed == 1
+    assert raw_conf.parsedRun.dt == 0.025
+    assert raw_conf.parsedRun.tstop == 10
+
+
+def test_neurodamus_accepts_simulation_config_object():
+    """Neurodamus can be initialised with a libsonata.SimulationConfig object."""
+    import json
+    from neurodamus import Neurodamus
```
mgeplf: is this sensitive `from neurodamus import Neurodamus` being done in the function?

neurodamus/core/configuration.py
cattabiani: yes, done

neurodamus/io/sonata_config.py
cattabiani: for the ``: we use sphinx and they are the standard way to refer to stuff no?

changed the rest

mgeplf: > for the ``: we use sphinx and they are the standard way to refer to stuff no?

I don't think it will resolve correctly; can you check?

neurodamus/node.py
cattabiani: ok

neurodamus/node.py
cattabiani: I checked but I did not find any issues. `SimConfig.init` looks uncorrelated to the rest. I was worried too

tests/unit/test_sonata_config.py
cattabiani: I thought tox with ruff and lint would complain. My bad I guess

tests/unit/test_sonata_config.py
cattabiani: same as before. I was testing stuff, sure that tox would catch any debug code. I was wrong

neurodamus/io/sonata_config.py
cattabiani: Screenshot 2026-05-06 at 09 53 35
 to mee it appears fine... no?

neurodamus/node.py
```diff
@@ -334,33 +334,30 @@ class Node:
     _default_population = "All"
     """The default population name for e.g. Reports."""
 
-    def __init__(self, config_file, options: dict | None = None):
+    def __init__(
+        self, config_path_or_obj: str | libsonata.SimulationConfig, options: dict | None = None
```
WeinaJi, line 338: I thought you wanted to pass a config_path:dict to the Neurodamus class. 
If config_path is a dict, then Neurodamus creates a `libsonata.SimulationConfig(json.dump(dict))`

cattabiani: that was our first thought but it is not enough. What you wrote does not work:

```
    simConf
        .def(py::init<const std::string&, const std::string&>(),
             "string of SimulationConfig JSON"_a,
             "base_path"_a)
```

the `init` signature that takes a string requires an additional value: the path. This meant that I needed to change the number of inputs in neurodamus. Thus, I decided that it was simpler to serve directly a SimulationConfig object that already contains both

WeinaJi: Ok, then it is fine to pass `SimulationConfig` object. I am thinking, maybe we can do some refactoring. 
As `conf = libsonata.SimulationConfig.from_file(config_file)` `conf` is also a `SimulationConfig` object,  we could pass `conf` object directly to SimConfig.init() if we receive a file path. The same as the class SonataConfig in `sonata_config.py`. 
As for the existing tests which uses `SonataConfig(config_file_path)` and `SimSetting.init(config_file_path)`, you can make a new fixture to return a libsonata.SimulationConfig object.
What do you think ?

neurodamus/core/configuration.py
```diff
@@ -261,18 +261,29 @@ def init(cls, config_file, cli_options):
...
+            )
 
-        log_verbose("ConfigFile: %s", config_file)
-        log_verbose("CLI Options: %s", cli_options)
-        cls.config_file = config_file
         cls._config_parser = cls._init_config_parser(config_file)
+
+        use_color = (cli_options or {}).pop("use_color", True)
+        log_file = cls._config_parser._sim_conf.output.log_file
+        Nd.init(log_filename=log_file, log_use_color=use_color)
```
WeinaJi: I prefer to initialize the neuron and the log file at the beginning of Neurodamus constructor. Here the log file would miss the msg" log_stage("Setting up Neurodamus configuration")" in Node.py.

tests/unit/test_modifications.py
```diff
@@ -179,54 +180,6 @@ def test_configure_all_sections_AugAssign(create_tmp_simulation_config_file):
...
-                        }
-
-                    ]
-                },
-            },
-        }
-    ],
-    indirect=True,
-)
-def test_configure_all_sections_AugAssign_name_clash(create_tmp_simulation_config_file):
```
jplanasc, line 211: Why did you remove the test? I think it's still relevant to check that 2 separate modifications end up with the same result as a single, composed one. You just need to rename the `name`, so that they're not the same to fix the test.

jplanasc (changes requested):

tests/unit/test_modifications.py
cattabiani: no, it is already checked in libsonata. There is a test there and here it would not have any additional benefit to check if I get the libsonata error or not

neurodamus/core/configuration.py
cattabiani: Good point. I refactored SimConfig.init to follow a clean sequence:

- Resolve the path/object → always get a libsonata.SimulationConfig object upfront
- Extract the log file path from it and call Nd.init() → logging is now fully set up
- Only then instantiate SonataConfig (which may emit warnings during circuit parsing)

This way there's no window where log messages can be lost. SonataConfig now only accepts a SimulationConfig object (no more dual path/object logic there). The log_stage in Node.__init__ is also moved after SimConfig.init() returns, so it's captured in the log file too.

neurodamus/node.py
cattabiani: **The first part** (resolving to a SimulationConfig object early and passing it through) is already done — it was addressed in the previous commit based on your earlier comment. SimConfig.init now resolves path→object upfront, sets up logging, then passes the object to SonataConfig

**For the test fixture:** I'd prefer to keep it returning a file path — it's useful for debugging when tests fail (you can inspect the JSON). The tests just do libsonata.SimulationConfig.from_file(path) inline before passing to SonataConfig. It's one extra line but keeps things explicit and the file stays on disk if you need it

tests/unit/test_modifications.py
cattabiani: I thought about it and decided to drop it because, in general, we do not re-check things that should be correct from libsonata. Otherwise we should, in theory, add tests for all our dependencies

neurodamus/core/configuration.py
WeinaJi: Why do you want to move Nd.init(log_filename=log_file, log_use_color=use_color) into SimConfig? It intialized Neuron with a lot of essential configurations. I prefer to keep it in the major Neurodamus class. 
SimSettings is supposed to handle the information from simulation configuration file.

cattabiani: Done. I moved Nd.init back to Node.__init__ where it belongs. The flow is now:

- Node._get_simulation_config() resolves path → SimulationConfig object (validates, no double read)
- Nd.init(log_filename=...) — NEURON + logging set up in Node
- SimConfig.init(sim_config_obj, options, config_file=path_or_none) — only does config parsing, receives the object directly

SimConfig.init signature is now (sim_config, cli_options, config_file=None) — the default None means existing tests that call it directly don't need to pass the path. For the tests I added a create_tmp_simulation_config fixture that writes the file (still on disk for debugging) but returns the SimulationConfig object

I let you check if this was what you had in mind

neurodamus/core/configuration.py
```diff
@@ -257,39 +257,30 @@ class _SimConfig:
...
-        # Import these objects scope-level to avoid cross module dependency
+    def init(
+        cls, sim_config: libsonata.SimulationConfig, cli_options, config_file: str | None = None
+    ):
+        """Initialize the simulation configuration.
+
+        Args:
+            sim_config: A ``libsonata.SimulationConfig`` instance.
+            cli_options: A dictionary of CLI options.
+            config_file: The original config file path, or None if a
```
mgeplf: `or None if a  SimulationConfig object was passed directly.`

Based on the signature, that's the only option; can you add the other types that can be passed to `sim_config`?

cattabiani: @mgeplf there are no other types. I do not get the question

tests/unit/test_modifications.py
jplanasc: Hmmm... I disagree. It is not about checking: the test should be correct (by renaming the modification `name`s) and it should check that 2 modifications on the same variable are actually performed.

It's not about checking duplicated keys (which I agree goes to libsonata).

jplanasc: Maybe the test name should be more representative, but the description is clear on what the test is doing.

neurodamus/core/configuration.py
mgeplf: isn't this referring to the `sim_config: A ``libsonata.SimulationConfig`` instance.` argument? which should only take a `libsonata.SimulationConfig` not None?

neurodamus/core/configuration.py
```diff
@@ -257,22 +257,33 @@ class _SimConfig:
     cell_requirements = property(lambda self: self._cell_requirements)
 
     @classmethod
-    def init(cls, config_file, cli_options):
-        # Import these objects scope-level to avoid cross module dependency
+    def init(
+        cls, sim_config: libsonata.SimulationConfig, cli_options, config_file: str | None = None
```
WeinaJi: Do we still need to keep `config_file` in the input list? Is it only used for getting the parent directory? If sim_config obj is create from config_file, sim_config.base_path should contain the path information. 
```
        cls.simulation_config_dir = (
            str(Path(config_file).resolve().parent) if config_file else sim_config.base_path
        )
```

neurodamus/core/configuration.py
cattabiani: `sim_config` -> the object
`config_file` -> the path of the object (if applicable)

> isn't this referring to the

To what are you referring to with this `this` ?

neurodamus/core/configuration.py
```diff
@@ -257,22 +257,33 @@ class _SimConfig:
...
         from . import NeuronWrapper as Nd
 
+        # Ensure NEURON is initialized (no-op if already done by Node)
         Nd.init()
-        if not os.path.isfile(config_file):
-            raise ConfigurationError("Config file not found: " + config_file)
-        logging.info("Initializing Simulation Configuration and Validation")
-
-        log_verbose("ConfigFile: %s", config_file)
-        log_verbose("CLI Options: %s", cli_options)
```
WeinaJi, line 270: Sometimes we need to check the CLI Options. Please keep this msg.

neurodamus/core/configuration.py
mgeplf: the text I quoted before.

`config_file` can either be path or None, right? it's none when `a SimulationConfig object was passed directly.`, the only argument I see that takes a `SimulationConfig` is `sim_config`; either `config_file` is always None, or `sim_config` can be something other than a `SimulationConfig`

tests/unit/test_modifications.py
cattabiani: I checked and the removed test was applying two modifications to **different** variables (e_pas and gnabar_hh), not the same one twice. It was verifying that both entries get applied despite the name clash — which is now a libsonata concern.

That said, I agree we should have a test that verifies two separate modification entries (with different names) both apply to the same variable in order, confirming the last one wins. That's not covered today — test_modification_augassign_multiple_types does sequential ops on the same variable, but within a single section_configure string, not across separate modification entries.

I'll add that test. Good suggestion — it covers a case we were missing regardless of the dropped test.

neurodamus/core/configuration.py
cattabiani: mmm no. Either you pass SimulationConfig and the path or just the SimulationConfig... 

ah you mean that this is ambyguous: `a SimulationConfig object was passed directly.` You implicitly assume "to sim_config" while I meant to Node.init. Good point

cattabiani: I let you check the new comment

neurodamus/core/configuration.py
cattabiani: it is not used just for the parent directory (in that it is already in SimulationConfig). We need the full path of the file to create the `.SUCCESS` file. Maybe I can move the responsibility to node?

neurodamus/core/configuration.py
cattabiani: true. I lost them on the way during development. Re-added

tests/unit/test_modifications.py
jplanasc: > I checked and the removed test was applying two modifications to different variables

Yes, but they belong to the same section, since it applies to `all sections`, so I'd still think that this is not completely useless.

> I'll add that test.

Thanks! That would be even better :)

neurodamus/core/configuration.py
mgeplf: Ok, yes, that is much more clear, I would never have guessed from the old description.

tests/unit/test_sonata_config.py
cattabiani: btw, fixed

jplanasc (commented):
I don't have further comments, I let the other reviewers approve once their conversations are resolved.

cattabiani:
@jplanasc may you remove the blocker of "request changes" if the new test is fine with you?

neurodamus/core/configuration.py
WeinaJi: Yes, I agree to move it to `node.py`. We don't need to call `SimConfig.config_file` in  Neurodamus.__init__   but use the `config_file`  from input parameter directly. It seems clearer, what do you think?

jplanasc (dismissed):
No further comments from my side

neurodamus/core/configuration.py
cattabiani: done. I let you check

WeinaJi (approved):
Looks good from my side.

cattabiani:
@mgeplf can I merge or are you still investigating?

neurodamus/io/sonata_config.py
mgeplf: Ok, I thought you wanted it to be a clickable reference.  We should standardize on which of ` and `` we use, but that can happen some other time.

mgeplf:
> can I merge [...]

If others are happy, I'm happy.

PR #520, opened 2026-05-06
Fix failing test due to libsonata updates: no duplicate keys allowed in sim config

jplanasc:
The latest release of libsonata doesn't allow having duplicate keys in the sim config file. Therefore, we need to fix one test in neurodamus that had a duplicated key.


mgeplf:
Is this covered by: https://github.com/openbraininstitute/neurodamus/pull/517

jplanasc:
Ah, yes, true! Thanks for pointing out!

PR #521, opened 2026-05-15
For Extracellular stimulus, switch from using vector.play to a PointProcess which computes the contribution during runtime, avoiding big memory allocation

jamesgking:
## Context
The initial implementation for extracellular stimulus use vector play which consumed a lot of memory since the values were precomputed for the entire sim duration / stim duration. This switches to use nmodl to calculate the applied values during runtime. This only supports one ElectrodeSource at the moment, with support for additional ones to be added in a subsequent PR

## Scope
For each segment stimulated, a single EFieldIntegrator is  instantiated which is passed the parameters and stored as IvocVec. Using "BEFORE BREAKPOINT" to trigger the calculation, compute the contribution, and set the referenced value.

## Testing
Tests for single ElectrodezSource are adapted to run the configured mod and gather the computed _e_ref value in order to be compared against a reference
Tests for multiple ElectrodeSources are disabled until that supported is restored

## Review


neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,157 @@
...
+            vector_vec(vphase)[i] = vector_vec(argphase)[i];
+            vector_vec(vfreq)[i] = vector_vec(argfreq)[i];
+        }
+    }
+#endif
+ENDVERBATIM
+}
+
+BEFORE BREAKPOINT {
+    LOCAL c, factor, efield_accum
```
WeinaJi: In my local test,  `efield_accum` seems to be initialized as `1`, so with displacementX/Y/Z=0 (i.e. soma), `efield_accum` is 1 instead of 0. So I think we should have a local var with initial 0 here, then pass its final value to `efield_accum`.

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,157 @@
...
+           rdfactor = (t-(delay+ramp_up+duration)) / ramp_down;
+       }
+
+       size = vector_capacity(vX);
+       for( i=0; i<size; i++ ) {
+           if( vector_vec(vfreq)[i] == 0 ) { //constant field
+               _lfactor = cos(2 * 3.141592654 * vector_vec(vfreq)[i] / 1000 * t + vector_vec(vphase)[i] );
+           } else {
+               _lfactor = 1;
+           }
```
WeinaJi:
```suggestion
           _lfactor = cos(2 * 3.141592654 * vector_vec(vfreq)[i] / 1000 * (t-delay) + vector_vec(vphase)[i] );
```
`t` -> `t-delay` ; 
And the constant field is already considered in the cosine wave when freq=0, the value of the constant depends on the `phase` value provided by user. If `phase=0`, `_lfactor = 1`.

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,154 @@
...
+       auto *vZ = *reinterpret_cast<IvocVect**>(&_p_Z);
+       auto *vphase = *reinterpret_cast<IvocVect**>(&_p_phase);
+       auto *vfreq = *reinterpret_cast<IvocVect**>(&_p_frequency);
+
+       // TODO: if in a ramp up/down window, then scale further
+       if( delay < t && t < delay+ramp_up ) {
+           rufactor = (t-delay) / ramp_up;
+       }
+       if( delay + ramp_up + duration < t && t < delay+duration+ramp_up+ramp_down ) {
+           rdfactor = (t-(delay+ramp_up+duration)) / ramp_down;
```
WeinaJi:
```suggestion
           rdfactor = 1 - (t-(delay+ramp_up+duration)) / ramp_down;
```

tools/efield_demo.py
```diff
@@ -0,0 +1,58 @@
...
+ramp_up = 20
+ramp_down = 30
+
+es_xvec = h.Vector([1, 2, 3])
+es_yvec = h.Vector([2, 4, 5])
+es_zvec = h.Vector([4, 8, 7])
+es_freqvec = h.Vector([10, 50, 100])
+es_phasevec = h.Vector([10, -10, 0])
+
+obj.add_electrode_source( delay, dur, ramp_up, ramp_down, es_xvec, es_yvec, es_zvec, es_freqvec, es_phasevec )
```
WeinaJi:
```suggestion
obj.add_electrode_source( delay, dur, ramp_up, ramp_down, es_xvec, es_yvec, es_zvec, es_phasevec, es_freqvec )
```

neurodamus/core/stimuli.py
```diff
@@ -533,72 +526,43 @@ def delay(self, duration):
...
             section = segment.sec
-            e_ext_vec = Nd.h.Vector(self.compute_potentials(displacement))
             if not section.has_membrane("extracellular"):
                 section.insert("extracellular")
-            # Neuron vector play without interpolation, continuous=0
-            e_ext_vec.play(segment.extracellular._ref_e, self.time_vec, 0)
-            self.segment_potentials.append(e_ext_vec)
+
+            # convert field data into multiple hoc vectors
+            freqVec = Nd.h.Vector(len(self.fields))
```
mgeplf: please use python casing: ie: `freq_vector`

tools/efield_demo.py
jamesgking: removed the efield_demo

neurodamus/core/stimuli.py
```diff
@@ -533,72 +526,43 @@ def delay(self, duration):
...
+            phaseVec = Nd.h.Vector(len(self.fields))
+            xVec = Nd.h.Vector(len(self.fields))
+            yVec = Nd.h.Vector(len(self.fields))
+            zVec = Nd.h.Vector(len(self.fields))
+            for fieldIndex, field in enumerate(self.fields):
+                freqVec.x[fieldIndex] = field.get("Frequency", 0)
```
mgeplf: why are defaults needed, doesn't libsonata always have a value?

neurodamus/core/stimuli.py
```diff
@@ -608,86 +572,6 @@ def cleanup(self):
...
-            self.efields,
-            other.time_vec.as_numpy(),
-            other.efields,
-            self._delay > 0,
-            other._delay > 0,
-            self.dt,
-        )
-        self.time_vec = Nd.h.Vector(combined_time_vec)
+        """Combined with another ElectrodeSource object"""
+        logging.error("Multiple ElectrodeSource not supported with nmodl")
```
mgeplf, line 575: shouldn't the process fail if the input isn't handled?

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+    if( enabled ) {
+        e_ext = 0
+    }
+}
+
+PROCEDURE add_electrode_source() {
+    : receive data elements for an ElectrodeSource
+    : delay, duration, ramp_up_time, ramp_down_time
+    : vectors for n fields: x, y, z, freq, phase,
+
```
mgeplf: trailing whitespace, and in a few other places

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+PROCEDURE add_electrode_source() {
+    : receive data elements for an ElectrodeSource
+    : delay, duration, ramp_up_time, ramp_down_time
+    : vectors for n fields: x, y, z, freq, phase,
+    
+VERBATIM
+#ifndef CORENEURON_BUILD
```
mgeplf, line 58: we should have a guard that fails if this is used with coreneuron, since the results will be wrong

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+    auto* argX = vector_arg(5);
+    auto* argY = vector_arg(6);
+    auto* argZ = vector_arg(7);
+    auto* argphase = vector_arg(8);
+    auto* argfreq = vector_arg(9);
+
+    size = vector_capacity(argX);
+    if( _p_phase != nullptr ) {
+        fprintf( stderr, "support for multiple ElectrodeSource info not implemented yet\n" );
+        //TODO: vector_resize( vphase, vector_capacity(vec) );
```
mgeplf, line 73: I don't understand this `TODO`

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
+NEURON {
+    POINT_PROCESS EFieldIntegrator
+    POINTER e_ext
+    :POINTER delay, duration, ramp_up, ramp_down, nfields : TODO: vectors will have size = nElectrodeSource
```
mgeplf: does this need to be commented?

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+    POINTER phase, frequency, X, Y, Z           : TODO: will have size = sum(nFields for each ElectrodeSource)
+    RANGE delay, duration, ramp_up, ramp_down
+    RANGE displacementX, displacementY, displacementZ
+    RANGE enabled
```
mgeplf: is this needed, it's always 1 from what I can tell

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+    
+VERBATIM
+#ifndef CORENEURON_BUILD
+    delay = *getarg(1);
+    duration = *getarg(2);
+    ramp_up = *getarg(3);
+    ramp_down = *getarg(4);
+
+    // copy data into locally managed vectors rather than storing pointer
+    //  the idea is to put all data into a single vector and track offsets
```
mgeplf: what do you mean by `track offsets`? if it's the `soa_container`, I don't think requires a comment

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+
+    int i, size;
```
mgeplf: these can be declared at time of use, to limit their scope, imo

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+            vector_vec(vphase)[i] = vector_vec(argphase)[i];
+            vector_vec(vfreq)[i] = vector_vec(argfreq)[i];
+        }
+    }
+#endif
+ENDVERBATIM
+}
+
+BEFORE BREAKPOINT {
+    LOCAL factor, efield_accum
```
mgeplf: is `factor` used?

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+       }
+
+       size = vector_capacity(vX);
+       for( i=0; i<size; i++ ) {
+           _lfactor = cos(2 * 3.141592654 * vector_vec(vfreq)[i] / 1000 * (t-delay) + vector_vec(vphase)[i] );
+	   _lefield_accum += 1e3 * rufactor * rdfactor * (displacementX * vector_vec(vX)[i]*_lfactor + displacementY * vector_vec(vY)[i]*_lfactor + displacementZ * vector_vec(vZ)[i]*_lfactor);
+       }
+    }
+    
+/* TODO: Adapting to multiple ElectrodeSources
```
mgeplf: please remove dead code

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+BEFORE BREAKPOINT {
+    LOCAL factor, efield_accum
+
+    : for each electrode source (pending)
+    :    for each field
+    :       in time window? compute contribution factor and apply to segment's e_ref via pointer
+VERBATIM
+#ifndef CORENEURON_BUILD
+    int i, size;
+    double rufactor=1, rdfactor=1;
```
mgeplf: since `{ru,rd}factor` are mutually exclusive (right?), can they be consolidated into a `ramp_factor`?

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+       auto *vY = *reinterpret_cast<IvocVect**>(&_p_Y);
+       auto *vZ = *reinterpret_cast<IvocVect**>(&_p_Z);
+       auto *vphase = *reinterpret_cast<IvocVect**>(&_p_phase);
+       auto *vfreq = *reinterpret_cast<IvocVect**>(&_p_frequency);
+
+       // TODO: if in a ramp up/down window, then scale further
+       if( delay < t && t < delay+ramp_up ) {
+           rufactor = (t-delay) / ramp_up;
+       }
+       if( delay + ramp_up + duration < t && t < delay+duration+ramp_up+ramp_down ) {
```
mgeplf: please use consistent spacing between operators; applies to whole file

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+PARAMETER {
+}
+
+ASSIGNED {
+    e_ext (mV)
+    delay
+    duration
+    ramp_up
+    ramp_down
+    nfields
```
mgeplf: never used?

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+       if( delay < t && t < delay+ramp_up ) {
+           rufactor = (t-delay) / ramp_up;
+       }
+       if( delay + ramp_up + duration < t && t < delay+duration+ramp_up+ramp_down ) {
+           rdfactor = 1 - (t-(delay+ramp_up+duration)) / ramp_down;
+       }
+
+       size = vector_capacity(vX);
+       for( i=0; i<size; i++ ) {
+           _lfactor = cos(2 * 3.141592654 * vector_vec(vfreq)[i] / 1000 * (t-delay) + vector_vec(vphase)[i] );
```
mgeplf: `factor` doesn't need to be defined as `LOCAL`, since it can be declared locally here, I think.

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,149 @@
...
+	   _lefield_accum += 1e3 * rufactor * rdfactor * (displacementX * vector_vec(vX)[i]*_lfactor + displacementY * vector_vec(vY)[i]*_lfactor + displacementZ * vector_vec(vZ)[i]*_lfactor);
```
mgeplf: we could pre-compute all the `displacement * X` values, and only store those - that would save us 3 multiples per field per iteration, I think.

tests/unit-mpi/test_extracellular_stimulus.py
```diff
@@ -4,14 +4,16 @@
...
-from neurodamus import Node
+from neurodamus import Neurodamus
 from neurodamus.core import MPI
 from neurodamus.stimulus_manager import SpatiallyUniformEField
 
 REF_CONSTANT_BIGCELL = np.array(
     [
         -0.0,
-        -0.482168,
+        -0.160722,
```
mgeplf, line 14: are these changes in the reference values safe?

I think we should add a test just for the mod file, and compare it to the initially implementation that Weina did, so that we know where're close, especially across various `dt` values.  Something like running `np.interp` on both recorded extracellular values, to rebase them to the same time points, and then find the maximum and average differences.

That way we can be confident that this approach matches, and what the maximum differences are.

neurodamus/core/stimuli.py
cattabiani: +1

neurodamus/data/mod/efield_integrator.mod
cattabiani: noob question: doesn't neuron provide a value for pi?

cattabiani: nitpick: possibly multiplying by 0.001 (or 1e-3) is better than dividing by 1000

tests/unit/test_single_extracellular_stimulus.py
```diff
@@ -8,62 +8,12 @@
...
 from tests.utils import read_ascii_report, record_compartment_reports, write_ascii_reports
 
-from neurodamus import Neurodamus, Node
+from neurodamus import Neurodamus
 from neurodamus.core.configuration import ConfigurationError
 from neurodamus.core.stimuli import ElectrodeSource
 from neurodamus.stimulus_manager import SpatiallyUniformEField
 
 
-def test_apply_ramp():
```
cattabiani, line 17: why did you remove this test? If it is because we are applying multiple electrodeSource you can trivially split it in 3 tests

neurodamus/data/mod/efield_integrator.mod
mgeplf: > nitpick: possibly multiplying by 0.001 (or 1e-3) is better than dividing by 1000

Those aren't the same operations: when I wrote this initially, it was on purpose to gain as close to the right value as possible.

cattabiani: yes, one is faster than the other. About the "right" value... we do not have a reference do we? So we can call whatever comes from the multiplication the "right" value ... no?

mgeplf: > About the "right" value... we do not have a reference do we? 

No, we know exactly how many milliseconds are in a second, so why not use the more precise value?

cattabiani: yes, 1 millisecond is exactly 0.001 or 1e-3 seconds. So... does this mean that you agree with me?

mgeplf: > yes, 1 millisecond is exactly 0.001 or 1e-3 seconds

but 0.001 is not representable in floating point, so the factor should be a division if the precise value is desired.

cattabiani: anyway, as I said, nitpick. I am not sure what is the best solution in this particular case. division by 1000 could be actually secretly removed by the compiler

tests/unit-mpi/test_extracellular_stimulus.py
WeinaJi: In `tests/unit/test_single_extracellular_stimulus.py`: I've added additional checks in `test_one_field_noramp` to ensure the references are following the cosine waves as computed in the mod file, for both references of the old vecplay and in the current mod. In `test_one_field_withramp`, the same check is added including ramp_up/down for the current implementation. I think these checks can ensure the validity of the references.
BTW, I didn't add the check for the vecplay references with ramping because I've found that the ramp up time was incorrect (it included t=0 as ramp_up) https://github.com/openbraininstitute/neurodamus/blob/c30a41d49c3a6d459cad47dbd6bf3c2dadc8ac77/neurodamus/core/stimuli.py#L586

neurodamus/data/mod/efield_integrator.mod
mgeplf: >  division by 1000 could be actually secretly removed by the compiler

Under what circumstances? I don't think that is the case.

mgeplf: (ignoring -ffast-math, which is not considered good practice for anything that does numerical methods)

cattabiani: I was not ignoring that ahaha

jamesgking: re: pi, I found example for how to access a built in copy
https://www.neuronsimulator.org/en/latest/nmodl/NMODL_language.html

tests/unit/test_single_extracellular_stimulus.py
WeinaJi: The ramping procedure has been moved to the MOD file,  so this test is not needed.

neurodamus/core/stimuli.py
jamesgking: now raise ConfigurationError (hopefully support comes soon)

tests/unit/test_single_extracellular_stimulus.py
cattabiani: I meant that we should test that the ramping should work. I understand that the test, as is, cannot work anymore since it was referencing inner stuff that is not there anymore. However, we should at least test that the ramping is working either with a unit test of the mod file of some sort or an integration test

neurodamus/data/mod/efield_integrator.mod
WeinaJi: We raise an exception on the python side https://github.com/openbraininstitute/neurodamus/blob/c30a41d49c3a6d459cad47dbd6bf3c2dadc8ac77/neurodamus/core/configuration.py#L942 that extratracellular stimulus with CoreNEURON will be rejected at the beginning of the simulation.

tests/unit/test_single_extracellular_stimulus.py
WeinaJi: I've added a check in `test_one_field_withramp`, see comment https://github.com/openbraininstitute/neurodamus/pull/521#discussion_r3260757768

neurodamus/data/mod/efield_integrator.mod
WeinaJi: I agree. But if Jim prefers to keep it atm for testing and validation, we can improve that later when implementing the multiple stimulus blocks.

neurodamus/data/mod/efield_integrator.mod
jamesgking: Will be removed with multiple ElectrodeSource implementations

neurodamus/data/mod/efield_integrator.mod
jamesgking: I'll keep as is for clarity

neurodamus/core/stimuli.py
jamesgking: I retained existing assignment from initial implementation

neurodamus/data/mod/efield_integrator.mod
jamesgking: Will be removed with multiple ElectrodeSource implementations

neurodamus/data/mod/efield_integrator.mod
mgeplf: I find it less clear; this way it implies there can be interactions between them, even though they are mutually exclusive.

neurodamus/core/stimuli.py
mgeplf: ok, but are they necessary? can you check? we potentially have the opportunity to clean things up and make them more clear, so we should take it IMO.

neurodamus/data/mod/efield_integrator.mod
jamesgking: removed comments

neurodamus/core/stimuli.py
WeinaJi: I've just pushed the change.

mgeplf: cool, thanks

neurodamus/data/mod/efield_integrator.mod
jamesgking: fixed a few; will need long term solution

neurodamus/data/mod/efield_integrator.mod
jamesgking: removed

neurodamus/core/stimuli.py
```diff
@@ -533,72 +527,43 @@ def delay(self, duration):
...
-            e_ext_vec.play(segment.extracellular._ref_e, self.time_vec, 0)
-            self.segment_potentials.append(e_ext_vec)
+
+            # convert field data into multiple hoc vectors
+            freq_vector = Nd.h.Vector(len(self.fields))
+            phase_vector = Nd.h.Vector(len(self.fields))
+            x_vec = Nd.h.Vector(len(self.fields))
+            y_vec = Nd.h.Vector(len(self.fields))
+            z_vec = Nd.h.Vector(len(self.fields))
+            for fieldIndex, field in enumerate(self.fields):
```
mgeplf: `fieldIndex` is still funnily cased (note that we're ignoring `non-lowercase-variable-in-function` with ruff, but that's meant to be for old code); is there a reason that it's not just `i`?

neurodamus/data/mod/efield_integrator.mod
jamesgking: We don't enforce that they don't overlap, so can't make this assumption
I will ask scientists if we should throw an error for any configurations that don't give sufficient time spans

neurodamus/core/stimuli.py
jamesgking: Thanks @WeinaJi for the additional fixes

jamesgking: Updated loop iteration to use standard i

neurodamus/data/mod/efield_integrator.mod
mgeplf: > We don't enforce that they don't overlap, so can't make this assumption

The way it's written, though, is that they don't overlap, no:
`rufactor` only happens until `t < delay+ramp_up`, and `rdfactor` starts `delay+ramp_up+duration < t`

duration is great than 0, right? So must not overlap?

neurodamus/data/mod/efield_integrator.mod
mgeplf: This still hasn't been addressed.

neurodamus/core/stimuli.py
mgeplf: thanks

neurodamus/data/mod/efield_integrator.mod
jamesgking: Added comment related to mcomplex.dat

neurodamus/data/mod/efield_integrator.mod
jamesgking: Now I understand better. Thanks. Updated

neurodamus/data/mod/efield_integrator.mod
mgeplf: interesting/weird; how can I play w/ this?
will the tests fail if I remove `enabled`?

jamesgking: Without this flag, the load balance test will fail when it goes to create mcomplex.dat

mgeplf: huh, weird, ok. Thanks.

neurodamus/data/mod/efield_integrator.mod
jamesgking: Yes, I will include this in the next PR

jamesgking: okay, you all convinced me and I made the change requested

mgeplf: thanks

mgeplf (commented):
Code looks reasonable to me, I'll leave final approval to someone more knowledgeable about the behavior.

cattabiani:
ok, let me check one more time

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,131 @@
+NEURON {
+    POINT_PROCESS EFieldIntegrator
+    POINTER e_ext
```
cattabiani, line 3: so we do not want to support coreneuron? BBCOREPOINTER is clunkier to use?

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,131 @@
...
+    :POINTER delay, duration, ramp_up, ramp_down : TODO: add nfields vector and vectors will have size = nElectrodeSource
+    POINTER phase, frequency, X, Y, Z           : TODO: will have size = sum(nFields for each ElectrodeSource)
+    RANGE delay, duration, ramp_up, ramp_down
+    RANGE displacementX, displacementY, displacementZ
+    RANGE enabled  : set when pointer is assigned, this prevents mcomplex calculation to access unassigned e_ext reference
```
cattabiani: nitpick: which pointer? You have e_ext but also phase, frequency, X, Y, Z. If all, use plural

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,131 @@
...
+        _p_frequency = vector_new1(size);
+
+        auto *vX = *reinterpret_cast<IvocVect**>(&_p_X);
+        auto *vY = *reinterpret_cast<IvocVect**>(&_p_Y);
+        auto *vZ = *reinterpret_cast<IvocVect**>(&_p_Z);
+        auto *vphase = *reinterpret_cast<IvocVect**>(&_p_phase);
+        auto *vfreq = *reinterpret_cast<IvocVect**>(&_p_frequency);
+
+        for( int i=0; i<size; i++ ) {
+            vector_vec(vX)[i] = displacementX * vector_vec(argX)[i];
```
cattabiani, line 88: probably it is just me but I find passing part of the variables as input to the procedure and part as a sneaky assignment in `stimuli.py` very confusing. From the pov of the point process delay and displacementX are 2 range assigned variables. However, one is passed as input and the other assigned outside. Why is that? Is it standard?

neurodamus/core/stimuli.py
```diff
@@ -517,13 +516,8 @@ def __init__(self, base_amp, delay, duration, fields, ramp_up_time, ramp_down_ti
         self.dt = dt
         self.ramp_up_time = ramp_up_time
         self.ramp_down_time = ramp_down_time
-        # for delay, add cur_t as the first point, then advance cur_t
-        if delay > 0:
-            self.time_vec.append(self._cur_t)
-            self._cur_t = delay
-        self.efields = self.add_cosines()  # np.array, E_x, E_y, E_z vectors varied by time
         self.segment_displacements = {}  # {segment: displacement vectors in x/y/z w.r.t ground}
```
cattabiani, line 519: I know it may out of scope for this pr but:

We should consider computing segment displacements on the fly inside apply_all_stimuli rather than storing them in a dict on ElectrodeSource — this eliminates the intermediate state and explicit cleanup, letting Python's normal scoping handle the lifetime of the temporary displacement vectors, and makes the code easier to follow since computation and consumption happen in one place.

If you do not want to do this in this pr but you agree with my idea, may you open an issue?

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -0,0 +1,131 @@
...
+           ramp_factor = (t-delay) / ramp_up;
+       }
+       if( delay+ramp_up+duration < t && t < delay+duration+ramp_up+ramp_down ) {
+           ramp_factor = 1 - (t-(delay+ramp_up+duration)) / ramp_down;
+       }
+
+       size = vector_capacity(vX);
+       for( i=0; i<size; i++ ) {
+           double wavefactor = cos(2 * PI * vector_vec(vfreq)[i] / 1000 * (t-delay) + vector_vec(vphase)[i] );
+	   _lefield_accum += 1e3 * ramp_factor * (vector_vec(vX)[i] * wavefactor + vector_vec(vY)[i] * wavefactor + vector_vec(vZ)[i] * wavefactor);
```
cattabiani, line 130: aaah, doing it in this way is so much simpler... and we even save memory

tests/unit/test_single_extracellular_stimulus.py
```diff
@@ -164,32 +132,52 @@ def test_one_field_noramp(create_tmp_simulation_config_file):
...
+        0.4809515,
+        0.2972444,
+        0,
+        -0.2972444,
+        -0.4809515,
+    ]
+    t_beforebreakpoint = np.arange(dt / 2, duration, dt)
+
+    # check the references against the cosine function
+    npt.assert_allclose(f_cos(t_beforebreakpoint), ref_stimvec_mod[1:], atol=1e-9)
```
cattabiani, line 153: I do not get why we have this check. We are comparing a hardcoded reference (produced with the previous version of the code) with another hardcoded function that computes a cosine. You can check this once for your computation but then it will always be the same except if you need to change the function. And if you want to change the function you will need to change the reference. 

TLDR: apart from initial debugging, we do not need this check. Just use `f_cos` and assume it is correct. It is a oneliner

tests/unit/test_single_extracellular_stimulus.py
```diff
@@ -164,32 +132,52 @@ def test_one_field_noramp(create_tmp_simulation_config_file):
...
+    npt.assert_allclose(f_cos(t_vecplay), ref_stimvec_vecplay[:-1], atol=1e-6)
```
cattabiani, line 154: same

tests/unit/test_single_extracellular_stimulus.py
```diff
@@ -164,32 +132,52 @@ def test_one_field_noramp(create_tmp_simulation_config_file):
...
 
-    n.clear_model()
+    # check the actually segment.extracellular._ref_e against the current reference
+    npt.assert_allclose(rec_dend, ref_stimvec_mod, atol=1e-9)
```
cattabiani, line 157: same

tests/unit/test_single_extracellular_stimulus.py
```diff
@@ -164,32 +132,52 @@ def test_one_field_noramp(create_tmp_simulation_config_file):
...
+    npt.assert_allclose(rec_soma, np.zeros(len(rec_dend)))
 
 
 REF_COSINE = np.array(
```
cattabiani, line 161: I would prefer to have a function like f_cos that produces these values. It is much easier to reason on what are these numbers

tests/unit/test_single_extracellular_stimulus.py
```diff
@@ -219,43 +207,76 @@ def test_one_field_noramp(create_tmp_simulation_config_file):
...
+            1 - (t_vec - (ramp_up_time + duration)) / ramp_down_time,
+            envelope,
+        )
+        return envelope
+
+    ramping_mod = make_ramp_envelope(t_beforebreakpoint)
+    npt.assert_allclose(f_cos(t_beforebreakpoint) * ramping_mod, REF_COSINE[1:], atol=1e-6)
 
 
 REF_CONSTANT = np.array(
```
cattabiani, line 274: I would prefer a function..

cattabiani (commented):
I have only a handful of (minor) comments and proposals. After replying/addressing them, I will give my approval :D

neurodamus/data/mod/efield_integrator.mod
mgeplf: no, it's because the `extracellular` mechanism isn't supported.

neurodamus/data/mod/efield_integrator.mod
mgeplf: I imagine it's to handle arrays

neurodamus/data/mod/efield_integrator.mod
cattabiani: ah yes, true. I forgot the fundamentals lol

neurodamus/core/stimuli.py
WeinaJi: Yes, in another PR for multiple stimulus blocks.

tests/unit/test_single_extracellular_stimulus.py
WeinaJi: Mike asked to ensure the safety of the reference. And I prefer to have a clear reference vector rather than computing it on the fly with `f_cos` . It is just my taste of testing.

tests/unit/test_single_extracellular_stimulus.py
WeinaJi: Again, I prefer to have a clear reference vector and it can be reused for other tests.

neurodamus/data/mod/efield_integrator.mod
cattabiani: I am talking about displacementX, not the array. I marked it here because it is the first spot it is used

neurodamus/core/stimuli.py
cattabiani: open issue then :D

WeinaJi: It still belongs to.https://github.com/openbraininstitute/prod-circuit-simulation/issues/172

tests/unit/test_single_extracellular_stimulus.py
cattabiani: a reference vector is much harder to reason about. It is a cosine... ok. what phase, what frequency, what amplitude? It is not immediately evident. Doing an FFT to find out parameters that should be clear from the beginning is too much. 

If you really want you can have a unit test for the `f_cos`. However, I find it pretty useless given that the function is a oneliner and the reference is hardcoded. It works once, it works

In addition the array reference is much more constraining. What if I want to extend the test to more ms? I need to extend the reference. A function can just spit the correct numbers as far in the future as I want

@mgeplf the safety of the reference, once tested, it is done. What do you think about a unit test that checks `f_cos` and the use the function itself? Honestly, I do not even get why the previous implementation should be taken as reference. The function picks directly phase, amplitude etc. That, if correctly written, is analitically? mathematically? correct. Much more correct than some doubles used in the previous implementation

neurodamus/core/stimuli.py
cattabiani: in this case... open sub-issue in neurodamus and link it there in this case

tests/unit/test_single_extracellular_stimulus.py
mgeplf: > Mike asked to ensure the safety of the reference. 

Hmm, that was not my intent, I figured the "best" test was to use the initial version using vecplay on a cell, and then also do the new MOD version, and then compare the results.  To make them comparable, since the sampling times are different, I suggested using the `np.interp` and compare the pointwise values.

Sorry if that wasn't clear.

neurodamus/data/mod/efield_integrator.mod
jamesgking: So if I pass the displacements in via mod function as well, you would be satisfied?

tests/unit/test_single_extracellular_stimulus.py
WeinaJi: `np.interp` is not enough to interpolate because the points follow a cosine wave, I also tried `scipy.interpolate.CubicSpline` but there were still visible differences with the interpolation. So I changed to use `f_cos` to ensure the correctness of the reference.

neurodamus/data/mod/efield_integrator.mod
cattabiani: it is not about being satisfied. I want to understand why there is this distinction. Maybe the current approach is the standard and better way to do things

tests/unit/test_single_extracellular_stimulus.py
mgeplf: Huh, interesting, thanks for trying.

I guess it's sensitive to dt; what value did you try?

WeinaJi: @cattabiani, computing reference for each test requires to retrieve every field' variables. There are a few for a single field test, but more for 2-fields, 3-fields tests.  I don't think that's necessary for every test.

WeinaJi: > Huh, interesting, thanks for trying.
> 
> I guess it's sensitive to dt; what value did you try?

@mgeplf , it is dt=1 for this test, perhaps it is too large here.

mgeplf: Yeah, I think I mentioned that when I was playing in my original PoC; there was a "large" difference.
Since there would only be 2 segments being simulated (the vecplay and the mod versions), maybe it's worth trying with a shorter `dt`.  Doesn't have to be now though.

tests/unit/test_single_extracellular_stimulus.py
jamesgking: I will make an issue for future discussion on this

tests/unit/test_single_extracellular_stimulus.py
jamesgking: will make issue for future discussion

neurodamus/core/stimuli.py
jamesgking: I'll make the new issue for discussion

tests/unit/test_single_extracellular_stimulus.py
cattabiani: @WeinaJi reference in the sense of using f_cos? or the array? Because, yes, you need to pick up a few values for each test but then the test automatically updates if we do changes to the test itself

tests/unit/test_single_extracellular_stimulus.py
cattabiani: to me the "clear reference" is the function itself (i.e. f_cos). Anyway, I think we do not agree. I am satisfied with the issue

cattabiani (dismissed):

neurodamus/data/mod/efield_integrator.mod
jamesgking: done

neurodamus/data/mod/efield_integrator.mod
jamesgking: done

neurodamus/core/stimuli.py
jamesgking: https://github.com/openbraininstitute/neurodamus/issues/524

tests/unit/test_single_extracellular_stimulus.py
jamesgking: https://github.com/openbraininstitute/neurodamus/issues/523

cattabiani (approved):
LGTM

PR #525, opened 2026-05-20
Changelog for 4.2.2

jamesgking:


CHANGES.rst
```diff
@@ -2,6 +2,17 @@
...
+Version 4.2.2
+=============
+**20 May 2026**
+
+*Improvements*
+  * For Extracellular stimulus, switch from using vector.play to a PointProcess which computes the contribution during runtime, avoiding big memory allocation (#521)
+  * adapt to new libsonata (#517)
+
+*Bug Fixes*
+  * Update condition modification implementation to match the spec (#519)
```
WeinaJi, line 11: In my opinion, this is not a bug fix but rather an improvement

jamesgking: okay, will change

WeinaJi (approved):

CHANGES.rst
```diff
@@ -2,6 +2,15 @@
 Change Log
 ==========
 
+Version 4.2.2
```
cattabiani, line 5: sorry for noob question but I do not get why we have the change log file AND there is a change log in the release page of github. Isn't it a duplication of work?

cattabiani (approved):
LGTM

CHANGES.rst
jamesgking: I suppose the CHANGES.rst is useful if you somehow get the source without GitHub or any git metadata. I don't mind the duplication, but this is also my first time doing the release :)

WeinaJi: I've found this documentation more useful than the GitHub release notes because it separates the changes into 3 categories." After updating the file, it is clear what the major changes will be in the new release and how we could bump the version.

PR #528, opened 2026-05-21
Create reusable scripts to install neurodamus and its dependencies

mgeplf:
- ci/scripts/*.sh each script is one piece of the puzzle for installing neurodamus dependencies
- refactor `.github/workflows/simulation_test.yml` to use above scripts
- refactored Dockerfile to also use above scripts
- allows for other locations where neurodamus is installed to install things with the same versions and arguments
- add `sccache` for c/c++ building caches and `uv` for installing python dependencies


.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,94 @@
...
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          build-args: |
+            NEURODAMUS_COMMIT=${{ github.sha }}
+          push: false
+          load: true
+          tags: |
+            ${{ steps.meta.outputs.tags }}
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:4.2.2
```
heerener: Is this consciously hardcoded?

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,146 @@
...
+      if: startsWith(matrix.os, 'macOS')
+      run: |
+        brew install hdf5-mpi openmpi  # macOS hdf5-mpi requires openmpi
+
+    - name: Run tox baseline
+      run: |
+        tox -e baseline
+
+    - name: Upload baseline coverage to Coveralls
+      if: startsWith(matrix.python-version, '3.12')
```
heerener: There's a lot of steps which check for python 3.12 - can that be extracted to a variable for easier updates by future whoever?

ci/scripts/build-libsonata.sh
```diff
@@ -0,0 +1,39 @@
...
+#   PRE              called on function entry
+#   BUILD_DIR        Directory for cloning/building source (e.g. /tmp)
+#   INSTALL_DIR      Prefix where dependencies are installed (e.g. /opt/obi)
+#   CMAKE_BUILD_TYPE CMake build type (e.g. RelWithDebugInfo)
+#   PIP              pip command to use (e.g. "uv pip")
+
+build-libsonata() {
+    PRE || true
+
+    : "${PIP:?PIP is not set}"
```
heerener, line 14: What's the colon at the start of the line for? Isn't that just a no-op placeholder?

heerener (commented):
I've got some minor questions, but overall looks very well-organised.

Not going to comment on the actual software building part of it, I'll leave that to the experts.

.github/workflows/docker-image.yml
mgeplf: no, but I wanted to check if it worked; once it's good, I'll only push images on tag releases, with the correct tag.

ci/scripts/build-libsonata.sh
mgeplf: yeah, shell-ism for setting a default value w/o running something

.github/workflows/unit-tests.yml
mgeplf: `:s/3.12/3.14` no?

but sure

tox.ini
```diff
@@ -43,6 +43,7 @@ deps =
     neuron-nightly
     morphio
     -r tests/unit-mpi/requirements.txt
+passenv = OMPI_ALLOW_RUN_AS_ROOT,OMPI_ALLOW_RUN_AS_ROOT_CONFIRM
```
cattabiani, line 46: there are reasons why it is usually not allowed. Do we really need this?

tests/unit-mpi/requirements.txt
```diff
@@ -1,4 +1,2 @@
 pytest-isolate-mpi
-
-# force building mpi4py so that it picks up openmpi or mpich, whichever is installed
-mpi4py --no-binary=mpi4py
+mpi4py
```
cattabiani, line 2: are you sure you do not need to rebuild mpi4py and you can just get the wheel?

.github/workflows/simulation_test.yml
```diff
@@ -1,297 +1,171 @@
...
+        && build-libsonatareport $LIBSONATAREPORT_COMMIT
 
-    - name: Install mpi4py
-      if: steps.cache-venv.outputs.cache-hit != 'true'
+    - name: Install libsonata
       run: |
-        MPICC=mpicc pip install --no-binary=mpi4py mpi4py
+        source $BASE_DIR/ci/scripts/build-libsonata.sh \
+        && source venv/bin/activate \
+        && PIP='uv pip' build-libsonata $LIBSONATA_COMMIT
```
cattabiani, line 114: can't you write these as 3 different commands? I thought run could accept more than 1 line

.github/workflows/simulation_test.yml
```diff
@@ -1,297 +1,171 @@
...
-          export HDF5_INCLUDEDIR=/usr/include/hdf5/openmpi
-          export HDF5_LIBDIR=/usr/lib/x86_64-linux-gnu/hdf5/openmpi
-        fi
-        # Upgrade setuptools again, to avoid build error from `project.license`
-        # Neuron installed older setuptools but it doesn't work for h5py
-        python -m pip install --upgrade pip setuptools
-        CC="mpicc" HDF5_MPI="ON" HDF5_INCLUDEDIR=$HDF5_INCLUDEDIR HDF5_LIBDIR=$HDF5_LIBDIR \
-          pip install --no-cache-dir --no-binary=h5py h5py --no-build-isolation
+        source $BASE_DIR/ci/scripts/build-neuron.sh \
+        && source venv/bin/activate \
```
cattabiani, line 119: since we always activate the venv we could place the path in the variables or activate it automatically... there was a way

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,93 @@
...
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+              type=ref,event=tag
+
+      - uses: docker/setup-buildx-action@v4
+        with:
+          driver: docker-container
+
+      - name: Build Docker image
```
WeinaJi, line 46: Which platform is the image built for? Does it build for both `linux/amd64` and `linux/arm64` ?

ci/build-dependencies.sh
```diff
@@ -0,0 +1,473 @@
+#!/bin/bash
+#set -euxo pipefail
```
cattabiani: this is commented. is it intended?

ci/build-dependencies.sh
```diff
@@ -0,0 +1,473 @@
...
+
+    local NEOCORTEX_MOD_MULTI=$BASE/neurodamus-models/
+    local NEOCORTEX_MOD_MULTI_BUILD=$NEOCORTEX_MOD/build
+    local NEOCORTEX_MOD_MULTI_INSTALL=$NEOCORTEX_MOD_BUILD/install
+
+    if [[ ! -e $NEOCORTEX_MOD_MULTI ]]; then
+        git clone --branch="$branch" --depth=1 https://github.com/openbraininstitute/neurodamus-models.git $NEOCORTEX_MOD_MULTI
+    fi
+
+    DATADIR=$(PYTHON -c "import neurodamus; from pathlib import Path; print(Path(neurodamus.__file__).parent / 'data')")
```
cattabiani: are you missing a $ before PYTHON?

ci/build-dependencies.sh
```diff
@@ -0,0 +1,473 @@
...
+
+build-neocortex-multiscale-models() {
+    local branch=$1
+
+    echo "build neurodamus-neocortex-multiscale model"
+
+    source $ENV_FILE
+
+    local NEOCORTEX_MOD_MULTI=$BASE/neurodamus-models/
+    local NEOCORTEX_MOD_MULTI_BUILD=$NEOCORTEX_MOD/build
```
cattabiani: `NEOCORTEX_MOD` and `NEOCORTEX_MOD_BUILD` are used but not defined. Is it correct? Is there a test that checks this?

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,93 @@
+name: Build and Push Docker Image
```
cattabiani, line 1: triggers on every push to every branch... isn't it a little too much? Is this intended?

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,93 @@
...
+        id: push
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          build-args: |
+            NEURODAMUS_COMMIT=${{ github.sha }}
+          push: true
+          tags: |
+            ${{ steps.meta.outputs.tags }}
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:4.2.2
```
cattabiani: you are hardcoding 4.2.2. Shouldn't this be dynamic?

.github/workflows/simulation_test.yml
```diff
@@ -1,297 +1,171 @@
...
-      - 'pyproject.toml'
-      - 'tox.ini'
-      - 'requirements.txt'
+      - 'neurodamus/neurodamus/**'
+      - 'neurodamus/tests/**'
+      - 'neurodamus/.github/workflows/**'
+      - 'neurodamus/ci/**'
+      - 'neurodamus/pyproject.toml'
+      - 'neurodamus/tox.ini'
+      - 'neurodamus/requirements.txt'
```
cattabiani: are you sure you need all these neurodamus? Before we did not need them and we start from the same place. Just double checking that the ci is actually not silently skipping everything

ci/scripts/install-hdf5.sh
```diff
@@ -0,0 +1,59 @@
...
+    local HDF5_BUILD=$HDF5/build
+
+    [[ -e $HDF5 ]] || \
+      git clone --branch="$branch" --shallow-submodules --depth=1 https://github.com/HDFGroup/hdf5/ $HDF5
+
+    # have to use `Release`; RelWithDebugInfo causes `h5pcc` wrapper to pass -lhdf5_debug eventhough
+    # it is not included
+    local CMAKE_ARGS=(
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_C_COMPILER=`type mpicc` \
```
cattabiani, line 24: isn't it better: `$(which mpicc)` ?

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,150 @@
...
+      - 'ci/**'
+      - 'pyproject.toml'
+      - 'tox.ini'
+      - 'requirements.txt'
+  push:
+    branches:
+      - main
+
+env:
+    COVERATE_PYTHON=: '3.12'
```
cattabiani:
```suggestion
    COVERAGE_PYTHON=: '3.12'
```

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,150 @@
...
+        flag-name: baseline
+        parallel: true
+        fail-on-error: false
+
+    - name: Run tox unit
+      run: |
+        tox -e unit
+
+    - name: Upload unit coverage to Coveralls
+      if: startsWith(matrix.python-version, env.COVERATE_PYTHON)
```
cattabiani:
```suggestion
      if: startsWith(matrix.python-version, env.COVERAGE_PYTHON)
```

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,150 @@
...
+        flag-name: unit-ngv-mpi
+        parallel: true
+        fail-on-error: false
+
+    - name: Run tox unit-allen-v1
+      run: |
+        tox -e unit-allen-v1
+
+    - name: Upload unit-allen-v1 coverage to Coveralls
+      if: startsWith(matrix.python-version, env.COVERATE_PYTHON)
```
cattabiani:
```suggestion
      if: startsWith(matrix.python-version, env.COVERAGE_PYTHON)
```

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,150 @@
...
+      uses: coverallsapp/github-action@v2
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        file: .coverage.unit-allen-v1
+        flag-name: unit-allen-v1
+        parallel: true
+        fail-on-error: false
+
+    - name: Finalize Coveralls parallel build
+      if: startsWith(matrix.python-version, env.COVERATE_PYTHON)
```
cattabiani:
```suggestion
      if: startsWith(matrix.python-version, env.COVERAGE_PYTHON)
```

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,150 @@
...
+      if: startsWith(matrix.os, 'macOS')
+      run: |
+        brew install hdf5-mpi openmpi  # macOS hdf5-mpi requires openmpi
+
+    - name: Run tox baseline
+      run: |
+        tox -e baseline
+
+    - name: Upload baseline coverage to Coveralls
+      if: startsWith(matrix.python-version, env.COVERATE_PYTHON)
```
cattabiani:
```suggestion
      if: startsWith(matrix.python-version, env.COVERAGE_PYTHON)
```

.github/workflows/unit-tests.yml
cattabiani: I think this is a typo that proliferated. let me know if I was wrong. I added a bunch of suggestions

ci/build-dependencies.sh
```diff
@@ -0,0 +1,473 @@
+#!/bin/bash
```
cattabiani: The ci/Dockerfile sources build-dependencies.sh, and the root Dockerfile + simulation_test.yml use the new modular scripts.

It is not DRY and a it is fragile. May we use just the new way?

ci/build-dependencies.sh
```diff
@@ -0,0 +1,473 @@
...
+    pushd $BASE/neurodamus/
+    # launch simulation with NEURON
+    pushd tests/simulations/usecase3/
+    mpirun --allow-run-as-root --use-hwthread-cpus -np 2 special -mpi -python $NEURODAMUS_PYTHON/init.py --configFile=simulation_sonata.json --output-path=`pwd`
+    ls reporting/*.h5
+
+    ## launch simulation with CORENEURON
+    #mpirun --allow-run-as-root --use-hwthread-cpus -np 2 special -mpi -python $NEURODAMUS_PYTHON/init.py --configFile=simulation_sonata_coreneuron.json
+    #ls reporting_coreneuron/*.h5
+    #popd
```
cattabiani: nitpick:

there are pushds without popds at the end. It is fine for now but maybe slightly error prone. Feel free to discard this comment if it is fine as is

cattabiani (commented):
A few comments. In addition, you have shebang mixing:

`#!/bin/bash` and `#!/usr/bin/env`. Mostly cosmetic but maybe you want to standardize

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,93 @@
...
+          -e OMPI_ALLOW_RUN_AS_ROOT=1
+          -e OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
+          ${REGISTRY}/${IMAGE_NAME}:test
+          /bin/bash -c 'source /opt/obi/env-neocortex.sh &&
+                        uv pip install pytest pytest-forked pandas &&
+                        cd /tmp/neurodamus &&
+                        pytest -s -x --forked --durations=5 --durations-min=15 tests/integration-e2e &&
+                        pytest -s -x --forked --durations=5 --durations-min=15 tests/scientific'
+
+      - name: Push docker image
```
WeinaJi, line 74: I guess we will only push the image at tagging? There will be an `if` condition, right?

.github/workflows/publish_wheels.yml
```diff
@@ -1,57 +1,46 @@
...
-          cd $GITHUB_WORKSPACE
-          pip3 install tox
-          DIST_FILE=`ls dist/*whl` && tox --installpkg=$DIST_FILE -e ${{ matrix.python_version }}
+      - uses: actions/checkout@v4
+
+      - name: Build SDist
+        run: pipx run build --sdist
+
+      - name: Check metadata
+        run: pipx run twine check dist/*
```
WeinaJi: We don't build wheels any more?

.github/workflows/simulation_test.yml
```diff
@@ -1,297 +1,171 @@
...
       - main
     paths:
-      - 'neurodamus/**'
-      - 'tests/**'
-      - '.github/workflows/**'
-      - 'ci/**'
-      - 'pyproject.toml'
-      - 'tox.ini'
-      - 'requirements.txt'
+      - 'neurodamus/neurodamus/**'
```
WeinaJi: Isn't it the relative path to the repo root ?

.github/workflows/simulation_test.yml
```diff
@@ -1,297 +1,171 @@
...
-        required: false
-      NEURODAMUS_MODELS_BRANCH:
-        description: 'neurodamus-models branch to use'
-        required: false
 
 env:
-  NEURON_COMMIT_ID: '9.0.1'
+  LIBSONATAREPORT_COMMIT: 2.0.0
+  LIBSONATA_COMMIT: v0.1.35
+  NEURODAMUS_MODELS_COMMIT: 1a3b8f98fbabefb34f255c1fe63f7d7a4422223f
```
WeinaJi: I just created a new release for neurodamus-models 2.4.4 before I start to move away the common folder. We can use this tag I think.

.github/workflows/unit-tests.yml
mgeplf: done

tox.ini
mgeplf: it's a passenv, so someone needs to set it in their own shell.
For tests, I'm failing to see what the reasons for not letting the user pick is - do you have any in mind?

tests/unit-mpi/requirements.txt
mgeplf: It's worked thus far for the ubuntu/debian system packages, the azure custom install they build, and the AWS custom install.

If it fails in the future, I guess we can revisit it, but it seems that it targets the generic MPI ABI, which is why it's worked thus far.

.github/workflows/simulation_test.yml
mgeplf: It does; I will change it.

mgeplf: Although, it is nice to be able to cut and paste it from the Dockerfile, and have it be the same.  Let me know if you have a strong preference.

.github/workflows/simulation_test.yml
mgeplf: Cross step, the environment isn't preserved.  There is the `echo [...] >> $GITHUB_ENV` idiom, but I prefer to be explicit.

ci/build-dependencies.sh
mgeplf: sigh, I'm so sorry, that wasn't supposed to be commited ;(
Sorry for making you review that.

ci/build-dependencies.sh
mgeplf: not supposed to be part of the PR :(

ci/build-dependencies.sh
mgeplf: file not supposed to be part of the PR :(

.github/workflows/docker-image.yml
mgeplf: it's for testing, once this goes in I will only have it run on tag.

Thanks for the reminder.

.github/workflows/docker-image.yml
mgeplf: fixed.

.github/workflows/unit-tests.yml
mgeplf: lol; I'll search and replace

.github/workflows/simulation_test.yml
mgeplf: Good point!
I had changed the checkout dir to have fewer conflicts, but decided that wasn't the right thing.

ci/scripts/install-hdf5.sh
mgeplf: `which` isn't supported everywhere, so I switched it to `type`

ci/build-dependencies.sh
mgeplf: the ci/Dockerfile is going away; the root one is the one that works.

ci/build-dependencies.sh
mgeplf: (sorry again, file wasn't supposed to be part of the PR)

.github/workflows/simulation_test.yml
```diff
@@ -17,281 +14,158 @@ on:
...
-    - name: baseline
-      run: |
-        tox -e baseline
-      
     - name: integration-e2e
       run: |
-        tox -e integration-e2e
+        export NEURON_INIT_MPI=1 \
+        && source $INSTALL_DIR/env-neocortex.sh \
+        && pytest -s -x --forked --durations=5 --durations-min=15 $BASE_DIR/tests/integration-e2e
```
WeinaJi, line 165: Why do you not use`tox` ? It sets the env var for each test env properly.

ci/scripts/build-libsonata.sh
```diff
@@ -0,0 +1,39 @@
...
+
+    local COMMIT=${1:-HEAD}
+
+    local LIBSONATA=$BUILD_DIR/libsonata
+
+    [[ -e $LIBSONATA ]] || git clone --filter=blob:none \
```
WeinaJi, line 20: Is it needed to check out the repo? We usually just pip install directly from github, as mentioned by https://github.com/openbraininstitute/libsonata/blob/master/README.rst

.github/workflows/simulation_test.yml
```diff
@@ -17,281 +14,158 @@ on:
...
+        && SUDO=sudo install-sccache
 
-    - name: Install libsonata from source
-      if: steps.cache-venv.outputs.cache-hit != 'true'
+    - name: Upgrade pip and install base python packages
       run: |
-        CC=mpicc CXX=mpic++ pip install git+https://github.com/openbraininstitute/libsonata@${{ env.LIBSONATA_BRANCH }}
+        uv venv --clear --python ${{ matrix.python-version }} venv \
+        && source venv/bin/activate \
+        && uv pip install --upgrade pip cmake ninja pytest pytest-coverage mpi4py \
```
WeinaJi, line 95: ninja?  Is it installed by apt-get ?

ci/scripts/build-neuron.sh
```diff
@@ -0,0 +1,67 @@
...
+
+    ( cd $NRN && \
+        git fetch --depth 1 origin $COMMIT &&
+        git checkout FETCH_HEAD &&
+        git submodule update --init --depth=1 --recursive \
+            external/spdlog \
+            external/Random123 \
+            external/fmt \
+            external/nanobind \
+    )
```
WeinaJi, line 36: Usually NEURON CMakeFile should check out the required submodule automatically. Does it work with pip install ?

.github/workflows/simulation_test.yml
```diff
@@ -17,281 +14,158 @@ on:
...
-      env:
-        cache-name: cache-venv
+    - name: sccache build cache
+      id: cache-sccache
+      uses: actions/cache@v5
       with:
-        path: venv
-        key: ${{ matrix.os }}-openmpi-libsonata-${{ env.LIBSONATA_BRANCH }}-py${{ steps.pyver.outputs.version }}
+        path: ${{ github.workspace }}/cache
+        key: sccache-${{ runner.os }}-${{ github.run_id }}
```
WeinaJi, line 82: I don't know about the sccache. Why is there no separate caches per python version?

WeinaJi (commented):
Is this PR related to any ticket? 
Since there are so many changes, it would be better to track them under a planned task.

.github/workflows/docker-image.yml
mgeplf: currently only amd64, since that's all we support, amd64 could come after, once if we have a use case.

.github/workflows/docker-image.yml
mgeplf: yeah

.github/workflows/publish_wheels.yml
mgeplf: for a source only package, I'm not sure there's a lot of value to build the wheel, but I can add it back if you think it's important.

Interestingly, the wheel is larger than our sdist, I should probably look why.

.github/workflows/simulation_test.yml
mgeplf: ok, I'll change it. thanks.

.github/workflows/simulation_test.yml
mgeplf: It's the same as ccache, just easier to install :)
I highly recommend it if you can add it to your c/c++ workflow.

I'm hoping to reuse the builds when possible from the different versions; it should handle differences in source files if there are any, from what I understand.

.github/workflows/simulation_test.yml
mgeplf: i install the python packaged version, since it will be the newest one.

.github/workflows/simulation_test.yml
mgeplf: for the integration tests, it will build its own virtualenv; this way I'm actually testing the one that was created.

ci/scripts/build-libsonata.sh
mgeplf: this way I can cache the repo, plus all repos are handled the same way when there is a choice of commit
I can change it if you want

ci/scripts/build-neuron.sh
mgeplf: it checks out more submodules than are needed for compilaion, so this is faster - don't need `Catch2` or `rxdtestdata` and such for testing, for example

mgeplf (commented):
@WeinaJi > Is this PR related to any ticket?

Good point, it's https://github.com/openbraininstitute/prod-circuit-simulation/issues/128

tox.ini
cattabiani: it is just that it is not the standard. I do not think it is really important in this case

.github/workflows/simulation_test.yml
cattabiani: nah. it is ok also in this way. It is just harder to read (IMHO)

.github/workflows/simulation_test.yml
cattabiani: yes, I meant `echo [...] >> $GITHUB_ENV`. Ok, personal choice

ci/scripts/install-hdf5.sh
cattabiani: O.O You learn always something

cattabiani (dismissed):
LGTM (for me). I would let @WeinaJi finish her review

.github/workflows/docker-image.yml
WeinaJi: Is the image going to be public? I think it would be good to support both platforms to share with users. With my MacOS, I can't `docker pull` this amd64 image.

mgeplf: They are public by default.

How have you installed docker on mac?  x86 should be supported through a translation layer.
With docker desktop, `docker run --platform linux/amd64` should work.

If you installed docker with `brew`, I believe that Gianlucca uses [colima](https://github.com/abiosoft/colima), so something like:
`colima start --vm-type vz --vz-rosetta --mount-type virtiofs --cpu 4 --memory 8`
gives good performance (by using rosetta).

.github/workflows/simulation_test.yml
WeinaJi: The tox [testenv:integration-e2e] uses `passenv = *`. In fact, the tox venv only installs the pytest relevant packages. It will still uses the packages that are installed by `env-neocortex.sh`. I guess they are equivalent.

ci/scripts/build-neuron.sh
WeinaJi: Oh, I didn't know that. It was recommended by NEURON devs before. Maybe it has changed.

ci/scripts/install-aws-deps.sh
WeinaJi, line 1: Is this script needed for CI ? The same question for `install-dnf-dependencies.sh` . Is it better to store them in another folder like `aws/scripts`?

mgeplf: good point, I'm of two minds; the `dnf` scripts are useful if one is installing on a redhat based distribution, and instead of "everyone" needing to copy those around, they might be useful here.

The AWS ones, I will move to the aws image building.

.github/workflows/simulation_test.yml
mgeplf: I'm a bit puzzled, my understanding is that `passenv = *` only makes the environment variables available, but doesn't impact how tox creates its "isolated" virtualenv.

It's true that `PYTHONPATH` is being explicitly set, but that's only giving access to NEURON.

I reran the old CI, with no caches, and I get:
<details>
<pre>
Run tox -e integration-e2e
  tox -e integration-e2e
  shell: /usr/bin/bash -e {0}
  env:
    NEURON_COMMIT_ID: 9.0.1
    RDMAV_FORK_SAFE: 1
    pythonLocation: /opt/hostedtoolcache/Python/3.10.20/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.10.20/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.20/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.20/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.10.20/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.10.20/x64/lib
    NEURON_BRANCH: master
    LIBSONATA_BRANCH: v0.1.36
    LIBSONATA_REPORT_BRANCH: 2.0.0
    NEURODAMUS_MODELS_BRANCH: 2.4.4
    NEURODAMUS_NEOCORTEX_ROOT: /home/runner/work/neurodamus/neurodamus/neurodamus-models/install
    PYTHONPATH: /home/runner/work/neurodamus/neurodamus/nrn/build/install/lib/python:
    HOC_LIBRARY_PATH: /home/runner/work/neurodamus/neurodamus/neurodamus-models/install/share/neurodamus_neocortex/hoc
    CORENEURONLIB: /home/runner/work/neurodamus/neurodamus/neurodamus-models/install/lib/libcorenrnmech.so
    NRNMECH_LIB_PATH: /home/runner/work/neurodamus/neurodamus/neurodamus-models/install/lib/libnrnmech.so
    NEURODAMUS_PYTHON: /home/runner/work/neurodamus/neurodamus/neurodamus/data
integration-e2e: install_deps> python -I -m pip install 'coverage>=7.10.6' 'pytest-cov>=7.0' -r tests/requirements.txt
.pkg: _optional_hooks> python /home/runner/work/neurodamus/neurodamus/venv/lib/python3.10/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: get_requires_for_build_sdist> python /home/runner/work/neurodamus/neurodamus/venv/lib/python3.10/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: freeze> python -m pip freeze --all
.pkg: hatch-vcs==0.5.0,hatchling==1.29.0,packaging==26.2,pathspec==1.1.1,pip==26.1.1,pluggy==1.6.0,setuptools==82.0.1,setuptools-scm==10.0.5,tomli==2.4.1,trove-classifiers==2026.5.22.10,typing_extensions==4.15.0,vcs-versioning==1.1.1
.pkg: build_sdist> python /home/runner/work/neurodamus/neurodamus/venv/lib/python3.10/site-packages/pyproject_api/_backend.py True hatchling.build
integration-e2e: install_package_deps> python -I -m pip install docopt-ng h5py 'libsonata>=0.1.35' packaging psutil scipy
integration-e2e: install_package> python -I -m pip install --force-reinstall --no-deps /home/runner/work/neurodamus/neurodamus/.tox/.tmp/package/5/neurodamus-4.2.3.dev1.tar.gz
integration-e2e: freeze> python -m pip freeze --all
integration-e2e: contourpy==1.3.2,coverage==7.14.1,cycler==0.12.1,docopt-ng==0.9.0,exceptiongroup==1.3.1,fonttools==4.63.0,h5py==3.16.0,iniconfig==2.3.0,kiwisolver==1.5.0,libsonata==0.1.36,matplotlib==3.10.9,neurodamus @ file:///home/runner/work/neurodamus/neurodamus/.tox/.tmp/package/5/neurodamus-4.2.3.dev1.tar.gz#sha256=e3c748d3231bd7ec54222e40b976dbe629756c52903d18357e0d0a52c08703ec,numpy==2.2.6,packaging==26.2,pandas==2.3.3,pillow==12.2.0,pip==26.1.1,pluggy==1.6.0,psutil==7.2.2,py==1.11.0,pyparsing==3.3.2,pytest==7.4.0,pytest-cov==7.1.0,pytest-forked==1.6.0,python-dateutil==2.9.0.post0,pytz==2026.2,scipy==1.15.3,setuptools==82.0.1,six==1.17.0,tomli==2.4.1,typing_extensions==4.15.0,tzdata==2026.2
integration-e2e: commands[0]> python -c 'import os; print(os.environ.get('"'"'HOC_LIBRARY_PATH'"'"', '"'"''"'"'))'
/home/runner/work/neurodamus/neurodamus/neurodamus/data/hoc:/home/runner/work/neurodamus/neurodamus/neurodamus-models/install/share/neurodamus_neocortex/hoc
integration-e2e: commands[1]> pytest -s -x --forked --durations=5 --durations-min=15 tests/integration-e2e
</pre>
</details>
(from https://github.com/openbraininstitute/neurodamus/actions/runs/26444860829/job/78800854380#step:27:1)

This has the dump of the python packages installed: 

```
integration-e2e: contourpy==1.3.2,coverage==7.14.1,cycler==0.12.1,docopt-ng==0.9.0,exceptiongroup==1.3.1,fonttools==4.63.0,h5py==3.16.0,iniconfig==2.3.0,kiwisolver==1.5.0,libsonata==0.1.36,matplotlib==3.10.9,neurodamus @ file:///home/runner/work/neurodamus/neurodamus/.tox/.tmp/package/5/neurodamus-4.2.3.dev1.tar.gz#sha256=e3c748d3231bd7ec54222e40b976dbe629756c52903d18357e0d0a52c08703ec,numpy==2.2.6,packaging==26.2,pandas==2.3.3,pillow==12.2.0,pip==26.1.1,pluggy==1.6.0,psutil==7.2.2,py==1.11.0,pyparsing==3.3.2,pytest==7.4.0,pytest-cov==7.1.0,pytest-forked==1.6.0,python-dateutil==2.9.0.post0,pytz==2026.2,scipy==1.15.3,setuptools==82.0.1,six==1.17.0,tomli==2.4.1,typing_extensions==4.15.0,tzdata==2026.2
```

But those all seem to be ones that were installed a couple lines before, and don't include things like `mpi4py` that would exist from the previous installation steps.

So, I think, the previous version isn't equivalent.  Thoughts @WeinaJi?

.github/workflows/simulation_test.yml
```diff
@@ -17,281 +14,158 @@ on:
...
-        required: false
-      NEURODAMUS_MODELS_BRANCH:
-        description: 'neurodamus-models branch to use'
-        required: false
 
 env:
-  NEURON_COMMIT_ID: '9.0.1'
+  LIBSONATAREPORT_COMMIT: 2.0.0
+  LIBSONATA_COMMIT: v0.1.35
+  NEURODAMUS_MODELS_COMMIT: 2.4.4
```
WeinaJi, line 24: Different from NEURON, we have fully control over these 3 repos. I don't think that we will test against a particular commit in the main branch for them. It is either `main` or a `release`. Perhaps a branch variable is already good.

mgeplf: Putting any commit that exists in the repository should work here, including the name of branch (at which point it will "chose" `HEAD`, but we wouldn't really want that.
Since that is possible, I don't think we need to have the option to specify a branch.

ci/scripts/install-python-dependencies.sh
```diff
@@ -0,0 +1,14 @@
...
+#
+# Environment variables:
+#   PRE              called on function entry
+#   PIP pip command to use (e.g. "uv pip")
+
+install-python-dependencies() {
+    PRE || true
+
+    $PIP install --upgrade pip setuptools
+    $PIP install -v cython numpy wheel pkgconfig jinja2 pyyaml cmake ninja morphio
```
jplanasc: Is there a reason why `morphio` is here? Shouldn't this be installed through the specific package dependencies?

.github/workflows/simulation_test.yml
```diff
@@ -1,8 +1,5 @@
-name: neurodamus build and test
+name: neurodamus integration and simulation tests
 on:
-  schedule:
-    # every evening at 20:00 UTC
-    - cron: '0 20 * * *'
```
jplanasc, line 5: Daily tests sounds excessive, but maybe once per week or bi-monthly would be useful, to catch inter-package compatibility issues?

Just an idea, but up to you.

ci/scripts/install-python-dependencies.sh
mgeplf: I forgot there was a `full` option to neurodamus; I'll switch it to there - thanks.

.github/workflows/simulation_test.yml
mgeplf: Agreed, that's what it was before, but I'd love to reduce it.

Dockerfile
```diff
@@ -0,0 +1,92 @@
+ARG UV_VERSION=0.11.15
+ARG PYTHON_VERSION=3.12
+
+FROM ghcr.io/astral-sh/uv:${UV_VERSION} AS uv
+FROM python:${PYTHON_VERSION}-slim
+
+ENV SCCACHE_DIR=/var/cache/sccache
+
+ARG LIBSONATAREPORT_COMMIT=de0abd0e73f29975ec7caeb80118bd617f2dbe0c
```
jplanasc: I thought we were trying to use tags/versions whenever possible, is this intended?

`.github/workflows/simulation_test.yml` is using tag `2.0.0`

mgeplf: Yup, fixed.

Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+ARG LIBSONATA_COMMIT=v0.1.35
+ARG NEURODAMUS_COMMIT=4.2.1
+ARG NEURODAMUS_MODELS_COMMIT=2.4.4
+ARG NEURON_COMMIT=9.0.1
+
+ENV USER_VENV_NAME=user_venv
+
+ENV INSTALL_DIR=/opt/obi
+ENV BUILD_DIR=/tmp
```
jplanasc, line 18: I guess this should be managed by the step calling this script, but just to be taken into account that `/tmp` partition may be quite limited in size, especially on certain cloud instances.

ci/scripts/build-libsonata.sh
WeinaJi: ok, that's fine for me.

.github/workflows/simulation_test.yml
WeinaJi: It is ok for me to name them as COMMIT in CI tests.

.github/workflows/simulation_test.yml
WeinaJi: How about weekly? Every Sunday morning ? '0 20 * * 0'

WeinaJi: oh, I was wrong, it was for 20h, then every saturday evening. '0 20 * * 6'

jplanasc: Are you sure you want to receive failure emails during the weekend? ;)

.github/workflows/simulation_test.yml
cattabiani: just as reference, if the problem is the name chatGpt suggests: "REVISION" to indicate all: commits, tags etc

Dockerfile
mgeplf: This is a dockerfile, so it's being built "locally", and doesn't impact the cloud system at all.

.github/workflows/simulation_test.yml
WeinaJi: I did receive some for macOS CI :) And the failure disappeared during the week. I suspect it was a different macOS runner during thew weekend.

mgeplf: changed to: `'0 20 * * 6'`

.github/workflows/docker-image.yml
WeinaJi: Now I am able to pull image `docker pull  --platform linux/amd64 ghcr.io/openbraininstitute/neurodamus:test`.
BTW, are you planning to change the image tag name to the actual neurodamus tag ?

mgeplf: > Now I am able to pull image docker pull  --platform linux/amd64 ghcr.io/openbraininstitute/neurodamus:test.

Cool, I have no idea why the GH interface copies a useless way.

It running w/ rosetta?

> BTW, are you planning to change the image tag name to the actual neurodamus tag?

Yes, that's what the `steps.meta.outputs.tags` should do, but I won't know for sure that works until it gets run with a release.

.github/workflows/simulation_test.yml
mgeplf: fixed

ci/scripts/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+This stack includes MPI, parallel HDF5, the python dependencies, libsonatareport and neuron.
+
+The goal is to make creating this environment reproducible to a degree and uniform across different installations.
+
+These scripts can be used as building blocks to install the stack or pieces of the stack in different environments and different configurations.
+
+The best way to understand this is to look both at the [Github Actions](https://github.com/openbraininstitute/neurodamus/blob/main/.github/workflows/simulation_test.yml) and also the [Docker build](https://github.com/openbraininstitute/neurodamus/blob/main/Dockerfile) to see how they can be used.
+
+Each of the scripts follows the same pattern comma it is sourced and then one of the functions within it is executed.
```
jplanasc: > the same pattern comma it is sourced

Was there meant to be a `,` or a `:` instead of `comma`?

ci/scripts/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+This directory contains scripts to create an environment that is useable by neurodamus.
+
+This stack includes MPI, parallel HDF5, the python dependencies, libsonatareport and neuron.
+
+The goal is to make creating this environment reproducible to a degree and uniform across different installations.
+
+These scripts can be used as building blocks to install the stack or pieces of the stack in different environments and different configurations.
+
+The best way to understand this is to look both at the [Github Actions](https://github.com/openbraininstitute/neurodamus/blob/main/.github/workflows/simulation_test.yml) and also the [Docker build](https://github.com/openbraininstitute/neurodamus/blob/main/Dockerfile) to see how they can be used.
```
jplanasc, line 11: Personal view: I would make this file self-contained instead of asking the reader to go look some other files. I think it will make it clearer.

ci/scripts/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Each of the scripts follows the same pattern comma it is sourced and then one of the functions within it is executed.
+
+It is assumed that "bash" is the shell that is being used.
+
+The order of operations is important because there are some dependencies between them, consult the GitHub actions to see what the latest suggested order is.
```
jplanasc, line 17: Same as above, even if it's a slight duplication, I'd make this file self-contained.

ci/scripts/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Not every step is required to be execute, for instance `hdf5` might be installable by the system package manager so it does not need to be compiled.
```
jplanasc: Typo: executed?

ci/scripts/README.md
mgeplf: fixed

ci/scripts/README.md
mgeplf: I disagree; it's too easy to forget to update documentation, but the simulation test will always be in the right order.
I've included a link, so it's easy to jump too.

ci/scripts/README.md
mgeplf: fixed

Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+ENV SCCACHE_DIR=/var/cache/sccache
+
+ARG LIBSONATAREPORT_COMMIT=2.0.0
+ARG LIBSONATA_COMMIT=v0.1.35
+ARG NEURODAMUS_COMMIT=4.2.1
+ARG NEURODAMUS_MODELS_COMMIT=2.4.4
+ARG NEURON_COMMIT=9.0.1
+
+ENV USER_VENV_NAME=user_venv
```
WeinaJi: Can we use an absolute path for this env var ? If the user change the directory in the container, the absolute path will be useful to source the virtual env.

And is `Dockerfile` required to be in the parent folder? Because we also have a README file in the `docker/` folder to guide external users. I think it is better to keep the `Dockerfile` in that folder and update the docker README.

mgeplf: > Can we use an absolute path for this env var ? 

Sure.

> And is Dockerfile required to be in the parent folder?

The file has to be in a place where all its dependencies are relatively children; since I use `ci/scripts/*`, it needs to be in a parent.

WeinaJi: Ok, then I think we can remove the entire `docker/` folder and just update the docker section in the neurodamus README.rst.

WeinaJi: Shall we copy the script `https://github.com/openbraininstitute/neurodamus/blob/main/docker/build_neurodamus.sh` into the docker image ? It links `nrnivmodl` with the relevant libs for libsonatareporting.
Otherwise it will be hard for external users to manually link them.

mgeplf: Sure, I will copy it in and update it.

ci/scripts/README.md
jplanasc: I still think that a short summary paragraph wouldn't hurt. I don't think redirecting the user to 2 source files (without further explanation) is a good practice. 

If it's too easy to forget updating documentation, why are we writing it at all? ;)

README.rst
```diff
@@ -95,22 +95,25 @@ Docker container
...
+Docker container
+----------------
+
+With the docker image, you can start a neurodamus container with an interactive Bash shell and meanwhile mount your local folder which contains your mod files and the circuit data.
+
+.. code-block:: sh
+
+  docker run --rm -it -v <full path of your folder_mods_circuit>:/mnt/mydata ghcr.io/openbraininstitute/neurodamus:$TAG /bin/bash
+
+To build new models, one can use the script in `/opt/obi/make-neurodamus-models.sh`.
```
WeinaJi:
```suggestion
To build new models, one can use the script in `/opt/obi/build-neurodamus-models.sh`.
```

README.rst
```diff
@@ -95,22 +95,25 @@ Docker container
...
+One can also examine the `/opt/obi/env-neocortex.sh` file to see how neocortext environment variables are setup so that the models are loaded.
+NEURON is installed, so by activating the virtual environment with `source $USER_ENV/bin/activate` one gets access to `neuron` and `neurodamus`
```
WeinaJi:
```suggestion
NEURON is installed, so by activating the virtual environment with `source $USER_VENV/bin/activate` one gets access to `neuron` and `neurodamus`
```

ci/scripts/README.md
mgeplf: > I still think that a short summary paragraph wouldn't hurt. 

I'm not sure what sort of summary you'd like that will be helpful that the Dockerfile.
An example usage is generally `source script.sh && script-function()`

> If it's too easy to forget updating documentation, why are we writing it at all? ;)

It depends on the documentation, IMO, this readme gives a general outline of how the functions work, and what the goal is.  To use them, it's self-explanatory once one has seen the examples in the Dockerfile.

WeinaJi (commented):
There is `ci/scripts/build-neurodamus-models.sh` , and the docker container has a script with the same name. I find it a bit confusing for developers.  Maybe rename `ci/scripts/build-neurodamus-models.sh`  -> `ci/scripts/build-neocortex-model.sh` ?

mgeplf:
> Maybe rename `ci/scripts/build-neurodamus-models.sh` -> `ci/scripts/build-neocortex-model.sh` ?

Good point, done.

README.rst
mgeplf: done

README.rst
mgeplf: done

.github/workflows/simulation_test.yml
mgeplf: I left it as `*_COMMIT`

ci/scripts/install-mpi4py.sh
WeinaJi, line 1: Is `install-mpi4py.sh` used somewhere? I don't find where it is used. We used to installed it manually with the local hdf5 library, maybe it is no longer needed.

mgeplf: I was using it in the machine-image build, but it seems to not need it. I'll remove it.

.github/workflows/simulation_test.yml
WeinaJi: hmm... now my local test `tox -e integration-e2e` complains about not finding NEURON, unless I manually set `export PYTHONPATH=$USER_VENV/lib64/python3.12/site-packages`.  Perhaps tox does only see its "isolated" virtual env. Because NEURON was passed to PYTHONPATH before but now it is installed in the virtual env outside tox. 
We can keep pytest here.

WeinaJi (dismissed):
LGTM! Thanks for the improvements.

@jplanasc , do you have more comments ?

jplanasc:
LGTM, thanks!

WeinaJi:
Great! 
@mgeplf , will you merge it? Then I will make a new release.

WeinaJi (approved):

PR #529, opened 2026-05-26
Compute segment displacements on the fly in apply_all_stimuli

### LINKED ISSUE #524: For Extracellular stimulus, make the segment displacement calculation more tightly coupled to where the final EFieldIntegrator is created
## Description
From [PR521](https://github.com/openbraininstitute/neurodamus/pull/521)

> We should consider computing segment displacements on the fly inside apply_all_stimuli rather than storing them in a dict on ElectrodeSource — this eliminates the intermediate state and explicit cleanup, letting Python's normal scoping handle the lifetime of the temporary displacement vectors, and makes the code easier to follow since computation and consumption happen in one place.

cattabiani:
## Context
Fix: #524

## Scope

Move displacement computation from add_new_stimuli into apply_all_stimuli, eliminating the intermediate segment_displacements dict on ElectrodeSource and the explicit cleanup() call. The displacement vectors now live as local variables scoped to the apply loop iteration.

The geometry computation (segment coordinates + displacements) was done once per stimList entry before, and it's still once per stimList entry now — just deferred from init time to apply time. stimList is already the union of all gids across stimulus blocks, so no work is duplicated.

## Testing

Internal refactoring: tests should be already in place and should not change

## Review

### DIFF
--- a/neurodamus/stimulus_manager.py
+++ b/neurodamus/stimulus_manager.py
@@ -875,22 +875,24 @@ class SpatiallyUniformEField:
         """Apply all consolidated stimuli to their segments.
         Computes segment displacements on the fly from cell geometry.
         """
-        if not cls._instance:
-            return
+        assert cls._instance, (
+            "SpatiallyUniformEField is a singleton but was never instantiated. "
+            "Probably interpret() was not called before apply_all_stimuli()."
+        )
         for gid, es in cls._instance.stimList.items():
             cell = cell_manager.get_cell(gid)
 
             # Compute segment displacements from the ground point (soma barycenter)
             all_seg_points = cell.compute_segment_global_coordinates()
             local_seg_points = cell.compute_segment_local_coordinates()
-            soma = cell.CellRef.soma[0]
-            soma_global_position = np.array(all_seg_points[soma.name()]).mean(axis=0)
-            soma_local_position = np.array(local_seg_points[soma.name()]).mean(axis=0)
+            soma_name = cell.CellRef.soma[0].name()
+            soma_global_position = np.array(all_seg_points[soma_name]).mean(axis=0)
+            soma_local_position = np.array(local_seg_points[soma_name]).mean(axis=0)
 
             def local_to_global(pos, cell=cell, soma_local_position=soma_local_position):
                 return cell.local_to_global_coord_mapping(np.vstack([soma_local_position, pos]))[1]
 
-            segment_displacements = {}
+            segment_displacements = []
             for sec in cell.CellRef.all:
                 for seg in sec:
                     segment_position = (
@@ -905,7 +907,7 @@ class SpatiallyUniformEField:
                         else soma_global_position
                     )
                     displacement_vec = (segment_position - soma_global_position) * 1e-6
-                    segment_displacements[seg] = displacement_vec
+                    segment_displacements.append((seg, displacement_vec))
 
             es.apply_segment_potentials(segment_displacements)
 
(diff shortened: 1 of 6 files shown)


WeinaJi:
LGTM. Thanks for the refactoring. 

@mgeplf , does this change address your comment in #532 ?

neurodamus/stimulus_manager.py
```diff
@@ -870,47 +870,44 @@ def add_new_stimuli(self, target_points, stim_info, cell_manager):
...
-                    # skip sections not in this split
-                    if not sc.exists():
-                        continue
+    @classmethod
+    def apply_all_stimuli(cls, cell_manager):
+        """Apply all consolidated stimuli to their segments.
+        Computes segment displacements on the fly from cell geometry.
+        """
+        if not cls._instance:
+            return
```
mgeplf: does `not cls._instance` ever happen?
should it be an assert?

neurodamus/stimulus_manager.py
```diff
@@ -870,47 +870,44 @@ def add_new_stimuli(self, target_points, stim_info, cell_manager):
...
+        for gid, es in cls._instance.stimList.items():
+            cell = cell_manager.get_cell(gid)
+
+            # Compute segment displacements from the ground point (soma barycenter)
+            all_seg_points = cell.compute_segment_global_coordinates()
+            local_seg_points = cell.compute_segment_local_coordinates()
+            soma = cell.CellRef.soma[0]
```
mgeplf: since only `soma.name()` is used, maybe it should just be gotten here?

neurodamus/stimulus_manager.py
```diff
@@ -870,47 +870,44 @@ def add_new_stimuli(self, target_points, stim_info, cell_manager):
...
+            soma_global_position = np.array(all_seg_points[soma.name()]).mean(axis=0)
+            soma_local_position = np.array(local_seg_points[soma.name()]).mean(axis=0)
+
+            def local_to_global(pos, cell=cell, soma_local_position=soma_local_position):
+                return cell.local_to_global_coord_mapping(np.vstack([soma_local_position, pos]))[1]
+
+            segment_displacements = {}
```
mgeplf: I don't think this needs to be a dictionary, since we never look things up by the key, right?

neurodamus/stimulus_manager.py
cattabiani: it is intentional. In case there are mpi ranks with no extracellular stimuli we can reach this point without any instantiation

cattabiani: added small comment since it is not apparent immediately

cattabiani: EDIT: I double checked and we should always instantiate a class (singleton) even in empty ranks. Added it as an assert

neurodamus/stimulus_manager.py
cattabiani: ok

neurodamus/stimulus_manager.py
cattabiani: true. I just transferred this and did not notice. A list is faster

neurodamus/stimulus_manager.py
mgeplf: thanks

mgeplf (approved):

mgeplf:
@WeinaJi > does this change address your comment in https://github.com/openbraininstitute/neurodamus/pull/532 ?

It's a big improvement, but I still think the coupling is quite tight, but that doesn't need to be fixed here.

PR #532, opened 2026-05-26
Combine multiple extracellular electric field stimulus blocks

### LINKED ISSUE #523: Discussion on generating reference data in tests
## Description
From extracellular stimulus [PR521](https://github.com/openbraininstitute/neurodamus/pull/521)

`
We are comparing a hardcoded reference (produced with the previous version of the code) with another hardcoded function that computes a cosine. You can check this once for your computation but then it will always be the same except if you need to change the function. And if you want to change the function you will need to change the reference. 
`

WeinaJi:
## Context
Part 2 for https://github.com/openbraininstitute/prod-circuit-simulation/issues/172, combine multiple  SpatiallyUniformEField stimulus blocks amd add electric fields into the EFieldIntegrator mechanism.

## Scope
1. In `stimuli.py`, each cell has an object `ElectrodeSource` which contains a list of EFields which are parsed per SpatiallyUniformEField stimulus. 
2. In case of mulitple SpatiallyUniformEField stimulus blocks, for every overlapping cell, its `ElectrodeSource` 's efields list is extended with the efields from the new block thanks to the singleton design pattern.
3. After looping over the stimulus blocks, call each cell's `ElectrodeSource.apply_segment_potentialsply`  that creates a `EFieldIntegrator` mod object with the list of EFields
4. Refactor `efield_integrator.mod` to take a list of e-fields, and loop over the e-fields during `BEFORE BREAKPOINT` process. 

## Testing
1. Include `tests/unit/test_multiple_extracellular_stimuli.py`
2. Refactor and compute references for tests regarding extracellular stimulus, fixes #523 

## Review

### DIFF
--- a/neurodamus/core/stimuli.py
+++ b/neurodamus/core/stimuli.py
@@ -516,9 +516,7 @@ class ElectrodeSource:
             n_fields = len(self.efields)
             freq_vector = Nd.h.Vector(n_fields)
             phase_vector = Nd.h.Vector(n_fields)
-            x_vec = Nd.h.Vector(n_fields)
-            y_vec = Nd.h.Vector(n_fields)
-            z_vec = Nd.h.Vector(n_fields)
+            peak_potential_vec = Nd.h.Vector(n_fields)
             delay_vec = Nd.h.Vector(n_fields)
             duration_vec = Nd.h.Vector(n_fields)
             rup_vec = Nd.h.Vector(n_fields)
@@ -527,9 +525,12 @@ class ElectrodeSource:
             for i, field in enumerate(self.efields):
                 freq_vector.x[i] = field.frequency
                 phase_vector.x[i] = field.phase
-                x_vec.x[i] = field.ex * displacement[0]
-                y_vec.x[i] = field.ey * displacement[1]
-                z_vec.x[i] = field.ez * displacement[2]
+                # dot product E·d in V, converted to mV (* 1e3)
+                peak_potential_vec.x[i] = 1e3 * (
+                    field.ex * displacement[0]
+                    + field.ey * displacement[1]
+                    + field.ez * displacement[2]
+                )
                 delay_vec.x[i] = field.delay
                 duration_vec.x[i] = field.duration
                 rup_vec.x[i] = field.ramp_up_time
@@ -542,9 +543,7 @@ class ElectrodeSource:
                 duration_vec,
                 rup_vec,
                 rdown_vec,
-                x_vec,
-                y_vec,
-                z_vec,
+                peak_potential_vec,
                 phase_vector,
                 freq_vector,
             )
--- a/neurodamus/core/stimuli.py
+++ b/neurodamus/core/stimuli.py
@@ -1,6 +1,7 @@
 """Stimuli sources. inc current and conductance sources which can be attached to cells"""
 
 import logging
+from dataclasses import dataclass
 
 from .configuration import ConfigurationError
 from .random import RNG, gamma
@@ -492,40 +493,45 @@ class ConductanceSource(SignalSource):
         )
 
 
-class ElectrodeSource:
-    """Constructs an extracellular potential field as the sum of multiple user-defined e-fields,
-    and applies the resulting signal to the segment's e_extracellular.
+@dataclass
+class EField:
+    """Dataclass for electric field definition, currently cosinusoid"""
 
-    Args:
-        base_amp: baseline amplitude when signal is inactive
-        delay: start time delay in ms
-        duration: duration of the signal, not including ramp up and ramp down
-        fields: list of user-defined electric field components (e.g. cosinuoid fields)
-        duration: duration of the signal, not including ramp up and ramp down.
-        ramp_up_time: duration during which the signal amplitude ramps up linearly from 0, in ms
-        ramp_down_time: duration during which the signal amplitude ramps down linearly to 0, in ms
+    ex: float  # Peak amplitude of x-direction in in V/m
+    ey: float  # Peak amplitude of y-direction in in V/m
+    ez: float  # Peak amplitude of z-direction in in V/m
+    frequency: float  # Frequency of the cosinusoid wave in Hz
+    phase: float  # Phase of the cosinusoid, in radians
+    duration: float  # duration of the signal, not including ramp up and ramp down in ms
+    delay: float  # start time delay in ms
+    ramp_up_time: float  # duration during which amplitude ramps up linearly from 0, in ms
+    ramp_down_time: float  # duration during which amplitude ramps down linearly to 0, in ms
+
+
+class ElectrodeSource:
+    """Manages extracellular electric field stimulation for a single cell
+    and applies to every segment.extracellular._ref_e via EFieldIntegrator mechanism.
+    EFieldIntegrator computes the potential at every time step as the sum of one or more e-fields,
+    each defined by direction, frequency, phase, delay, ramp up and ramp down,
     """
 
-    def __init__(self, base_amp, delay, duration, fields, ramp_up_time, ramp_down_time, dt):
-        self.time_vec = Nd.h.Vector()  # Time points for stimulus waveform
-        self._cur_t = 0
-        self._base_amp = base_amp
-        self._delay = delay
-        self.fields = fields
-        self.duration = duration
-        self.dt = dt
-        self.ramp_up_time = ramp_up_time
-        self.ramp_down_time = ramp_down_time
+    def __init__(self, delay, duration, fields, ramp_up_time, ramp_down_time):
         self.segment_displacements = {}  # {segment: displacement vectors in x/y/z w.r.t ground}
-        self.segment_efield_integrators = []
-
-    def delay(self, duration):
-        """Increments the ref time so that the next created signal is delayed"""
-        # NOTE: We rely on the fact that Neuron allows "instantaneous" changes
-        # and made all signal shapes return to base_amp. Therefore delay() doesn't
-        # need to introduce any point to avoid interpolation.
-        self._cur_t += duration
-        return self
+        self.segment_efield_integrators = []  # list of EFieldIntegrator attached to segments
+        self.fields = [
+            EField(
+                ex=f["Ex"],
+                ey=f["Ey"],
+                ez=f["Ez"],
+                frequency=f["Frequency"],
+                phase=f["Phase"],
+                duration=duration,
+                delay=delay,
+                ramp_up_time=ramp_up_time,
+                ramp_down_time=ramp_down_time,
+            )
+            for f in fields
+        ]  # list of EFields objects
 
     def apply_segment_potentials(self):
         """Apply potentials to segment.extracellular._ref_e"""
@@ -541,20 +547,20 @@ class ElectrodeSource:
             y_vec = Nd.h.Vector(len(self.fields))
             z_vec = Nd.h.Vector(len(self.fields))
             for i, field in enumerate(self.fields):
-                freq_vector.x[i] = field["Frequency"]
-                phase_vector.x[i] = field["Phase"]
-                x_vec.x[i] = field["Ex"]
-                y_vec.x[i] = field["Ey"]
-                z_vec.x[i] = field["Ez"]
+                freq_vector.x[i] = field.frequency
+                phase_vector.x[i] = field.phase
+                x_vec.x[i] = field.ex
+                y_vec.x[i] = field.ey
+                z_vec.x[i] = field.ez
             efi = Nd.h.EFieldIntegrator(segment)
             Nd.h.setpointer(segment.extracellular._ref_e, "e_ext", efi)
             efi.enabled = 1
             efi.set_displacement(Nd.h.Vector(displacement))
             efi.add_electrode_source(
-                self._delay,
-                self.duration,
-                self.ramp_up_time,
-                self.ramp_down_time,
+                self.fields[0].delay,
+                self.fields[0].duration,
+                self.fields[0].ramp_up_time,
+                self.fields[0].ramp_down_time,
                 x_vec,
                 y_vec,
                 z_vec,
--- a/neurodamus/core/stimuli.py
+++ b/neurodamus/core/stimuli.py
@@ -1,10 +1,10 @@
 """Stimuli sources. inc current and conductance sources which can be attached to cells"""
 
 import logging
-from dataclasses import dataclass
 
 from .random import RNG, gamma
 from neurodamus.core import NeuronWrapper as Nd
+from neurodamus.io.sonata_config import EField
 
 
 class SignalSource:
@@ -490,21 +490,6 @@ class ConductanceSource(SignalSource):
             self._reversal,
             represents_physical_electrode=self._represents_physical_electrode,
         )
-
-
-@dataclass
-class EField:
-    """Dataclass for electric field definition, currently cosinusoid"""
-
-    ex: float  # Peak amplitude of x-direction in in V/m
-    ey: float  # Peak amplitude of y-direction in in V/m
-    ez: float  # Peak amplitude of z-direction in in V/m
-    frequency: float  # Frequency of the cosinusoid wave in Hz
-    phase: float  # Phase of the cosinusoid, in radians
-    duration: float  # duration of the signal, not including ramp up and ramp down in ms
-    delay: float  # start time delay in ms
-    ramp_up_time: float  # duration during which amplitude ramps up linearly from 0, in ms
-    ramp_down_time: float  # duration during which amplitude ramps down linearly to 0, in ms
 
 
 class ElectrodeSource:
--- a/neurodamus/core/stimuli.py
+++ b/neurodamus/core/stimuli.py
@@ -555,9 +555,9 @@ class ElectrodeSource:
             for i, field in enumerate(self.fields):
                 freq_vector.x[i] = field.frequency
                 phase_vector.x[i] = field.phase
-                x_vec.x[i] = field.ex
-                y_vec.x[i] = field.ey
-                z_vec.x[i] = field.ez
+                x_vec.x[i] = field.ex * displacement[0]
+                y_vec.x[i] = field.ey * displacement[1]
+                z_vec.x[i] = field.ez * displacement[2]
                 delay_vec.x[i] = field.delay
                 duration_vec.x[i] = field.duration
                 rup_vec.x[i] = field.ramp_up_time
@@ -565,7 +565,6 @@ class ElectrodeSource:
             efi = Nd.h.EFieldIntegrator(segment)
             Nd.h.setpointer(segment.extracellular._ref_e, "e_ext", efi)
             efi.enabled = 1
-            efi.set_displacement(Nd.h.Vector(displacement))
             efi.add_electrode_source(
                 delay_vec,
                 duration_vec,
--- a/neurodamus/core/stimuli.py
+++ b/neurodamus/core/stimuli.py
@@ -3,7 +3,6 @@
 import logging
 from dataclasses import dataclass
 
-from .configuration import ConfigurationError
 from .random import RNG, gamma
 from neurodamus.core import NeuronWrapper as Nd
 
@@ -517,7 +516,7 @@ class ElectrodeSource:
 
     def __init__(self, delay, duration, fields, ramp_up_time, ramp_down_time):
         self.segment_displacements = {}  # {segment: displacement vectors in x/y/z w.r.t ground}
-        self.segment_efield_integrators = []  # list of EFieldIntegrator attached to segments
+        self.segment_efield_integrators = []  # list of EFieldIntegrator mechs attached to segments
         self.fields = [
             EField(
                 ex=f["Ex"],
@@ -582,11 +581,9 @@ class ElectrodeSource:
 
     def cleanup(self):
         """Clear unused list variable to free memory"""
-        self.efields = None
         self.segment_displacements = None
 
     def __iadd__(self, other):
         """Combined with another ElectrodeSource object"""
-        logging.error("Multiple ElectrodeSource not supported with nmodl")
-        raise ConfigurationError("Support for multiple ElectrodeSources pending")
-        # return self
+        self.fields.extend(other.fields)
+        return self
--- a/neurodamus/core/stimuli.py
+++ b/neurodamus/core/stimuli.py
@@ -1,6 +1,7 @@
 """Stimuli sources. inc current and conductance sources which can be attached to cells"""
 
 import logging
+from dataclasses import dataclass
 
 from .random import RNG, gamma
 from neurodamus.core import NeuronWrapper as Nd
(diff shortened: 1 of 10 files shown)


neurodamus/data/mod/efield_integrator.mod
```diff
@@ -26,109 +23,129 @@ ASSIGNED {
...
+    size = vector_capacity(vX);
+
+    for( i=0; i<size; i++ ) {
+        double ramp_factor = 1;
+        double cur_delay = vector_vec(vdelay)[i];
+        double cur_duration = vector_vec(vduration)[i];
+        double cur_ramp_up = vector_vec(vramp_up)[i];
+        double cur_ramp_down = vector_vec(vramp_down)[i];
+        if( cur_delay < t && t < cur_delay + cur_duration + cur_ramp_up + cur_ramp_down ) {
+            if( cur_delay < t && t < cur_delay + cur_ramp_up ) {
```
mgeplf: this is true, by definition, due to the check above, no?

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -26,109 +23,129 @@ ASSIGNED {
...
+                ramp_factor = (t-cur_delay) / cur_ramp_up;
+            }
+            if( cur_delay + cur_ramp_up + cur_duration < t && t < cur_delay + cur_duration + cur_ramp_up + cur_ramp_down ) {
```
mgeplf: this is just an `else` branch, of the if above, right?

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -26,109 +23,129 @@ ASSIGNED {
...
-       if( delay+ramp_up+duration < t && t < delay+duration+ramp_up+ramp_down ) {
-           ramp_factor = 1 - (t-(delay+ramp_up+duration)) / ramp_down;
-       }
-
-       size = vector_capacity(vX);
-       for( i=0; i<size; i++ ) {
-           double wavefactor = cos(2 * PI * vector_vec(vfreq)[i] / 1000 * (t-delay) + vector_vec(vphase)[i] );
-	   _lefield_accum += 1e3 * ramp_factor * (vector_vec(vX)[i] * wavefactor + vector_vec(vY)[i] * wavefactor + vector_vec(vZ)[i] * wavefactor);
-       }
+    auto *vX = *reinterpret_cast<IvocVect**>(&_p_X);
```
mgeplf: `v[XYZ]` aren't used (size could be retrieved from one of the other vectors

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -26,109 +23,129 @@ ASSIGNED {
...
+    auto *vY = *reinterpret_cast<IvocVect**>(&_p_Y);
+    auto *vZ = *reinterpret_cast<IvocVect**>(&_p_Z);
+    auto *vphase = *reinterpret_cast<IvocVect**>(&_p_phase);
+    auto *vfreq = *reinterpret_cast<IvocVect**>(&_p_frequency);
+    auto *vdelay = *reinterpret_cast<IvocVect**>(&_p_delay);
+    auto *vduration = *reinterpret_cast<IvocVect**>(&_p_duration);
+    auto *vramp_up = *reinterpret_cast<IvocVect**>(&_p_ramp_up);
+    auto *vramp_down = *reinterpret_cast<IvocVect**>(&_p_ramp_down);
+    size = vector_capacity(vX);
```
mgeplf: IMO, declare and define is nicer: `int size`  Can scope it closer to usage, than; same for `i`

neurodamus/stimulus_manager.py
```diff
@@ -927,21 +918,20 @@ def apply_all_stimuli(cls):
...
-        self.delay = float(stim_info["Delay"])  # start time [ms]
-        if not (np.isclose(self.delay % self.dt, 0) or np.isclose(self.delay % self.dt, self.dt)):
-            self.delay = np.ceil(self.delay / self.dt) * self.dt
-            logging.warning(
-                "%s delay %s is not divisible by dt %s, rounded up to the next time point %s",
-                self.__class__.__name__,
-                stim_info["Delay"],
-                self.dt,
-                self.delay,
+        self.efields = [
```
mgeplf: seems a bit over engineered to create a dataclass that is an exact replica of the "config" we've already created/converted from SONATA? I wonder if make sense to either reuse the SONATA config directly, or, move the dataclass to the `config` part of the code, and do the translation there instead of creating a new dictionary?

tests/unit/test_multiple_extracellular_stimuli.py
```diff
@@ -594,28 +295,38 @@ def test_two_blocks_nodeset_overlap(create_tmp_simulation_config_file):
     cell 1,2 contain the 2nd stimluli
     """
 
-    n = Node(create_tmp_simulation_config_file)
-    n.load_targets()
-    n.create_cells()
-    n.enable_stimulus()
+    from neurodamus.core import (
+        NeuronWrapper as Nd,
+    )  # Import at function level, otherwise will impact other tests
```
mgeplf, line 301: `otherwise will impact other tests`; I tried removing all the imports, and the tests passed.
We're using pytest-forked, so I'm wondering how it could impact?

tests/utils.py
```diff
@@ -396,10 +407,52 @@ def load_and_filter(file_path):
...
+    efi: the Neuron EFieldIntegrator mechanism in order to get the amplitude
+    fields: list of Efields"""
+
+    def _f_cos(t, freq, phase, amp):
+        """compute the cosine value"""
+        return amp * np.cos(2 * np.pi * freq / 1000 * t + phase)
+
+    def _make_ramp_envelope(t_vec, ramp_up_time, ramp_down_time, duration):
+        """With a given time vector, return a vector taking into account ramp up and down time"""
+        envelope = np.ones(len(t_vec))
```
mgeplf: does using `np.clip` make this cleaner?

tests/utils.py
```diff
@@ -396,10 +407,52 @@ def load_and_filter(file_path):
...
+        envelope = np.where(t_vec < ramp_up_time, t_vec / ramp_up_time, envelope)
+        envelope = np.where(
+            t_vec > ramp_up_time + duration,
+            1 - (t_vec - (ramp_up_time + duration)) / ramp_down_time,
+            envelope,
+        )
+        return envelope
+
+    ref_dend = np.zeros(len(tot_tvec))
```
mgeplf: it doesn't have to be a `dend` right, can also be soma or other neurites?

neurodamus/core/stimuli.py
```diff
@@ -492,40 +492,32 @@ def attach_to(self, section, position=0.5):
...
-        """Increments the ref time so that the next created signal is delayed"""
-        # NOTE: We rely on the fact that Neuron allows "instantaneous" changes
-        # and made all signal shapes return to base_amp. Therefore delay() doesn't
-        # need to introduce any point to avoid interpolation.
-        self._cur_t += duration
-        return self
+    def __init__(self, efields: list[EField]):
+        self.segment_displacements = {}  # {segment object: displacement vec in x/y/z w.r.t ground}
+        self.segment_efield_integrators = []  # list of EFieldIntegrator mechs attached to segments
+        self.efields = efields  # list of EFields objects
```
mgeplf: this comment isn't really helpful, imo

neurodamus/core/stimuli.py
```diff
@@ -492,40 +492,32 @@ def attach_to(self, section, position=0.5):
...
-    def delay(self, duration):
-        """Increments the ref time so that the next created signal is delayed"""
-        # NOTE: We rely on the fact that Neuron allows "instantaneous" changes
-        # and made all signal shapes return to base_amp. Therefore delay() doesn't
-        # need to introduce any point to avoid interpolation.
-        self._cur_t += duration
-        return self
+    def __init__(self, efields: list[EField]):
+        self.segment_displacements = {}  # {segment object: displacement vec in x/y/z w.r.t ground}
+        self.segment_efield_integrators = []  # list of EFieldIntegrator mechs attached to segments
```
mgeplf: do they need to be saved for a particular reason?

neurodamus/core/stimuli.py
```diff
@@ -492,40 +492,32 @@ def attach_to(self, section, position=0.5):
...
-
-    def delay(self, duration):
-        """Increments the ref time so that the next created signal is delayed"""
-        # NOTE: We rely on the fact that Neuron allows "instantaneous" changes
-        # and made all signal shapes return to base_amp. Therefore delay() doesn't
-        # need to introduce any point to avoid interpolation.
-        self._cur_t += duration
-        return self
+    def __init__(self, efields: list[EField]):
+        self.segment_displacements = {}  # {segment object: displacement vec in x/y/z w.r.t ground}
```
mgeplf, line 503: I find this pattern fragile; `segment_displacements` is a required property, but it just randomly gets set by a call in `add_new_stimuli`.

I think it could be made more clear by having the `ElectrodeSource` take the gid and target points, and have it compute the required info.
Alternatively, the `ElectrodeSource could take the `segment_displacements` as a parameter for `__init__`

cattabiani, line 503: I would have preferred to merge my pr first:

https://github.com/openbraininstitute/neurodamus/pull/529

since it simplifies this but ok. If you prefer to go first I can take 1 hour to fix all the conflicts will come

@mgeplf I already had similar complaints and I proposed a solution that does not instantiate this var at all (same pr as before)

WeinaJi: Maybe that will be solved by #529

neurodamus/core/stimuli.py
WeinaJi: The EFieldIntegrator mechs ? Yes, otherwise they will be removed by Python garbage collector and won't be used by the NEURON solver.

neurodamus/data/mod/efield_integrator.mod
cattabiani: this part is redundant: `cur_delay < t`. The other part is not (`t < cur_delay + cur_ramp_up`). I think you meant that

neurodamus/data/mod/efield_integrator.mod
cattabiani: no there is also the ` cur_duration` in addition:

`cur_delay + cur_ramp_up =/= cur_delay + cur_ramp_up + cur_duration`

neurodamus/core/stimuli.py
mgeplf: that's what I thought; I think that's a much more important thing to note in the comment

neurodamus/data/mod/efield_integrator.mod
WeinaJi: Ah, right. I forgot to clean up after moving the logic into function `get_potential_amplitude(idx)`. I will remove them.

neurodamus/stimulus_manager.py
cattabiani: I agree that the EField dataclass is redundant with the config dictionary we already have. Instead of introducing a new dataclass that just mirrors the config, we could keep using the config dict (or a lightweight wrapper from the config layer) and implement a simple merge/accumulate method on ElectrodeSource itself — so that when multiple stimulus blocks target the same cell, we just append the field definitions into a single list. The ElectrodeSource stays a flat container of disjoint field entries, and the += operator handles the accumulation without needing an intermediate data structure.

tests/utils.py
```diff
@@ -1,31 +1,29 @@
...
 from neurodamus.report_parameters import create_report_parameters
-
-from typing import List, Dict, Tuple
-import pandas as pd
-import copy
+from neurodamus.target_manager import TargetManager, TargetSpec
 
 
-def merge_dicts(parent: dict, child: dict):
+def merge_dicts(parent: dict, child: dict):  # noqa: C901
```
cattabiani: I do not really like to ignore stuff from ruff. Maybe we can simplify the function by extracting `merge_vals`:

```
def _merge_vals(k, parent, child):
```

and prepending it with `_`?

cattabiani (commented):
some comments

neurodamus/stimulus_manager.py
WeinaJi: It is different from the data structure in Sonata SPEC/Config. 
In SPEC, we have per stimulus
`duration, delay, ramp_up, ramp_down, fields: [(Ex/y/z, freq, phase)]`
Here we need `fields: [(duration, delay, ramp_up, ramp_down, Ex/y/z, freq, phase)]`
I don't use this dataclass in sonata_config.py because that's not the same structure as we get directly from libsonata.

neurodamus/data/mod/efield_integrator.mod
WeinaJi: The outermost `if` requires to be in the total time interval. 
Then inside first `if` for within the ramp up time, second `if` for within the ramp down time, the rest is the flat `ramp_factor=1` time.
But true I could remove the cur_delay < t in the `if` for ramp_up.

neurodamus/stimulus_manager.py
mgeplf: If the conversion is necessary, I would still say it's cleaner to do it in the config portion.
I find the extra layers confusing; for instance, the sonata name `ramp_up_time` gets translated to `RampUpTime` for the (untyped) dictionary, and then converted back to `ramp_up_time` here.

cattabiani:
I have a small suggestion on how to improve the integrator:

https://github.com/openbraininstitute/neurodamus/pull/533

neurodamus/data/mod/efield_integrator.mod
cattabiani: it is mostly a style choice: either you keep the 2 if and remove the external one or you remove these 2 checks that are redundant inside the interior ifs

WeinaJi: > you remove these 2 checks that are redundant inside the interior ifs

Done, please check again.

neurodamus/data/mod/efield_integrator.mod
WeinaJi: Now they are removed.

neurodamus/data/mod/efield_integrator.mod
WeinaJi: Done for `i`, I also replace `size` with `vector_capacity(vdelay)` because it is only used by the for loop.

neurodamus/core/stimuli.py
WeinaJi: Yes, removed.

neurodamus/core/stimuli.py
WeinaJi: I've added a comment here.

neurodamus/core/stimuli.py
WeinaJi: I prefer to merge this PR first because it brings back `test_multiple_extracellular_stimuli.py` which will help review all the following refactoring PRs.

neurodamus/stimulus_manager.py
WeinaJi: I've moved the `EField` dataclass to `sonata_config.py`, please check again.

tests/utils.py
WeinaJi: I didn't know that one,  now it is used.

tests/utils.py
WeinaJi: Done

tests/utils.py
WeinaJi: Right, it was directly copied from the test file and I forgot to rename. Now it is done.

tests/unit/test_multiple_extracellular_stimuli.py
WeinaJi: I've seen a lot of this error msg printed out but it doesn't kill the process. Does it bother us to see them? 
```
NEURON: POINT_PROCESS data instance not valid
 near line 0
 ^
        ProbAMPANMDA_EMS[1].clearRNG()
```
I don't know how it has been solved before it stopped the process. I could move the import at the file level for the tests in the PR and later refactor the other tests.

mgeplf: ok, let's worry about it another time, and leave it how it is.

Thanks for looking.

WeinaJi: It may still impact other tests, e.g. `tests/unit/test_cli_shm.py` fails now. I will open an issue to check later( I thought we already opened an issue for that).

mgeplf: super strange; I don't see how that can jump the fork boundary.

If you find/open the issue, tag me and I'll try and have a look sometime.

WeinaJi:
@mgeplf @cattabiani  review comments have been addressed. Please review again.
After this PR is merged, the tests added here will help validate PR #529.

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -23,112 +20,114 @@ ASSIGNED {
...
+        double cur_ramp_down = vector_vec(vramp_down)[i];
+        if( cur_delay < t && t < cur_delay + cur_duration + cur_ramp_up + cur_ramp_down ) {
+            if( t < cur_delay + cur_ramp_up ) {
+                ramp_factor = (t-cur_delay) / cur_ramp_up;
+            }
+            if( cur_delay + cur_ramp_up + cur_duration < t ) {
+                ramp_factor = 1 - (t - (cur_delay + cur_ramp_up + cur_duration)) / cur_ramp_down;
+            }
+            double wavefactor = cos(2 * PI * vector_vec(vfreq)[i] / 1000 * (t-cur_delay) + vector_vec(vphase)[i]);
+            _lefield_accum += ramp_factor * get_peak_potential(i) * wavefactor;
```
mgeplf, line 129: I'm not sure `get_peak_potential` is worth having a function for, since it's just a call to `vector_vec(vpeak)[idx]`; I assume it was added for testing, but in effect, the test is only checking that the parameter is passed right, which I don't think is necessary.

neurodamus/stimulus_manager.py
```diff
@@ -861,24 +861,12 @@ def add_new_stimuli(self, target_points, stim_info, cell_manager):
...
-                ramp_up_time=self.ramp_up_time,
-                ramp_down_time=self.ramp_down_time,
-                dt=self.dt,
-            )
+            es = ElectrodeSource(stim_info["Fields"])
             gid = target_point_list.gid
             if gid in self.stimList:
-                # Consolidate with existing stimulus - not yet supported in mod impl
-                # TODO: update iadd method to allow: self.stimList[gid] += es; raise error for now
+                # Consolidate with existing stimulus
```
mgeplf: don't think this comment is needed; it just describes using `+=`

tests/unit/test_multiple_extracellular_stimuli.py
```diff
@@ -337,46 +60,47 @@ def test_combine_time_stim_vectors():  # noqa: PLR0915
...
+    dend_efi = es.segment_efield_integrators[3]
+    assert np.isclose(dend_efi.get_peak_potential(0), -0.505702)
+    assert np.isclose(dend_efi.get_peak_potential(1), -0.964335)
+
+    tot_tvec = np.concatenate([[0], np.arange(Nd.dt / 2, Nd.tstop, Nd.dt)])
+    ref_soma = np.zeros(len(tot_tvec))
+    ref_dend = get_expected_extracellular_potentials(tot_tvec, dend_efi, es.efields)
 
-    n.clear_model()
+    npt.assert_allclose(rec_soma, ref_soma)
```
mgeplf: can't this check be all(rec_soma == 0) (same with other instances)

neurodamus/core/stimuli.py
mgeplf: whichever choice is made for who merges whom, please let me review the last PR of the chain

mgeplf (commented):
I'm fine with everything, small nits; I assume the output is the same as before?

neurodamus/data/mod/efield_integrator.mod
WeinaJi: The test also checks the generated cosine wave for which I need the peak amplitude. 
Because I clean the vector `segment_displacements` on the python side at the end, and soon it will be computed on the fly in #529, the amplitude only exists in MOD file as POINTER. Can I access the mod  POINTER `peak_potential` on the python side? If yes, this function can be removed.

mgeplf: I don't know of a way to introspect `POINTER`; thanks for the explanation.  I guess it's the best we can do.

tests/unit/test_multiple_extracellular_stimuli.py
WeinaJi: Done.

mgeplf: thanks.

neurodamus/stimulus_manager.py
WeinaJi: Done, the comment is removed.

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -23,112 +20,114 @@ ASSIGNED {
...
+        double ramp_factor = 1;
+        double cur_delay = vector_vec(vdelay)[i];
+        double cur_duration = vector_vec(vduration)[i];
+        double cur_ramp_up = vector_vec(vramp_up)[i];
+        double cur_ramp_down = vector_vec(vramp_down)[i];
+        if( cur_delay < t && t < cur_delay + cur_duration + cur_ramp_up + cur_ramp_down ) {
+            if( t < cur_delay + cur_ramp_up ) {
+                ramp_factor = (t-cur_delay) / cur_ramp_up;
+            }
+            if( cur_delay + cur_ramp_up + cur_duration < t ) {
```
mgeplf: this can be `else if` so it's clear they don't overlap?

WeinaJi: How do they overlap? cur_duration is always positive. 
Maybe I should write the 2nd if 
`if( t > (cur_delay + cur_ramp_up + cur_duration ) )`
to make it clearer.

mgeplf: they *don't* overlap; it's impossible for both `if( t < cur_delay + cur_ramp_up )` and `if( cur_delay + cur_ramp_up + cur_duration < t )` to be true at the same time, right?

In cases like that, I think it makes more sense to write:
```
if( t < cur_delay + cur_ramp_up ) {
[...]
}else if( cur_delay + cur_ramp_up + cur_duration < t ) {
```
to make it clear.

WeinaJi: ok, I've added the `else if` to be clearer and slightly more efficient.

mgeplf: thanks

mgeplf:
I'm good, I'll leave it up to @cattabiani for his review.

WeinaJi:
Thanks for the reviews @mgeplf. 
@cattabiani, do you still have comments?

neurodamus/stimulus_manager.py
cattabiani: LGTM

cattabiani (approved):
LGTM

WeinaJi:
> LGTM

Thanks @cattabiani. After merging this PR, I will check #529

PR #533, opened 2026-05-27
Simplify EFieldIntegrator: pre-compute peak potential on Python side

cattabiani:
## Scope

Replace the 3 separate X, Y, Z vectors in the mod file with a single peak_potential vector. The dot product E·d is now fully computed in Python before being passed to the mechanism, removing redundant storage and making the mod file's intent clearer.

Rename get_potential_amplitude -> get_peak_potential to better reflect that the value is the unscaled peak potential (before ramp/cosine).

## Review

### DIFF
--- a/neurodamus/data/mod/efield_integrator.mod
+++ b/neurodamus/data/mod/efield_integrator.mod
@@ -1,7 +1,7 @@
 NEURON {
     POINT_PROCESS EFieldIntegrator
     POINTER e_ext
-    POINTER phase, frequency, X, Y, Z, delay, duration, ramp_up, ramp_down
+    POINTER phase, frequency, peak_potential, delay, duration, ramp_up, ramp_down
     RANGE enabled  : set when e_ext pointer is assigned, this prevents mcomplex calculation to access unassigned e_ext reference
 }
 
@@ -20,9 +20,7 @@ ASSIGNED {
     ramp_down
     phase
     frequency
-    X
-    Y
-    Z
+    peak_potential
     enabled
 }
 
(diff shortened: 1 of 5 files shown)


mgeplf:
isn't @WeinaJi working in this code at the moment? won't this complicate her changes?

cattabiani:
@mgeplf this IS a suggestion on top of Weina's changes. It was just a little too complex to add it as an inline suggestion to the pr review

neurodamus/data/mod/efield_integrator.mod
```diff
@@ -144,7 +129,7 @@ VERBATIM
                 ramp_factor = 1 - (t - (cur_delay + cur_ramp_up + cur_duration)) / cur_ramp_down;
             }
             double wavefactor = cos(2 * PI * vector_vec(vfreq)[i] / 1000 * (t-cur_delay) + vector_vec(vphase)[i] );
-            _lefield_accum += ramp_factor * get_potential_amplitude(i) * wavefactor;
+            _lefield_accum += ramp_factor * vector_vec(vpeak)[i] * wavefactor;
```
WeinaJi: It is better to call `get_peak_potential(i)` here, otherwise this function will never be called inside this mod file.

WeinaJi:
Either I cherry-pick the change into #532 , or we merge this PR after #532. As you prefer.

mgeplf:
> this IS a suggestion on top of Weina's changes. It was just a little too complex to add it as an inline suggestion to the pr review

Ah, ok; not used to this workflow, I hardly ever look at the target branch.

neurodamus/data/mod/efield_integrator.mod
cattabiani: true but it does exactly the same. Since it is just a lookup do you want to remove it altogether and remove the test or do you prefer to keep it? it is just a lookup, I do not think it deserves to be tested

cattabiani: I think it is better if f_cos does not re-use methods inside the .mod file if possible. It should arrive at the same result independently so that we do not use possibly bugged code in the reference (in case there is a bug)

cattabiani: ok, it is still used in the tests and getting this outside in the tests takes a lot of effort. it is just better to use this small service function. Done

cattabiani:
@WeinaJi this branch is on top of yours (updated to latest version in repo). You can just merge into your pr (without cherry-picking) if you think it is a good addition

PR #537, opened 2026-06-01
Avoid shared reference in ElectrodeSource.efields to fix mulitple elecric fields stimulus on mulitple cells

### LINKED ISSUE #538: Avoid shared reference in ElectrodeSource.efields
I discuss this issue with ToyCircuit-S1-6k circuit by comparing the simulation results from these 2 equivalent electric fields setting:
1. sim1: single `spatially_uniform_e_field` input on all cells with two e-fields 

[simulation_config_singleefield.json](https://github.com/user-attachments/files/28460858/simulation_config_singleefield.json)

2. sim2: 2 `spatially_uniform_e_field` inputs  on all cells each contains one e-field 

[simulation_config_efields.json](https://github.com/user-attachments/files/28460865/simulation_config_efields.json)

These two simulations should produce the same results but they didn't. After looking at the e-fields produced at run time, I discovered that the sim 2 case (multiple spatially_uniform_e_field inputs) had more e-fields for each cell (3 instead of 2). The root cause is that `ElectrodeSource.efields` for each cell is instantiated with a shared reference 
https://github.com/openbraininstitute/neurodamus/blob/edb273e513cec4e42fb9077c2955cdcb18eb8277/neurodamus/stimulus_manager.py#L866
https://github.com/openbraininstitute/neurodamus/blob/edb273e513cec4e42fb9077c2955cdcb18eb8277/neurodamus/core/stimuli.py#L505
At the time of `__iadd__`, `ElectrodeSource.efields` gets extended with the e-fields from the 2nd `spatially_uniform_e_field` input. If it is shared reference, `ElectrodeSource.efields` of cell A gets extended at the time of `__iadd__` for cell B, cell C ... etc. That's wrong. 

Solution: `ElectrodeSource.efields` should have an independent copy.  Should be fixed for https://github.com/openbraininstitute/prod-circuit-simulation/issues/172

WeinaJi:
## Context
Fixes #538 
The current `ElectrodeSource.efields` for each cell is instantiated with a shared reference.  At the time of `__iadd__`, `ElectrodeSource.efields` gets extended with the e-fields from the 2nd spatially_uniform_e_field input. If it is shared reference, `ElectrodeSource.efields` of cell A gets extended at the time of `__iadd__` for cell B, cell C ... etc. That's wrong. 
This PR makes `ElectrodeSource.efields` to hold an independent copy.

## Testing
1: `tests/unit/test_multiple_extracellular_stimuli.py`: change the input target to the same muiltiple cells that could reveal the issue.

## Review

### DIFF
--- a/neurodamus/core/stimuli.py
+++ b/neurodamus/core/stimuli.py
@@ -502,7 +502,7 @@ class ElectrodeSource:
     def __init__(self, efields: list[EField]):
         # list of EFieldIntegrator mechs attached to segments, required to avoid garbage collection
         self.segment_efield_integrators = []
-        self.efields = efields
+        self.efields = efields.copy()  # copy to avoid shared reference when combined via __iadd__
 
     def apply_segment_potentials(self, segment_displacements):
         """Apply potentials to segment.extracellular._ref_e
--- a/tests/unit/test_multiple_extracellular_stimuli.py
+++ b/tests/unit/test_multiple_extracellular_stimuli.py
@@ -32,7 +32,7 @@ from neurodamus.stimulus_manager import SpatiallyUniformEField
                         "module": "spatially_uniform_e_field",
                         "delay": 0,
                         "duration": 5,
-                        "node_set": "RingA_Cell0",
+                        "node_set": "RingA",
                         "fields": [
                             {"Ex": 50, "Ey": -25, "Ez": 75, "frequency": 100, "phase": 1.570796},
                         ],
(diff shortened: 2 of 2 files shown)


neurodamus/core/stimuli.py
```diff
@@ -502,7 +502,7 @@ class ElectrodeSource:
     def __init__(self, efields: list[EField]):
         # list of EFieldIntegrator mechs attached to segments, required to avoid garbage collection
         self.segment_efield_integrators = []
-        self.efields = efields
+        self.efields = efields.copy()  # copy to avoid shared reference when combined via __iadd__
```
mgeplf, line 505: can we make it a set, and then it will handle duplicates for us?
you may have to make the dataclass `frozen` so they are hashable.

WeinaJi: Why do we reject duplicates? A user may inject 2 fields with the same values on purpose.

mgeplf: True, good point.

mgeplf (approved):

cattabiani (approved):
LGTM

PR #539, opened 2026-06-01
Publish to PyPI on GitHub Release

cattabiani:
Aligns the publish workflow with BlueRecording and cookiecutter-python-lib:

- Trigger PyPI publish on GitHub Release (published) instead of tag push
- Scope push trigger to main branch only
- Use fetch-depth: 0 for reliable hatch-vcs version resolution
- Bump action versions (setup-python@v6, checkout@v6, upload-artifact@v7, download-artifact@v8)
- Move publish condition to job level to skip unnecessary steps
- Keep existing wheel test step

Requested by @eleftherioszisis . I also think that we should standardize this

### DIFF
--- a/.github/workflows/publish_wheels.yml
+++ b/.github/workflows/publish_wheels.yml
@@ -1,46 +1,58 @@
-name: Neurodamus Wheels
+name: Publish to PyPI
 
-on: [pull_request, push, workflow_dispatch]
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  release:
+    types: [published]
+  workflow_dispatch:
 
 jobs:
   wheel:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python_version: [3.11]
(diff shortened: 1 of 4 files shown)


.github/workflows/publish_wheels.yml
```diff
@@ -1,46 +1,58 @@
...
+    types: [published]
+  workflow_dispatch:
 
 jobs:
   wheel:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python_version: [3.11]
+        python_version: ["3.11"]
```
eleftherioszisis: Is the matrix needed here?

cattabiani: no. I was lazy and used the old way. Changed

WeinaJi:
LGTM. I will let the others to give the final approval.

eleftherioszisis (approved):

Issue #536, opened 2026-06-01
When populating the resistance vector for relative_ornstein_uhlenbeck stimulus, update logic to avoid nonsensical values


jamesgking:
## Description
From https://github.com/openbraininstitute/prod-circuit-simulation/issues/157, running the simulation_config.json has some gids enter a bad state where Vm goes extreme and finally nan

This happens because the values generated for the rs value used in ConductanceSource gets close to zero when interpolating between values

| time   | rs |
| -----  | ------ |
| 386.75  | 543.196232 |
| 387.0  | 658.993148 |
| 387.25 | 1679.262295 |
| 387.5 | 2062.264701 |
| 387.75 | -2521.630022 |
| 388.0 | 2240.755025 |

This vector was populated using code:
>self.stim_vec = Nd.h.Vector(
>[1 / x if abs(x) > 1e-9 else (1e9 if x >= 0 else -1e9) for x in self.stim_vec]
>)
>self.stim_vec.play(self.clamp._ref_rs, self.time_vec, 1)

## Expected behavior
The target rs values should be prevented from taking on invalid values. 

## Acceptance criteria
- [ ] New logic that builds a vector for rs which is valid

### Type
Bug

### Projects
Select 'OBI 2025-' project (plus others as relevant)
P0

jamesgking:
Discussing with @cattabiani , he suggests we switch to Reflected Ornstein–Uhlenbeck Process to avoid negative values

@MWolfR @james-isbister for your awareness

mgeplf:
Hmm, I think this actually a result of refactoring in this PR:
https://github.com/openbraininstitute/neurodamus/pull/69/

It [went from](https://github.com/openbraininstitute/neurodamus/pull/69/changes#diff-927ecaa877e4808d72fa8bcd137edcc2f2da370a2f7a1947097bf9316f670163L493):
`[1 / x if x > 1e-9 and x < 1e9 else 1e9 for x in self.stim_vec]`
to
`[1 / x if abs(x) > 1e-9 else (1e9 if x >= 0 else -1e9) for x in self.stim_vec]`

cattabiani:
yes, I remember that change and it improved already the line. Before there was no limit on how negative 1/x could be. My solution was just placing limits to the negative part too. 

I did not want to change the general behavior so I left the possibility to go negative (even if it was a little strange to me at the time)

IMHO:
[...]

PR #540, opened 2026-06-02
Update how stimvec values played in ConductanceSource.rs are clipped

jamesgking:
## Context
Update how stimvec values played in ConductanceSource.rs are clipped

Depending on distribution parameters, it's possible negative values are generated in the random sequence. These are clipped back to 1e9 as was done in previous implementation. In fact, values generated greater than 1e9 are also clipped to 1e9.

see issue https://github.com/openbraininstitute/neurodamus/issues/536

## Testing
New unit test where a rng stream with negative values are converted to 1e9

## Review


tests/unit/test_stimuli.py
```diff
@@ -489,3 +489,34 @@ def setup_method(self):
...
+            1e9,
+            10.381426494807684,
+            14.874458020035506,
+            18.52106302047261,
+            1e9,
+            self.base_amp,
+        ]
+        assert np.allclose(
+            dynclamp.stim_vec, expected_stim_vec
+        ), f"{list(dynclamp.stim_vec)}, {list(expected_stim_vec)}"
```
cattabiani: nitpick: if you want, you can you add a test to check that it does not jump before and after 1e9. Not necessary tho

cattabiani (commented):
LGTM (if linting and tests pass)

reask once they pass

mgeplf (dismissed):
seems reasonable

tests/scientific/test_sonata_properties.py
```diff
@@ -56,7 +56,7 @@ def test_input_resistance():
 
     # check spikes
     nspike = sum(len(spikes) for spikes, _ in n._spike_vecs)
-    assert nspike == 10
```
mgeplf, line 59: why is this different, though?
this comparison was from 2024-10-24, which predates the change.

Or is it the difference between the original `[1 / x if x > 1e-9 and x < 1e9 else 1e9 for x in self.stim_vec]` and `[1 / x if x > 1e-9 else 1e9 for x in self.stim_vec]`?

edit: fixed duplicate cut and paste

cattabiani: oh I have never seen the og

jamesgking: Using Sirio's stim_vec code, I still get 11 spikes in the branch. I installed neurodamus 3.6.0 and there we have 10 spikes. I'll see if I can bisect where this would have chenged

jamesgking: Okay, I see this is related to switching to 0-based gids and etypes using StochKv (e.g. STUT)

It was by coincidence that with both changes (stim_vec+gid) that there are still 10 spikes. By reverting so only gid change is present, then we get 11 spikes.

I think a separate conversation with the scientists is necessary for handling seeding on StochKv. I will follow up with BlueCellulab devs as well for their thoughts

mgeplf: > Okay, I see this is related to switching to 0-based gids

Which commit are you referring to?

jamesgking: PR https://github.com/openbraininstitute/neurodamus/pull/391
which went into release 4.0.0

tests/unit/test_stimuli.py
jamesgking: Will go back to Sirio's original implementation

tests/scientific/test_sonata_properties.py
jamesgking: I've opened a ticket for discussion with scientists and bluecellulab devs on StochKv using etypes

https://github.com/openbraininstitute/prod-singlecell-simulation/issues/289

mgeplf: Huh, so the argument is that:
Oct 2024 > assert nspike == 10
Feb 2025 > change that allows negative rs; spike count remains 10
Oct 2025 > change to 0 base, causing StockKv seeding to change; spike count remains 10
Now > revert Feb 2025 change, spikes changes to 11

I wonder if we can change the StockKv seeding to be the same as before, and see if it returns the spikes to 10.

I still think we should simpler channels/etc to do our testing, so we can reason about them, and don't have to deal with seeding problems.

jamesgking: > I wonder if we can change the StockKv seeding to be the same as before, and see if it returns the spikes to 10.

Yes, changing the hoc file to add 1 causes the spikes to go back to the original 10

> I still think we should simpler channels/etc to do our testing, so we can reason about them, and don't have to deal with seeding problems.

I hope we can decide that with the scientists using that ticket I opened

mgeplf: > Yes, changing the hoc file to add 1 causes the spikes to go back to the original 10

Nice.

> I hope we can decide that with the scientists using that ticket I opened

I don't think we need their input for that; it's an internal check for neurodamus.
Especially as we move into a world where they are providing the .mod files with the circuit.

tests/unit/test_stimuli.py
```diff
@@ -489,3 +489,34 @@ def setup_method(self):
...
+            27.655225425824607,
+            18.850572907882057,
+            1e9,
+            10.381426494807684,
+            14.874458020035506,
+            18.52106302047261,
+            1e9,
+            self.base_amp,
+        ]
+        assert np.allclose(
```
mgeplf: sorry, mised it before, but I think https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_equal.html provides a nice message; does it not work in this case?

jamesgking: Done

mgeplf (approved):

PR #543, opened 2026-06-03
work around pytests trying to load `Nd`

mgeplf:
* many tests lazy load `NeuronWrapper` since importing it at the module level causes strange behavior with pytest & pytest-forked.
* this is due to `NeuronWrapper` having its `_pytestfixturefunction` queried, which triggers `__getattr__`, which then loads NEURON.
* this happens before pytest-forked forks the process, so NEURON's state is already initialized
* `NeuronWrapper`s purpose, from what I can tell, is to delay this loading; if `_pytestfixturefunction` is defined, the dynamic `__getattr__` won't be executed, and `NeuronWRapper`s lazy loading of NEURON is preserved.


eleftherioszisis (approved):
lgtm

PR #544, opened 2026-06-03
Add unit-local tox env for testing with locally-built NEURON

### LINKED ISSUE #542: Allow running unit tests against a locally-built NEURON (MPI-enabled)
## Problem

When developing breaking changes in NEURON, we need to validate against neurodamus unit tests before merging. The `unit` tox env installs `neuron-nightly` (no MPI) from PyPI. There's no way to test with a local NEURON build (which has MPI enabled).

The `integration-e2e` and `scientific` envs already use the local NEURON via `passenv = *`. We need the same for unit tests.

## Why it's not trivial

The unit test suite requires `--forked` (pytest-forked) because NEURON/neurodamus uses global singletons (`SimConfig`, `Neuron.h`) that leak state between tests. Without `--forked`, tests fail due to pollution from previous tests.

But a locally-built NEURON has MPI. This creates an impossible triangle:

| Configuration | Result |
|---|---|
| `NEURON_INIT_MPI=0` + `--forked` | ✅ Pure tests pass. ❌ Subprocess CLI tests fail (need MPI context for CoreNEURON) |
| `NEURON_INIT_MPI=1` + `--forked` | ❌ macOS crash dialogs — fork after `MPI_Init` is undefined behavior, OpenMPI aborts the forked children (see [OpenMPI docs: fork() in MPI processes](https://docs.open-mpi.org/en/main/tuning-apps/fork-system-popen.html)) |
| `NEURON_INIT_MPI=1` + no `--forked` | ❌ Test isolation failures — singleton state leaks between tests |

Additionally, subprocess-based tests (`test_cli.py`, `test_coreneuron_save_restore.py`, `test_dump_cell_state.py`) produce `out.h5` because the local NEURON is linked against libsonatareport, causing assertion failures on expected output files.

cattabiani:
## Summary

Adds a `unit-local` tox environment that runs unit tests against a locally-built NEURON (with MPI) instead of `neuron-nightly` from PyPI.

This allows validating NEURON breaking changes against neurodamus before merging to NEURON main.

Fix: #542

## How it works

- Uses `passenv = *` to inherit the local NEURON from PATH/PYTHONPATH
- Sets `NEURON_INIT_MPI=1` + `--forked` (same as integration/scientific)
- Deselects 1 test that requires CoreNEURON reports in a subprocess (needs mpirun)
- `test_coreneuron_save_restore.py` has `out.h5` accepted as optional

## Result

334 passed, 1 deselected, 0 failed (out of 340 total unit tests).

## Usage

```bash
# With local NEURON on PATH (e.g. via .envrc):
tox -e unit-local
```


tox.ini
```diff
@@ -37,6 +39,25 @@ commands =
...
+# Like unit but uses the local environment (NEURON, libsonatareport, neurodamus-models)
+# instead of neuron-nightly. Useful for testing breaking changes before merging.
+deps =
+    {[testenv]deps}
+    morphio
+    memray
+passenv = *
+setenv =
+    NEURON_INIT_MPI=1
+    PYTHONPATH={toxinidir}:{env:PYTHONPATH:}
```
WeinaJi, line 55: It will have an issue if the local neuron is built with pip install in another virtual env and tox only sees its isolate virtual env. I discovered that when reviewing https://github.com/openbraininstitute/neurodamus/pull/528. 

>hmm... now my local test tox -e integration-e2e complains about not finding NEURON, unless I manually set export PYTHONPATH=$USER_VENV/lib64/python3.12/site-packages. Perhaps tox does only see its "isolated" virtual env. Because NEURON was passed to PYTHONPATH before but now it is installed in the virtual env outside tox.
We can keep pytest here.

Another question, why do we need to address it in tox.ini? In local, we can always use pytest after all the env vars are set instead of tox.

cattabiani: 1. I tried various ways but yes, setting the pythonpath is the easiest solution.  It's a tox limitation, the workaround is setting PYTHONPATH, and it's now documented

2. `tox -m local` gives a single command that runs format + all test suites with the right deselects and env vars. With bare pytest you'd need to replicate that manually each time. AFAIK there's no built-in way to make pytest run "the same as tox". tox adds the format check, the ndcore build step (`build_ndcore.sh`), the `.envfile` sourcing, and specific deselects. You'd have to script that yourself, at which point you've just rebuilt tox. My objective is to avoid (as much as possible) to add a feature, tests only some tests or in a way that is not exactly the same as tox, merge the feature in neuron and then have another test in neurodamus that fails because I did not think about that

tox.ini
```diff
@@ -37,6 +39,28 @@ commands =
...
+    PYTEST_DEBUG_TEMPROOT={env:PYTEST_DEBUG_TEMPROOT}
+allowlist_externals =
+    {toxinidir}/ci/build_ndcore.sh
+    sh
+commands =
+    {toxinidir}/ci/build_ndcore.sh {toxinidir}/neurodamus/data {toxworkdir}
+    sh -c ". {toxworkdir}/.envfile && pytest -x --forked --deselect=tests/unit/test_cli.py::test_cli_report_buff_size --deselect=tests/unit/test_coreneuron_save_restore.py::test_full_run_vs_save_restore {posargs:tests/unit}"
```
WeinaJi: What errors do you get in local for these 2 tests?

cattabiani: the errors are very unhelpful because they appear in a subprocess and do not bubble up. I have found some workarounds but were clunky. If you really want I can investigate further

WeinaJi: I ran in my local `pytest -s -x tests/unit/test_cli.py` and all tests in this file passed. Please check again.

For `tests/unit/test_coreneuron_save_restore.py::test_full_run_vs_save_restore`, the diff is at `tstop` in `SonataReport` mechanism so that's normal after including libSonataReport. Can be excluded.
```
            # Compare the files
>           assert filecmp.cmp(file1, file2, shallow=False)
E           AssertionError: assert False
E            +  where False = <function cmp at 0x11ab64c20>(PosixPath('output_0_26/0_cpu_t26.000000.corenrn'), PosixPath('output_13_26/0_cpu_t26.000000.corenrn'), shallow=False)
E            +    where <function cmp at 0x11ab64c20> = filecmp.cmp
```
Again, is it really needed to include this tox env for local testing? In local, everyone can have his/her way to test. If we push it to the repo,  we need to maintain it to ensure it always works.

cattabiani: again, yes, it is very helpful. I do not get how you can run all the tests with pytest as they are in tox. If you find a better solution I am happy to drop this pr.

about maintaining: this stuff is "on the side" of everything and it is on a best effort approach. It is not in the ci. 

btw, I just found how to make it work even for the 2 tests that you really want to add. In my setup now unit-local-env runs all the tests that unit runs

cattabiani: it is also useful for breaking changes in libsonata because unit installs from pypi (I think)

WeinaJi (approved):
I didn't try in my local. But as this is only used for local not CI, you can merge it.

PR #546, opened 2026-06-05
remove unused `--experimental-stims`

mgeplf:
* can't find where it's used
* "pythonized" `ConfigFile` -> `config_file` in command parsing

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -69,7 +69,7 @@ Once neurodamus is installed, you should be able to find the executable `neuroda
 
   $ neurodamus
     Usage:
-        neurodamus <config_file> [options]
+        neurodamus <ConfigFile> [options]
         neurodamus --help
 
 Among the options you will find flags to tune run behavior.
--- a/README.rst
+++ b/README.rst
@@ -69,7 +69,7 @@ Once neurodamus is installed, you should be able to find the executable `neuroda
 
   $ neurodamus
     Usage:
-        neurodamus <ConfigFile> [options]
+        neurodamus <config_file> [options]
         neurodamus --help
 
 Among the options you will find flags to tune run behavior.
--- a/neurodamus/commands.py
+++ b/neurodamus/commands.py
@@ -19,7 +19,7 @@ def neurodamus(args=None):
     """Neurodamus
 
     Usage:
-        neurodamus <config_file> [options]
+        neurodamus <ConfigFile> [options]
         neurodamus --version
         neurodamus --help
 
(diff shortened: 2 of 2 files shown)


neurodamus/commands.py
```diff
@@ -39,7 +39,6 @@ def neurodamus(args=None):
                                 Otherwise deleted at the end. ``--save=<PATH>`` overrides this.
         --modelbuilding-steps=<number>
                                 Set the number of ModelBuildingSteps for the CoreNeuron sim
-        --experimental-stims    Shall use only Python stimuli? [default: False]
```
WeinaJi, line 42: Good spot. It was used while migration from hoc to python for stimulus. No more needed.

neurodamus/commands.py
```diff
@@ -19,7 +19,7 @@ def neurodamus(args=None):
     """Neurodamus
 
     Usage:
-        neurodamus <ConfigFile> [options]
+        neurodamus <config_file> [options]
```
WeinaJi: Sure this is working and more pythonic. But we still use `--configFile` in the command line. Although the args is parsed and passed in init.py, but I think it is better to be consistent with the command line to avoid any confusion.

mgeplf: > But we still use --configFile in the command line. Although the args is parsed and passed in init.py, but I think it is better to be consistent with the command line to avoid any confusion.

Any idea why the tests pass?

WeinaJi: Because in `neurodamus/data/init.py` 
```
args = extract_arguments(sys.argv)
commands.neurodamus(args)
```
The `extract_arguments` function  extracts the args from `--configFile`.  How is the argument named in `neurodamus.__doc__` doesn't change the behavior of `extract_arguments`. But I feel it might confuse developers if we rename them differently. Just my feeling.

mgeplf: I'll remove that change.

However, it's the docs that are misleading, I see: the `--configFile` "option" isn't described.

That documentation I changed is saying `neurodamus simulation_config.json` is the command line, and you're talking about `neurodamus --configFile simulation_config.json`

WeinaJi: `--configFile` is  when we launch via `init.py`: `special -mpi -python $NEURODAMUS_PYTHON/init.py --configFile=simulation_config.json`
`neurodamus --configFile simulation_config.json` is not a valid command I think.

README.rst
```diff
@@ -69,7 +69,7 @@ Once neurodamus is installed, you should be able to find the executable `neuroda
 
   $ neurodamus
     Usage:
-        neurodamus <ConfigFile> [options]
+        neurodamus <config_file> [options]
```
WeinaJi: This change is not needed either

WeinaJi (approved):
LGTM

PR #550, opened 2026-06-10
Numpy-based LFP report for NEURON

### LINKED ISSUE #549: LFP: check if it is possible to do reports with neuron

### LINKED ISSUE #553: LFP: split unit tests
We should move this in libsonata. In the mean time, tests and object should get their own files so the swap is trivial. It is good in general to have this organized

cattabiani:
Add numpy-based LFP reporting for NEURON (previously CoreNEURON-only).

- `WeightedSummationReport` base class using numpy vectorized matrix multiply
  instead of per-compartment ALU point processes
- `LFPReport` subclass: reads electrode scaling factors from HDF5, computes
  `output = beta @ i_membrane_values` via a pre-record callback in SonataReportHelper
- One Python callback per report per rank (vs thousands of ALU MOD events)
- Registered in `ReportManager` like all other report types (no special-case factory logic)

## How it works

- `SonataReportHelper.mod` exposes a C function pointer (`pre_record_callback`)
- `WeightedSummationReport.register_pre_record_callback()` sets it via ctypes
- Before each `sonata_record_data`, the callback reads `i_membrane_` from all
  compartments and writes the weighted sum into a NEURON Vector buffer
- SonataReport samples the buffer at `report_dt` as usual
- replaced LD_PRELOAD with a targeted ctypes.CDLL(path, mode=RTLD_GLOBAL) call in _neurodamus.py. After NEURON loads libnrnmech.so (with RTLD_LOCAL), we re-open the already-loaded library with RTLD_GLOBAL to promote its symbols. No env-level changes, no impact on child processes.

## Testing

- Scientific test validates NEURON-only path produces same results as CoreNEURON
- Unit tests for `LFPFileReader` extracted into `tests/unit/test_lfp_reader.py`
- Multi-report tests (multiple electrodes files) ready but disabled pending CI support

Fix: https://github.com/openbraininstitute/neurodamus/issues/549
Fix: https://github.com/openbraininstitute/neurodamus/issues/553

Base for: https://github.com/openbraininstitute/neurodamus/issues/549

Performance analysis: https://github.com/openbraininstitute/neurodamus/issues/544


.github/workflows/simulation_test.yml
```diff
@@ -30,6 +30,7 @@ env:
   SCCACHE_DIR: ${{ github.workspace }}/cache/sccache
   UV_CACHE_DIR: ${{ github.workspace }}/cache/uv
   BASE_DIR: ${{ github.workspace }}
+  LD_LIBRARY_PATH: ${{ github.workspace }}/opt/obi/lib
```
cattabiani: We removed LD_LIBRARY_PATH thinking HDF5 was the only thing needing it, but libsonatareport also lives in that prefix and Linux needs the hint at runtime.

macos ci was already fine

neurodamus/core/coreneuron_configuration.py
```diff
@@ -31,11 +31,8 @@ def process_section(
     ):
         secvec, segvec, lfp_factors = Nd.Vector(), Nd.Vector(), Nd.Vector()
         num_segments = 0
-        section_attr = getattr(cell._cellref, sec_list, None)
-        if section_attr:
-            for sec in section_attr:
-                section_id = cell.get_section_id(sec)
-                num_segments += self.create_section_vectors(section_id, sec, secvec, segvec)
+        for section_id, sec in cell.iter_section_list(sec_list):
+            num_segments += self.create_section_vectors(section_id, sec, secvec, segvec)
```
cattabiani, line 35: lfp ordering problem. coreneuron vs bluerecording vs neurodamus need to agree on the order or we apply weights to the wrong compartments

.github/workflows/simulation_test.yml
```diff
@@ -21,7 +21,7 @@ on:
 env:
   LIBSONATAREPORT_COMMIT: 2.0.0
   LIBSONATA_COMMIT: v0.1.37
-  NEURON_COMMIT: 2ac5cc7191e44805cdf40abf0ad6d3fac1481d49
+  NEURON_COMMIT: b0dcd70f923449245c854d9b65b246b28bcb4878
```
cattabiani, line 25: we need the fix for:

https://github.com/neuronsimulator/nrn/pull/3813

for our smoke test

cattabiani:
the coverall drop is because I added a new kind of reports and they are difficult to test in unit. Things are tested in scientific mostly

PR #555, opened 2026-06-18
Add `neurodamus-compile-mods` command to simplify building of NEURON models

mgeplf:
* since "scientific" and "simulator" mod files have been separated, we can simplify building the required NEURON/CORENEURON libraries but having `neurodamus` supply its "simulator" files
* the `neurodamus-compile-mods` can take directories and a SONATA circuit config to gather all the required mod files
* deduplication happens based on hash of the file, and the file name. The last location for a file is the one that takes precendence
* Caching is included, since compilation can be slow
* Switched unit tests, CI and Dockerfile to the new system
* Allow manual release of docker images, update docker image


mgeplf:
For https://github.com/openbraininstitute/prod-build-circuit/issues/37

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,205 @@
...
+    assert spec
+    assert spec.origin
+    path = Path(spec.origin).parent / "data/mod/"
+    return [path]
+
+
+def build_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--internal-mods", action="store_true", help="Include neurodamus internal mods"
```
eleftherioszisis: Would "--with-internal-mods" be a little more descriptive?

mgeplf:
I'll add more test for coverage, but I'd like some feedback now if possible.

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,205 @@
...
+    return h.hexdigest()
+
+
+def _get_mod_files(input_dirs: list[Path]) -> dict[Path, str]:
+    # files with the same names; the last one "wins"
+    files = {
+        f.name: f.absolute() for f in it.chain.from_iterable(d.glob("*.mod") for d in input_dirs)
+    }
+    # deduplicate based on md5 hash
+    files = {_md5sum(p): p for p in files.values()}
```
WeinaJi: Is this to remove the mod file which has a different file name but the same file content?
I think `nrnivmodl` will raise an error about duplicate POINT_PROCESS name in this case, we should not skip it silently.

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,205 @@
...
+
+def _check_cache(mod_files: dict[Path, str], output_dir: Path, options: Options) -> bool:
+    metadata = _metadata_path(output_dir)
+    if not metadata.exists():
+        return False
+
+    with metadata.open() as fd:
+        old = json.load(fd)
+    new = _generate_mod_metadata(mod_files, options)
+    return old == new
```
WeinaJi: Is this check of "modules.json" sufficient? Should we also check if the library `libnrnmech{ext}` exists?

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,205 @@
...
+    return _output_files(output_dir, options)
+
+
+def _output_files(output_dir: Path, options: Options) -> dict[str, str]:
+    base = (output_dir / platform.machine()).absolute()
+    ext = ".dylib" if sys.platform == "darwin" else ".so"
+
+    # "/Users/runner/work/neurodamus/neurodamus/opt/obi/neocortex/arm64/./libnrnmech.dylib"
+    libnrnmech = base / f"libnrnmech{ext}"
+    assert libnrnmech.exists(), f"Missing {libnrnmech}"
```
WeinaJi: ok, I see that the lib is asserted here, but should the process go ahead to build mod files? If it stops here, how do we fix the process? Delete the cache on AWS and launch again, or have a force build option?

tests/unit/test_compile_mods.py
```diff
@@ -0,0 +1,110 @@
...
+def test__internal_mods_path():
+    paths = test_module._internal_mods_path()
+    assert len(paths) == 1
+    assert paths[0].name == "mod"
+    assert paths[0].exists()
+
+
+#def test__build_modules(tmp_path):
+#    options = Options(simulator=Simulator.coreneuron, incflags="", loadflags="")
+#    res = test_module._build_modules(tmp_path, "echo", options)
```
WeinaJi: Are the commented out lines still needed?

.github/workflows/docker-image.yml
```diff
@@ -3,7 +3,11 @@ name: Build and Push Docker Image
 on:
   release:
     types: [published]
-
+  workflow_dispatch:
```
cattabiani, line 6: handy!

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,205 @@
...
+def _md5sum(path: Path) -> str:
+    h = hashlib.md5(usedforsecurity=False)
+    with open(path, "rb") as f:
+        for chunk in iter(lambda: f.read(8192), b""):
+            h.update(chunk)
+    return h.hexdigest()
+
+
+def _get_mod_files(input_dirs: list[Path]) -> dict[Path, str]:
+    # files with the same names; the last one "wins"
```
cattabiani: can you promote this to an actual docstring?

neurodamus/utils/compile_mods.py
cattabiani: asserts are stripped with `python -0`. A proper `FileNotFoundError` would be better

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,205 @@
...
+    return [path]
+
+
+def build_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--internal-mods", action="store_true", help="Include neurodamus internal mods"
+    )
+    parser.add_argument("--circuit-config", help="Circuit config for `mechanisms_dir` discovery")
+    parser.add_argument("--input-dir", nargs="*", help="Input directory")
```
cattabiani, line 239: edge case:

--input-dir with no dirs and no --circuit-config or --internal-mods

_get_mod_files([]) returns {} and _build_modules then calls nrnivmodl with an empty mod_files dir. You should check that at least a source of mod files exists no?

ci/build_ndcore.sh
```diff
@@ -1,77 +1,48 @@
...
+MODS_DIR="${2:-}"
+
+CMD=(
+    neurodamus-compile-mods
+    --input-dir $MODS_DIR
+    --output-dir $BUILD_DIR
+    --internal-mods
+    --incflags="-DDISABLE_REPORTINGLIB"
+    --output-type shell
+    )
```
cattabiani, line 28: If $2 isn't provided, --input-dir gets an empty string argument. The unit-allen-v1 tox env does exactly this.
No second arg. The --input-dir "" will try to resolve Path("").absolute() which is the current directory

was this intended? (connected to previous issue)

tox.ini
```diff
@@ -116,8 +116,9 @@ allowlist_externals =
     {toxinidir}/ci/build_ndcore.sh
     sh
 commands =
-    {toxinidir}/ci/build_ndcore.sh -allen_v1 {toxworkdir} {toxinidir}/neurodamus/data
-    sh -c ". {toxworkdir}/.envfile && pytest -x --cov=neurodamus --cov-report=term-missing --forked {posargs:tests/unit-allen-v1}"
+#{toxinidir}/ci/build_ndcore.sh -allen_v1 {toxworkdir} {toxinidir}/neurodamus/data
```
cattabiani: did you forget to remove this commented line?
```suggestion
```

tests/unit/test_compile_mods.py
cattabiani:
```suggestion
```

did you forget to remove this code?

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,205 @@
...
+
+    mod_dir = (output_dir / "mod_files").absolute()
+    mod_dir.mkdir(exist_ok=True)
+    for path in mod_files:
+        shutil.copy(path, mod_dir)
+
+    cmd.append(str(mod_dir))
+
+    res = subprocess.Popen(cmd, cwd=str(output_dir), stdout=sys.stderr)  # noqa: S603
+    res.wait()
```
cattabiani: You're calling Popen + .wait() with no interaction with the process in between

consider subprocess.run. It is simpler and clearer
```suggestion
    res = subprocess.run(cmd, cwd=str(output_dir), stdout=sys.stderr)  # noqa: S603
```

docs/source/build_mod_files.rst
```diff
@@ -0,0 +1,52 @@
+Building MOD Files
+==================
+
+MOD (or NMOD) files are a way of extending the behavior of NEURON.
+See `Using NMODL Files <https://www.neuronsimulator.org/en/latest/courses/using_nmodl_files.html>`_ for more details.
+
+When running neurodamus, these files need to be pre-compiled and available via the ``NRNMECH_LIB_PATH`` (for `NEURON`) and ``CORENEURONLIB``  (for `coreneuron`) environment variables.
+
+Generally, there are 3 classes of MOD files considered within neurodamus; `internal support`, `shared support` and `scientific`.
+Both the `support` ones are bundled with the neurodamus install, so the user only needs to supply the `scientifc` ones.
```
cattabiani:
```suggestion
Both the `support` ones are bundled with the neurodamus install, so the user only needs to supply the `scientific` ones.
```

docs/source/build_mod_files.rst
```diff
@@ -0,0 +1,52 @@
...
+Will compile the files within `some-mod-dir` along with the support MOD files since `--internal-mods` was specified.
+The compiled files will be put in the `output` directory.
+
+Finally, the `--output-type` specifies what the tool will return on its ``STDOUT``.
+If `shell` is chosen, then environment variable setting is output:
+
+EX::
+
+    NRNMECH_LIB_PATH=/path/to/libnrnmech.so
+    SPECIALS_PATH=/some/path/to/where/the/speciels/are
```
cattabiani:
```suggestion
    SPECIALS_PATH=/some/path/to/where/the/specials/are
```

ci/scripts/make-neurodamus-nrnivmodl.sh
```diff
@@ -0,0 +1,56 @@
...
+
+ARGS=()
+[[ -n $CORENEURON ]] && ARGS+=("-coreneuron")
+ARGS+=("-incflags" "$INCFLAGS")
+ARGS+=("-loadflags" "$LOADFLAGS")
+[[ -n $NMODL ]] && ARGS+=("-nmodl" "$NMODL")
+[[ -n $NMODLFLAGS ]] && ARGS+=("-nmodlflags" "$NMODLFLAGS")
+
+
+echo "nrnivmodl "${ARGS[@]}" "${POSITIONAL[@]}""
```
cattabiani:
```suggestion
echo "nrnivmodl ${ARGS[*]} ${POSITIONAL[*]}"
```

cattabiani: I suspect the quotes are wrong for an echo

ci/build_ndcore.sh
cattabiani, line 28: nitpick:
you are using both ` ` and `=` for the flags. both work fine with argparse but it would be better to agree on one

cattabiani (commented):
a few comments, nothing big

neurodamus/utils/compile_mods.py
mgeplf: sure, done

neurodamus/utils/compile_mods.py
mgeplf: Why not skip it silently?  I don't think the filename is important, just the contents.

The only way it will be skipped is if the files contain the same content, so whoever is compiling the mods (which is us, it's not really something users will call, imo) isn't too concerned as long as the whateever is in the mod files is required.  This, for instance, allows us to the do the rename to keep the files the same as the `Point Process` name.

neurodamus/utils/compile_mods.py
mgeplf: Yeah, it's to make sure that what is compiled in the target directory is the same as what is expected; if I don't do the check, if there is a change to the mod file contents, then there would be a mistmatch, and things get hard to debug then if since the data is stale.

neurodamus/utils/compile_mods.py
mgeplf: Good point, I will change it to compile the recompile the mod files.

tests/unit/test_compile_mods.py
mgeplf: that's part of me going for more coverage, it won't be there in the final version.

neurodamus/utils/compile_mods.py
mgeplf: done

neurodamus/utils/compile_mods.py
cattabiani: I like the system

images

neurodamus/utils/compile_mods.py
mgeplf: good point, done

ci/build_ndcore.sh
mgeplf: > The --input-dir "" will try to resolve Path("").absolute() which is the current directory

Did you try this? When I run it with:
`neurodamus-compile-mods --input-dir --output-dir /home/mgevaert/src/neurodamus/.tox/unit/mod --with-internal-mods --incflags=-DDISABLE_REPORTINGLIB --output-type shell`; the `args.input_dir` is an empty list, so there is no current directory getting found.

tox.ini
mgeplf: done

tests/unit/test_compile_mods.py
mgeplf: part of my attempt to increase coverage, will be used later.

neurodamus/utils/compile_mods.py
mgeplf: ok

docs/source/build_mod_files.rst
mgeplf: done

docs/source/build_mod_files.rst
mgeplf: done

ci/scripts/make-neurodamus-nrnivmodl.sh
mgeplf: I removed the echo

ci/build_ndcore.sh
mgeplf: I need to use the `=` since `-D` is to be passed to nrnivmodl, so I have to do that one this way.  I could switch the other ones, but to me it draws attention to the difference.  Up to you.

mgeplf:
Note the tests are failing because they use a newer libsonata and we are pinning to `<=0.1.36`

cattabiani (dismissed):
LGTM if and when the CI passes

neurodamus/utils/compile_mods.py
WeinaJi: If a circuit comes with 2 mod files that contain the same content, that's wrong, right? The provider of the mod files should be informed and fix that. With this script one of the mod files will be skipped and we will not know such issue.

cattabiani: it is confusing for sure. Is it wrong? I do not think so. 

Example:

you add mech 1 for a set of cells, and mech 2 for another set and compare the results vs reference (everything with mech 1 for example). After, you tweak mech 2 to be different. In this way you are sure that any changes are because of the changes to mech 2 instead of how and where you place the mechs

WeinaJi: Not sure that you get your example well.

> you add mech 1 for a set of cells, and mech 2 for another set and compare the results vs reference (everything with mech 1 for example).

mech 1 and mech2 are with the same contents? If so, why do we keep mech2 in the system?

WeinaJi: I meant, why do we keep mech2 on AWS if it is the same as mech1 ?

cattabiani: as development. You start from the same and then move away from brnchmark. Will it be possible in the platform? I am not sure of that

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,245 @@
...
+    if _check_cache(mod_files, output_dir, options):
+        return _output_files(output_dir, options)
+
+    nrnivmodl = nrnivmodl_path or shutil.which("neurodamus-nrnivmodl") or shutil.which("nrnivmodl")
+    if not nrnivmodl:
+        sys.exit("nrnivmodl not found in PATH")
+
+    cmd = [nrnivmodl]
+
+    incflags = []
```
eleftherioszisis: Is this needed? Seems to always be an empty list.

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,245 @@
...
+    if not mod_files:
+        msg = "No mod files selected to be compiled"
+        raise RuntimeError(msg)
+
+    if _check_cache(mod_files, output_dir, options):
+        return _output_files(output_dir, options)
+
+    nrnivmodl = nrnivmodl_path or shutil.which("neurodamus-nrnivmodl") or shutil.which("nrnivmodl")
+    if not nrnivmodl:
+        sys.exit("nrnivmodl not found in PATH")
```
eleftherioszisis: Why not raise here too?

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,245 @@
...
+    return h.hexdigest()
+
+
+def _get_mod_files(input_dirs: list[Path]) -> dict[Path, str]:
+    """Get all the mod files, with their md5_sum.
+
+    Note: files with the same names; the last one "wins"
+    """
+    files = {
+        f.name: f.absolute() for f in it.chain.from_iterable(d.glob("*.mod") for d in input_dirs)
```
eleftherioszisis: Would it be worth to enforce an explicit ordering of the globbed mod files for the sanity of having a deterministic behavior? I know that insertion ordering is a thing, but I would still prefer using a list here and make the deduplication by name logic explicit.

eleftherioszisis: In addition, what would happen if the user wanted to override a single internal mechanism by providing a file with the same name but different md5? Wouldn't it be ignored because the internal mechanisms are added to the input_dirs list last? Shouldn't it be the other way around that user-defined mechanisms can override the internal ones?

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,245 @@
...
+            cmd.append("-coreneuron")
+
+    if incflags or options.incflags:
+        cmd.extend(["-incflags", f"{options.incflags or ''} {''.join(incflags)}"])
+
+    if options.loadflags:
+        cmd.extend(["-loadflags", options.loadflags])
+
+    mod_dir = (output_dir / "mod_files").absolute()
+    mod_dir.mkdir(exist_ok=True)
```
eleftherioszisis: Does this need to be cleaned in case there are existing mod files in it from previous runs?

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def _extract_mechanisms_dir(circuit_config_path: Path) -> list[Path]:
+    """Get `mechanisms_dir` from SONATA config."""
+    cc = libsonata.CircuitConfig.from_file(circuit_config_path)
+    paths = set()
+    for name in cc.node_populations:
+        properties = cc.node_population_properties(name)
+        if d := properties.mechanisms_dir:
+            paths.add(Path(d).absolute())
```
eleftherioszisis, line 220: I assume that d is resolved from the config path + anchors and is not relative to CWD, right?

neurodamus/utils/compile_mods.py
```diff
@@ -0,0 +1,245 @@
...
+def compile_mods():
+    args = build_parser().parse_args()
+    input_dirs = [Path(d).absolute() for d in args.input_dir] if args.input_dir else []
+    output_dir = Path(args.output_dir)
+
+    if args.circuit_config:
+        input_dirs.extend(_extract_mechanisms_dir(Path(args.circuit_config)))
+
+    if args.with_internal_mods:
+        input_dirs.extend(_internal_mods_path())
```
eleftherioszisis, line 262: Related to what I mentioned above about duplication precedence. Wouldn't we want user-defined mechs to have precedence over internals to allow overrides?

neurodamus/utils/compile_mods.py
mgeplf: yup, thanks

neurodamus/utils/compile_mods.py
mgeplf: sure, done

neurodamus/utils/compile_mods.py
mgeplf: I think insertion order is good enough, but I will change it.

Overriding an internal mod file seems like a dangerous use case, so I would prefer not to support it.  If they want to do that, then they shouldn't run it w/ `--with-internal-mods`, imo.

neurodamus/utils/compile_mods.py
mgeplf: yup

neurodamus/utils/compile_mods.py
mgeplf: libsonata does absolute paths.

neurodamus/utils/compile_mods.py
mgeplf: replied above

neurodamus/utils/compile_mods.py
eleftherioszisis: In that case, maybe would be better to not allow at all duplicates by name and raise that shadowing is not allowed?

neurodamus/utils/compile_mods.py
mgeplf: In the latest commit, I have it emit a warning when the md5sum's match, does that help?

WeinaJi: ok, then a warning is better. We will tolerate that kind of duplication.

WeinaJi (dismissed):
LGTM. How about the other reviewers?

neurodamus/utils/compile_mods.py
mgeplf: fixed

neurodamus/utils/compile_mods.py
mgeplf: it's more if there are multiple mod directories; but that use case will come later if/when we combine circuits.

WeinaJi (approved):
LGTM

Issue #563, opened 2026-06-23
libsonata 0.1.37 segfaults due to HDF5 ABI mismatch


cattabiani:
The simulation CI segfaults in `H5Literate2` → `H5E_printf_stack` on all matrix entries (Python 3.10–3.13, Ubuntu 24.04).

The root cause is two incompatible HDF5 libraries loaded in the same process:
- System `libhdf5_openmpi.so.103` (HDF5 1.10.7) installed via `apt-get install libhdf5-openmpi-dev`
- libsonata `v0.1.37` built from source with HighFive, which uses HDF5 1.12+ APIs

At runtime the dynamic linker resolves HDF5 symbols to the wrong library, causing struct layout mismatches and segfaults.

## Fix

Replace `libhdf5-openmpi-dev` from apt with the existing `install-hdf5.sh` script which builds HDF5 1.14.6 from source into `$INSTALL_DIR`. This ensures all components (libsonata, h5py, libsonatareport, NEURON) link against the same HDF5 version.

mgeplf:
This should be fixed separately than in #541, imo.
I don't understand, though, why it's working in the current configuration; what is being perturbed?

cattabiani:
@mgeplf I can't. It is a catch 22 situation. No ci can pass without my pr

cattabiani:
> I don't understand, though, why it's working in the current configuration; what is being perturbed?

libsonata wheel is with a newer hdf5. 

Solution: 

we need to use the script to build hdf5 and not get the one provided by apt

cattabiani:
> neuron-nightly has a breaking change now. Only #541 can pass the ci... it is required but not sufficient. Only the combination of fixes (or more stuff) that I propose in #541 can pass the ci

mgeplf:
fixed with #564 564

Issue #561, opened 2026-06-23
build-neuron.sh fails on cached nrn clone


cattabiani:
`build-neuron.sh` fails with:

```
error: Your local changes to the following files would be overwritten by checkout:
    pyproject.toml
```

This happens because:

1. The CI caches the `$BUILD_DIR/nrn` clone across runs (via `sccache` cache step).
2. After checkout, `change_name.py` modifies `pyproject.toml` in the nrn working tree.
3. On the next CI run, `git checkout FETCH_HEAD` refuses to overwrite the dirty file.

mgeplf:
This should be fixed in a separate PR, IMO.

cattabiani:
@mgeplf again, I cannot do that

cattabiani:
> neuron-nightly has a breaking change now. Only #541 can pass the ci... it is required but not sufficient. Only the combination of fixes (or more stuff) that I propose in #541 can pass the ci

mgeplf:
Fixed by #564

PR #541, opened 2026-06-03
LFP: pass electrode_offsets vector for multi-report support

cattabiani:
In `coreneuron_configuration.py`, `process_section` now passes:
- `Nd.Vector([0, num_electrodes])` when electrodes are present
- `Nd.Vector()` when no electrodes

Companion PR to [neuronsimulator/nrn#3779](https://github.com/neuronsimulator/nrn/pull/3779). Passes electrode offsets as a CSR-style vector instead of a scalar integer to `nrnbbcore_register_mapping`.

Fix: https://github.com/openbraininstitute/neurodamus/issues/553
Fix: https://github.com/openbraininstitute/neurodamus/issues/548

The spec change that triggered this change: https://github.com/openbraininstitute/sonata-extension/pull/34
This matches the updated NEURON-side API that expects an offsets vector.

## Context

This is part of the LFP electrode offsets refactor that prepares the data path for per-report electrode file support (multiple LFP reports with different electrode configurations). See the nrn PR for the full design.

## Testing

- Requires the nrn branch `katta/lfp-electrode-offsets` to be built first
- All existing `test_lfp.py` tests pass unchanged

PR #564, opened 2026-06-23
Fix CI issues:

mgeplf:
* https://github.com/openbraininstitute/neurodamus/issues/561: reset git commit for NRN
* https://github.com/openbraininstitute/neurodamus/issues/563: install hdf5 manually, don't install `libhdf5-openmpi-dev`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,7 +24,7 @@ commands =
 [testenv:unit]
 deps =
     {[testenv]deps}
-    neuron-nightly
+    neuron-nightly<9.0.2.dev71
     morphio
     memray
 setenv =
@@ -65,7 +65,7 @@ commands =
 [testenv:unit-ngv-mpi]
 deps =
     {[testenv]deps}
-    neuron-nightly
+    neuron-nightly<9.0.2.dev71
     morphio
     -r tests/unit-mpi/requirements.txt
 passenv = OMPI_ALLOW_RUN_AS_ROOT,OMPI_ALLOW_RUN_AS_ROOT_CONFIRM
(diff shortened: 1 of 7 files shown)


cattabiani:
fyi this:
```
       self.pc.nrnbbcore_register_mapping(
            cell.gid, sec_type, secvec, segvec, lfp_factors, num_electrodes
        )
E       RuntimeError: hocobj_call error: hoc_execerror: nrnbbcore_register_mapping: arg 6 must be a Vector of electrode offsets. Version mismatch: update neurodamus or downgrade NEURON.
```
is the error that you get when you try to use the new neuron-nightly with the old neurodamus api. #541 has the new call

tox.ini
```diff
@@ -24,7 +24,7 @@ commands =
 [testenv:unit]
 deps =
     {[testenv]deps}
-    neuron-nightly
+    neuron-nightly<9.0.2.dev71
```
cattabiani, line 27: boooo cheater! :P

cattabiani (approved):
LGTM if the ci passes. 

However, you are not capturing the problems that I encountered with using a git commit instead of a tag for neuron since you are still using the old neuron tag for the integration tests.

tox.ini
mgeplf: yeah; but it's true, up until this point, that's the version that needs to be installed - we just happened to get lucky that we didn't need to bound it before now.

mgeplf:
> with using a git commit

Do you mean the problem w/ the reset, or the GitVersion one?
I tried with the commit, and it checked out the right one.

PR #565, opened 2026-06-24
Fix GitVersion parser for shallow-clone builds

### LINKED ISSUE #562: GitVersion crashes on shallow-clone NEURON builds
When NEURON is built from a shallow clone (`git fetch --depth 1`), `git describe` can't find a tag ancestor and produces either:
- An empty string
- A non-standard format like `HEAD (2ac5cc71+) 2026-06-19`

`GitVersion.__init__` fails to parse these and raises `ValueError`, crashing neurodamus at startup before any simulation runs.

cattabiani:
Handle shallow-clone and empty NEURON version strings that crash neurodamus on startup.

Fix: https://github.com/openbraininstitute/neurodamus/issues/562

- Handle `HEAD (commit+) date` version strings from shallow-clone builds
- Handle empty `neuron.__version__` gracefully (treat as recent dev build)
- Add unit tests for both cases
- CI: use NEURON 9.0.1 SHA instead of tag to exercise the fix


mgeplf (approved):

neurodamus/utils/version.py
```diff
@@ -23,16 +24,45 @@ class GitVersion:
...
+    _regex_shallow = re.compile(r"^HEAD\s+\(([0-9a-f]+)\+?\)")
 
     def __init__(self, vstr):
-        """Parse a version string like '9.0a-1485-g0d990513b' or '9.0a'."""
+        """Parse a version string like '9.0a-1485-g0d990513b' or '9.0a'.
+
+        For shallow-clone builds that produce non-standard strings (e.g.
+        'HEAD (2ac5cc71+) 2026-06-19') or empty strings, the version is
+        treated as newer than any parseable release by using a very high
+        base version.
```
WeinaJi, line 36: Does `git fetch origin <COMMIT_ID>` unshallow the history? Can we use that instead of making such assumption ?

cattabiani: No, git fetch --depth 1 origin <SHA> doesn't unshallow, it fetches exactly that one commit. The resulting version string (HEAD (abc123+) date) is baked into the NEURON wheel at build time. By the time neurodamus starts and reads neuron.__version__, there's no repo to unshallow: it's just a string in an installed package.

We could git fetch --unshallow in our build script, but that pulls the full NEURON history (~2GB) which is what we're avoiding with shallow clones

WeinaJi: ok, then it is fine to have such assumption.

WeinaJi (approved):

PR #568, opened 2026-06-26
Randomize gaba rise time only when randomize_gaba_rise_time is True

WeinaJi:
## Context
Fixes https://github.com/openbraininstitute/prod-circuit-simulation/issues/194

`GABAABHelper.hoc` requires `ProbGABAAB_EMS` synapse model. The previous PR tries to always randomize gaba_rise_time and only raises an exception when `randomize_gaba_rise_time` is True, in order to adapt to the old version of `GABAABHelper.hoc`. However, this is not sufficient in mpi case when missing `ProbGABAAB_EMS.mod`, e.g. Allen_V1 circuit, because Neuron doesn't raise an exception that can be caught by neurodamus but ends the mpi process itself. So this PR changes to check `ProbGABAAB_EMS`  and randomize gaba_rise_time  only when `randomize_gaba_rise_time` is explicitly True (default is False set by libsonata). It will work for circuits that don't have `ProbGABAAB_EMS` synapse model.
**Note:** this aligns with neurodamus-models 2.4.4+ after changing the default randomize_Gaba_risetime from True to False in `GABAABHelper.hoc`, but not older version.

## Testing
Move allen_v1 unit tests to mpi test `allen_v1_mpi` so they are sensitive to such issue.


mgeplf (dismissed):

WeinaJi:
@mgeplf , I just did a small change on the error msg and resolved the merging. Could you please approve again? Thanks!

mgeplf (approved):

PR #569, opened 2026-06-26
CI: LD_PRELOAD libnrnmech on Linux for ctypes symbol visibility

### LINKED ISSUE #570: Linux: `undefined symbol: sonata_report_helper_register_pre_record_callback`
Scientific tests fail on Linux CI (sometimes) with `target_simulator: "NEURON"`:

```
neurodamus.utils.pyutils.CumulativeError: Operation failed with multiple errors:
  setup: AttributeError -> .../python3: undefined symbol: sonata_report_helper_register_pre_record_callback
```

The same test passes on macOS and when `target_simulator: "CORENEURON"`.

## Root cause

`sonata_report_helper_register_pre_record_callback` is a C function defined in
`SonataReportHelper.mod`, compiled into `libnrnmech.so`.

Neurodamus accesses it at runtime via `ctypes.CDLL(None)`, which looks up symbols
from the process's **global** symbol table.

On **Linux**, NEURON's `nrn_load_dll()` calls `dlopen("libnrnmech.so", RTLD_LOCAL)`.
`RTLD_LOCAL` keeps the library's symbols private — invisible to `ctypes.CDLL(None)`.

On **macOS**, the dynamic linker uses a flat namespace by default, so all loaded
dylib symbols are globally visible regardless of flags.

CoreNEURON is unaffected because `SonataReportHelper.mod` wraps all reporting code
in `#ifndef CORENEURON_BUILD` — the callback is never called.

## References

- PR: https://github.com/openbraininstitute/neurodamus/pull/569
- `neurodamus/report.py` — `register_pre_record_callback()` method
- `neurodamus/data/mod/SonataReportHelper.mod` — symbol definition
- `ci/scripts/make-env.sh` — where the fix lives

cattabiani:
On Linux, NEURON's `nrn_load_dll()` loads `libnrnmech.so` via `dlopen()` with `RTLD_LOCAL`. This means symbols defined in MOD files (specifically `sonata_report_helper_register_pre_record_callback` from `SonataReportHelper.mod`) are not visible in the process's global symbol table.

When neurodamus calls `ctypes.CDLL(None).sonata_report_helper_register_pre_record_callback`, it searches the global symbol table and fails with:

```
undefined symbol: sonata_report_helper_register_pre_record_callback
```

This only affects Linux. On macOS, dylibs use a flat namespace that makes all symbols globally visible regardless of how they were loaded.

The issue manifests when running with `target_simulator: NEURON` (not CoreNEURON, because the MOD file guards all reporting code with `#ifndef CORENEURON_BUILD`).

Fix: https://github.com/openbraininstitute/neurodamus/issues/570

## Fix

Add `LD_PRELOAD=$NRNMECH_LIB_PATH` to `env-neocortex.sh` on Linux. This forces `libnrnmech.so` to be loaded with global symbol visibility before NEURON's own `dlopen` call, making the symbol accessible via `ctypes.CDLL(None)`.

The preloaded library is the same one that would be loaded anyway, so there is no performance or correctness cost.

### DIFF
--- a/ci/scripts/make-env.sh
+++ b/ci/scripts/make-env.sh
@@ -21,5 +21,10 @@ make-env() {
     export CORENEURONLIB=$INSTALL_DIR/$ARCH/libcorenrnmech.$LIB_EXT
     export NRNMECH_LIB_PATH=$INSTALL_DIR/$ARCH/libnrnmech.$LIB_EXT
     export PATH=$INSTALL_DIR/$ARCH:$PATH
+    # On Linux, NEURON loads libnrnmech via dlopen with RTLD_LOCAL, hiding its
+    # symbols from ctypes.CDLL(None). LD_PRELOAD forces global visibility.
+    if [[ \$(uname) == Linux ]]; then
+        export LD_PRELOAD=$INSTALL_DIR/$ARCH/libnrnmech.$LIB_EXT
+    fi
 _EOF
 }


mgeplf:
Not sure this is the right way to fix this; `LD_PRELOAD` is heavy handed, and means that all process executed after will have it loaded into their address space.

I think either we need to use `RPATH` or update the search path for the libraries, however that only will happen at a system level.

cattabiani:
the problem is actually related to what I introduced in katta/numpy*. I will fix it directly there

PR #573, opened 2026-06-30
Dockerfile improvements

mgeplf:
* Add sccache to hdf5 build
* clear unused git repositories, to keep the final docker image smaller
* update {NEURON,LIBSONATA}_COMMIT to be in sync with changes to neurodamus


cattabiani (approved):
LGTM

PR #574, opened 2026-07-01
check CI failures

cattabiani:
testing ci failures for scientific


cattabiani:
problem found and fixed in neuron

Issue #576, opened 2026-07-13
Gap Junctions and condition "extracellular_calcium" can't be used together


jamesgking:
When attempting to launch a thalamus simulation using a simulation_config.json generated by the platform, the simulation fails with error:

```
  File "neurodamus/io/synapse_reader.py", line 106, in _patch_scale_U_param
    syn_params.u_hill_coefficient, extra_cellular_calcium
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "numpy/core/records.py", line 455, in __getattribute__
    raise AttributeError("recarray has no attribute %s" % attr) from e
AttributeError: recarray has no attribute u_hill_coefficient
```

## To Reproduce
Use a circuit with Gap Junctions (electrical) and 

"conditions": {
    "extracellular_calcium": 1.1,
    ...
}

PR #577, opened 2026-07-15
Check for "u_hill_coefficient" among synapse parameters

jamesgking:
When condition ca_concentration is present, don't assume that synapse
parameters have "u_hill_coefficient". Check it is present otherwise return

For issue https://github.com/openbraininstitute/neurodamus/issues/576
There running a sim with gap junctions and condition "calcium_concentration" failed when attempting to access nonexistent synapse parameter

Now when loading synapse parameters, the patch_scale_U_param function will early return if "u_hill_coefficient" is not present

### DIFF
--- a/neurodamus/io/synapse_reader.py
+++ b/neurodamus/io/synapse_reader.py
@@ -20,27 +20,9 @@ class SynapseParameters:
     https://sonata-extension.readthedocs.io/en/latest/sonata_tech.html#edge-file
     """
 
-    _fields = {
-        "sgid": np.int64,
-        "delay": np.float64,
-        "isec": np.int32,
-        "ipt": np.int32,
-        "offset": np.float64,
-        "weight": np.float64,
-        "U": np.float64,
-        "D": np.float64,
-        "F": np.float64,
-        "DTC": np.float64,
-        "synType": np.int32,
-        "nrrp": np.int32,
-        "u_hill_coefficient": np.float64,
-        "conductance_ratio": np.float64,
-        "maskValue": np.float64,
-        "location": np.float64,
-    }
-
-    _optional = {"u_hill_coefficient": 0.0, "conductance_ratio": -1.0}
-    _reserved = {"maskValue": -1.0, "location": 0.5}
+    _fields = { "location": np.float64 }
+    _reserved = {"location": 0.5}
+    _optional = {}
 
     @classmethod
     def all_fields(cls):
@@ -83,40 +65,8 @@ class SynapseParameters:
         dt = Nd.dt
         records.delay = (records.delay / dt + 1e-5).astype("i4") * dt
 
-    @staticmethod
-    def _constrained_hill(K_half, y):  # noqa: N803
-        """Constrained Hill function for scaling synaptic parameters.
-
-        Note: it is iused only in scale_U_param. It is its own function
-        because it deserves to be tested separately.
-        """
-        K4 = K_half**4
-        y4 = y**4
-        return (K4 + 16) / 16 * y4 / (K4 + y4)
-
-    @staticmethod
-    def _patch_scale_U_param(syn_params, extra_cellular_calcium, extra_scale_vars):
-        """Scale 'U' and other vars using constrained Hill function based on
-        extracellular calcium.
-        """
-        if extra_cellular_calcium is None or "u_hill_coefficient" not in syn_params.dtype.names:
-            return
-
-        scale_factors = SynapseParameters._constrained_hill(
-            syn_params.u_hill_coefficient, extra_cellular_calcium
-        )
-        syn_params.U *= scale_factors
-        for var in extra_scale_vars:
-            syn_params[var] *= scale_factors
-
     @classmethod
-    def make_synapse_parameters_array(
-        cls,
-        data: dict,
-        extra_fields: list[str],
-        extra_cellular_calcium: float | None,
-        extra_scale_vars: list[str],
-    ):
+    def make_synapse_parameters_array(cls, data: dict, extra_fields: list[str]):
         """Create a recarray from data with optional extra fields and apply patches."""
         if not data:
             return np.recarray(0, dtype=cls.dtype(extra_fields=None))
@@ -134,7 +84,6 @@ class SynapseParameters:
                 raise AttributeError(f"Missing mandatory attribute {name} in the SONATA edge file")
 
         cls._patch_delay_fp_inaccuracies(arr)
-        cls._patch_scale_U_param(arr, extra_cellular_calcium, extra_scale_vars)
 
         return arr
 
@@ -171,14 +120,15 @@ class SonataReader:
     }
 
     def __init__(self, edge_file, population=None, *_, **kw):
-        self._ca_concentration = kw.get("extracellular_calcium")
         self._syn_params = {}  # Parameters cache by post-gid (previously loadedMap)
         self._open_file(edge_file, population, kw.get("verbose", False))
-        # NOTE u_hill_coefficient and conductance_scale_factor are optional, BUT
-        # while u_hill_coefficient can always be readif avail, conductance reader may not.
-        self._uhill_property_avail = self.has_property("u_hill_coefficient")
         self._extra_fields = set()
-        self._extra_scale_vars = []
+
+    @staticmethod
+    def _get_override_attributes(mod_override, suffix):
+        """Return formatted attribute names ("attr1;attr2;attr3") from HOC helper in a list."""
+        attr_names = getattr(Nd, f"{mod_override}Helper_{suffix}", None)
+        return attr_names.split(";") if attr_names else []
 
     def configure_override(self, mod_override):
         if not mod_override:
@@ -187,20 +137,22 @@ class SonataReader:
         override_helper = mod_override + "Helper"
         Nd.load_hoc(override_helper)
 
-        # Read attribute names with format "attr1;attr2;attr3"
-        attr_names = getattr(Nd, override_helper + "_NeededAttributes", None)
-        if attr_names:
+        needed_attrs = self._get_override_attributes(mod_override, "NeededAttributes")
+        if needed_attrs:
             log_verbose(
-                'Reading parameters "{}" for mod override: {}'.format(
-                    ", ".join(attr_names.split(";")), mod_override
-                )
+                'Reading parameters "%s" for mod override: %s',
+                ", ".join(needed_attrs),
+                mod_override,
             )
-            self._extra_fields = set(attr_names.split(";"))
+            self._extra_fields = set(needed_attrs)
 
-        # Read attribute names with format "attr1;attr2;attr3"
-        attr_names = getattr(Nd, override_helper + "_UHillScaleVariables", None)
-        if attr_names:
-            self._extra_scale_vars = attr_names.split(";")
+    def _make_synapse_parameters_array(self, data) -> np.recarray:
+        """Create a synapse parameter array from raw SONATA data.
+
+        Subclasses may override this method to provide additional simulation state
+        required when constructing the parameter array.
+        """
+        return self.Parameters.make_synapse_parameters_array(data, self._extra_fields)
 
     def get_synapse_parameters(self, gid) -> np.recarray:
         """Return the synapse parameters record array for the given gid,
@@ -214,10 +166,9 @@ class SonataReader:
                 self._preload_data_chunk([gid])
                 data = self._data[gid]
 
-            # create the synapse parameters array, already patched
-            syn_params = self.Parameters.make_synapse_parameters_array(
-                data, self._extra_fields, self._ca_concentration, self._extra_scale_vars
-            )
+            # create the synapse parameters array, allowing custom logic in derived classes
+            syn_params = self._make_synapse_parameters_array(data)
+
             # cache the results
             self._syn_params[gid] = syn_params
         return syn_params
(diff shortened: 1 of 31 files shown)


cattabiani (dismissed):
LGTM. I have just a non-blocking question:

why is `extracellular_calcium` passed to gap junction readers at all? The `ConnectionManagerBase._open_synapse_reader` unconditionally passes `SimConfig.extracellular_calcium` to every reader. A cleaner long-term fix might be for `GapJunctionManager` to override that and pass `None`, making the intent explicit. 

Anyway, the defensive check in `_patch_scale_U_param` is still better in this case because it's robust against future synapse types that also lack u_hill_coefficient.

**Clearer design ideas:**

Each Parameters class declares what patches it supports. For example, a class attribute like supports_hill_scaling = True/False, or simply checking "u_hill_coefficient" in cls._fields at the class level. Then make_synapse_parameters_array skips _patch_scale_U_param entirely when the class doesn't define the field.

Override _optional and _reserved in GapJunctionConnParameters to empty dicts. Right now the docstring says they're "inherited and apply unchanged," but they reference fields (u_hill_coefficient, conductance_ratio, maskValue) that don't exist in the gap junction _fields. That's dead weight and could confuse future contributors.

Make the reader accept only what it needs. Instead of ConnectionManagerBase always passing extracellular_calcium, the SynapseReader subclass (or its Parameters class) could declare which global conditions it consumes. Then the manager passes only those. This inverts the responsibility: children declare their needs, parent provides.

If you do not want to do this one (I understand, this is a fix after all) maybe you can create an issue to explain the awkwardness of the inheritance in ConnectionManager?

jamesgking:
Good point, I opened issue https://github.com/openbraininstitute/neurodamus/issues/578 for further design as time allows

tests/unit/test_synapse_reader.py
```diff
@@ -104,3 +104,24 @@ def test_conn_manager_syn_stats():
     assert total_synapses == total_synapses_metype_x + additional_synapses
     assert stats.metype_cell_syn_average["metype-x"] == 2
     assert stats.metype_cell_syn_average["metype-y"] == 2
+
+
+def test_no_hill():
+    """
+    Regression test to make sure that syanpse types may not have a
```
mgeplf: `syanpse` -> synapse

neurodamus/io/synapse_reader.py
```diff
@@ -99,7 +99,7 @@ def _patch_scale_U_param(syn_params, extra_cellular_calcium, extra_scale_vars):
         """Scale 'U' and other vars using constrained Hill function based on
         extracellular calcium.
         """
-        if len(syn_params) == 0 or extra_cellular_calcium is None:
+        if extra_cellular_calcium is None or "u_hill_coefficient" not in syn_params.dtype.names:
```
mgeplf: if `extracellular_calcium` is set, but `u_hill_coefficient` is missing, that to me seems like an error that should stop the simulation; the user is trying to set something that will be ignored.

cattabiani: isn't it the case for hippocampus?

mgeplf: > isn't it the case for hippocampus?

Can you expand on this?

cattabiani: sorry, thalamus:

https://github.com/openbraininstitute/neurodamus/issues/576

mgeplf: Yes, that is what this PR is trying to fix; what I'm saying is that maybe we want to rethink this approach: if a config tries to use a feature that doesn't work with the circuit, we shouldn't just silently ignore it - since this means that the user thinks something is working when it's not.

cattabiani: I don't think that concern applies here. `extracellular_calcium` is a global condition: it correctly applies to chemical synapses in the same simulation. Skipping it for gap junctions isn't _silently ignoring a feature that doesn't work_, it's not applying a patch to a synapse type where it's physically meaningless. We can't forbid calcium conditions just because the circuit also has electrical synapses

mgeplf: > gap junctions isn't silently ignoring a feature that doesn't work, it's not applying a patch to a synapse type where it's physically meaningless

Ok, in that case, I don't think detecting that this is a gap junction based on adding `"u_hill_coefficient" not in syn_params.dtype.names` is the correct solution - we're adding technical debt IMO to fix one problem, and I'm still not convinced the solution is correct - `u_hill_coefficient` is optional for chemical synapses; at which point the user is asking to change the `extracellular_calcium` but it's a silent null operation.

cattabiani: You're right; the check conflates "gap junction" with "missing u_hill_coefficient." For chemical synapses, the optional handling already guarantees the field is present with default 0.0, so the only case where it's absent is gap junctions. A cleaner fix would be having `GapJunctionSynapseReader` pass `None` for `extracellular_calcium`, so the existing `if extra_cellular_calcium is None` guard handles it. That makes the intent more explicit

jamesgking: Thanks for suggestions, I am following up

jamesgking: Since extracellular calcium is only needed for chemical synapses, factored those out into a new subclass and reduced actions in the base class. This allows other subclassed SynapseTypes (gap junctions, exp2syn, etc) to also work with condition extracellular calcium without caring about u_hill_coefficient at all

neurodamus/io/synapse_reader.py
```diff
@@ -20,27 +20,9 @@ class SynapseParameters:
...
-        "nrrp": np.int32,
-        "u_hill_coefficient": np.float64,
-        "conductance_ratio": np.float64,
-        "maskValue": np.float64,
-        "location": np.float64,
-    }
-
-    _optional = {"u_hill_coefficient": 0.0, "conductance_ratio": -1.0}
-    _reserved = {"maskValue": -1.0, "location": 0.5}
+    _fields = {}
```
mgeplf: if `location` is in `_reserved`, shouldn't it be here too?

neurodamus/io/synapse_reader.py
```diff
@@ -83,40 +65,8 @@ def _patch_delay_fp_inaccuracies(records):
         dt = Nd.dt
         records.delay = (records.delay / dt + 1e-5).astype("i4") * dt
```
mgeplf, line 77: will `delay` always exist now that it's not in `fields`?

neurodamus/io/synapse_reader.py
```diff
@@ -20,27 +20,9 @@ class SynapseParameters:
     https://sonata-extension.readthedocs.io/en/latest/sonata_tech.html#edge-file
```
mgeplf, line 31: I don't think this comment applies any more, since this contains only the `locations`

neurodamus/io/synapse_reader.py
```diff
@@ -214,10 +166,9 @@ def get_synapse_parameters(self, gid) -> np.recarray:
                 self._preload_data_chunk([gid])
                 data = self._data[gid]
 
-            # create the synapse parameters array, already patched
-            syn_params = self.Parameters.make_synapse_parameters_array(
-                data, self._extra_fields, self._ca_concentration, self._extra_scale_vars
-            )
+            # create the synapse parameters array, allowing custom logic in derived classes
```
mgeplf: this comment isn't necessary, imo; the `_make_synapse_parameters_array` docstring mentions the subclassing

neurodamus/io/synapse_reader.py
```diff
@@ -490,3 +441,88 @@ def get_conn_counts(self, tgids):
...
+        K4 = K_half**4
+        y4 = y**4
+        return (K4 + 16) / 16 * y4 / (K4 + y4)
+
+    @classmethod
+    def _patch_scale_U_param(cls, syn_params, extra_cellular_calcium, extra_scale_vars):
+        """Scale 'U' and other vars using constrained Hill function based on
+        extracellular calcium.
+        """
+        if len(syn_params) == 0 or extra_cellular_calcium is None:
```
mgeplf: when is syn_params empty?

tests/scientific/test_simulation.py
```diff
@@ -131,3 +131,27 @@ def test_v5_gap_junction(create_tmp_simulation_config_file):
...
+        "extra_config": {
+            "conditions": {
+                "v_init": -80,
+                "extracellular_calcium": 1.1
+            }
+        }
+    }
+], indirect=True)
+def test_v5_gap_junction_with_conditions(create_tmp_simulation_config_file):
+    import numpy as np
```
mgeplf: please only import the imports that need to be done lazily in the function, which should hopefully be none.

neurodamus/io/synapse_reader.py
jamesgking: Good point, I have added it so it is clear they work together

neurodamus/io/synapse_reader.py
jamesgking: I have added delay to the base class _fields, to make it clearer that it should be present

neurodamus/io/synapse_reader.py
jamesgking: This comment I modified that the linked spec is for the subclasses since there one can find the more specific files

neurodamus/io/synapse_reader.py
jamesgking: done

neurodamus/io/synapse_reader.py
jamesgking: True, now that this function is restricted to Chemical synapses only, it should never get a 0 length syn_params

tests/scientific/test_simulation.py
jamesgking: done

neurodamus/io/synapse_reader.py
```diff
@@ -490,3 +441,88 @@ def get_conn_counts(self, tgids):
...
+        if extra_cellular_calcium is None:
+            return
+
+        scale_factors = cls._constrained_hill(syn_params.u_hill_coefficient, extra_cellular_calcium)
+        syn_params.U *= scale_factors
+        for var in extra_scale_vars:
+            syn_params[var] *= scale_factors
+
+    @classmethod
+    def make_synapse_parameters_array(
```
mgeplf, line 518: The `SynapseParameters::make_synapse_parameters_array` has a different signature, so I'm a bit concerned that we're not following "normal" inheritance patterns.  Is there any way this could be refactored so that they match?

neurodamus/io/synapse_reader.py
```diff
@@ -490,3 +441,88 @@ def get_conn_counts(self, tgids):
...
+        K4 = K_half**4
+        y4 = y**4
+        return (K4 + 16) / 16 * y4 / (K4 + y4)
+
+    @classmethod
+    def _patch_scale_U_param(cls, syn_params, extra_cellular_calcium, extra_scale_vars):
+        """Scale 'U' and other vars using constrained Hill function based on
+        extracellular calcium.
+        """
+        if extra_cellular_calcium is None:
```
mgeplf, line 509: It doesn't look like we have a test that covers the case where `extra_cellular_calcium` is not None.
https://coveralls.io/jobs/185267294/source_files/8541499674#L488

I think we should have a test in `tests/unit/test_ringcells.py` where this code path is exercised by a short sim, what do you think?  I think it would be useful in this PR since this code is getting reorganized.

neurodamus/io/synapse_reader.py
jamesgking: I now added a Context object which can be subclassed to contain the relevant data according to the synapse type

neurodamus/io/synapse_reader.py
jamesgking: Added such a test

neurodamus/io/synapse_reader.py
```diff
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import dataclasses
```
mgeplf, line 5: why import both `dataclasses` and `from dataclasses import dataclass`?

neurodamus/io/synapse_reader.py
```diff
@@ -12,35 +14,26 @@
 from neurodamus.utils.pyutils import gen_ranges
 
 
+@dataclass
+class SynapseCreationContext:
```
mgeplf, line 18: I don't see the value in having an empty class, as I don't think it's solving the problem wrt class hierarchy; you're still narrowing `ChemicalSynapseParameters::make_synapse_parameters_array`'s `creation_context`.

Perhaps using a Generic would make more sense.

tests/scientific/test_simulation.py
```diff
@@ -131,3 +131,26 @@ def test_v5_gap_junction(create_tmp_simulation_config_file):
...
+            "conditions": {
+                "v_init": -80,
+                "extracellular_calcium": 1.1
+            }
+        }
+    }
+], indirect=True)
+def test_v5_gap_junction_with_conditions(create_tmp_simulation_config_file):
+    from neurodamus import Neurodamus
+    from neurodamus.gap_junction import GapJunctionManager
```
mgeplf, line 150: why is `GapJunctionManager` imported?

tests/unit/test_ringcells.py
```diff
@@ -186,3 +187,45 @@ def test_enable_soma_stimulation(create_tmp_simulation_config_file):
...
+            "src_dir": RINGTEST_DIR
+        }
+    ],
+    indirect=True,
+)
+def test_condition_extracellular_calcium_applies_uhill_patch(create_tmp_simulation_config_file):
+    """Verify that configuring extracellular calcium condition updates Use for chemical synapses.
+    """
+    from neurodamus import Neurodamus
+    from neurodamus.core.configuration import SimConfig
```
mgeplf, line 219: SimConfig is already imported

tests/unit/test_ringcells.py
```diff
@@ -186,3 +187,45 @@ def test_enable_soma_stimulation(create_tmp_simulation_config_file):
...
+    assert SimConfig.extracellular_calcium == 1.2
+
+    tgt = n._target_manager.get_target("RingA:oneCell")
+    postgid = tgt.gids(raw_gids=True)[0]
+
+    expected_use = 16*ChemicalSynapseParameters._constrained_hill(-1.0, 1.2)
+
+    cellref = n.circuits.get_node_manager("RingA").get_cellref(postgid)
+    syn = cellref.synlist.o(0)
+    assert syn.Use == pytest.approx(expected_use)
```
mgeplf, line 231: I think we should run the loaded simulation, and check that it's generating values that are different than the ones that would have been created without the conditions

tests/scientific/test_synapses.py
```diff
@@ -215,7 +215,7 @@ def get_edge_properties(edge_pop, selection, properties=[]):
 
 
 def test__constrained_hill():
-    from neurodamus.io.synapse_reader import SynapseParameters
+    from neurodamus.io.synapse_reader import ChemicalSynapseParameters
```
mgeplf, line 218: this import doesn't need to be scoped to the test

jamesgking:
Hi @darshanmandge,
This is the pending PR which fixes the issue blocking thalamus + gap junctions from running on the platform. Maybe you can sync with @mgeplf on what can be done to get it merged

